PR #15, opened 2017-09-27
Rescan race fix

aakselrod:
This fixes race conditions caused by receiving cfheaders messages out of the order in which they were requested and ensures block subscriptions are always unsubscribed when a rescan exits.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -292,7 +292,6 @@ rescanLoop:
 			select {
 
 			case <-ro.quit:
-				s.unsubscribeBlockMsgs(subscription)
 				return nil
 
 			// An update mesage has just come across, if it points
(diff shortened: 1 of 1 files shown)


jimpo:
Oh, I was looking into the same issue and didn't see you submitted this until I was almost done. I have an alternative change [here](https://github.com/lightninglabs/neutrino/pull/16).

I think this is still missing a few cases:

1. If we rewind and create multiple subscriptions, only the last one gets unsubscribed. 
2. This waits for both filters in the case where we are current, but there's still the case where the rescan loop is catching up (current = false), and it fetches a header with no filters.

aakselrod:
@jimpo, I agree that your PR fixes more issues and I also like the refactor. I'm testing it now; will approve shortly. Thanks!

aakselrod:
Obsoleted by #16

PR #16, opened 2017-09-27
rescan: Refactor main loop and fix a few bugs/race conditions.

jimpo:
This fixes a few bugs that I believe are causing flaky tests on lnd. Some of the bugs are:

1. If an update is applied at the wrong time and hits the second statement handling updates, the block notification for the first block after the rewind would not get sent. Basically, if it hit the update filter clause at the end of the loop instead of the one when `current = true` it would restart at the top of the loop and fetch the next block.
2. If rewinds were applied, multiple subscriptions would be made and only one would be unsubscribed at the end. Now we unsubscribe ever time current is set to false. Furthermore, passing `ro.quit` into the subscription created a race where there were multiple consumers of that channel.
3. We need to wait for both the basic filter and extended filter before notifying about a block. This fixes a race condition where the GetCFilter queries in notifyBlock would fail with an EOF error.

https://github.com/lightninglabs/neutrino/pull/15 also noted some of the issues in 2 & 3, but I think this is a more comprehensive fix.

I also refactored this loop into a few smaller methods that are hopefully clearer to understand.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -292,7 +292,6 @@ rescanLoop:
 			select {
 
 			case <-ro.quit:
-				s.unsubscribeBlockMsgs(subscription)
 				return nil
 
 			// An update mesage has just come across, if it points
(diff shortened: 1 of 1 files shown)


aakselrod:
ACK. LGTM.

aakselrod:
Thanks for the PR!

PR #17, opened 2017-09-28
rescan: add `DisableDisconnectedNtfns` update option for rewind

aakselrod:
This PR adds a new functional `UpdateOption` that disables `OnBlockDisconnected` and `OnFilteredBlockDisconnected` notifications when rewinding a rescan. This will be used to make client implementation easier in `lnd` and therefore should be available to all clients.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -157,7 +157,7 @@ func (s *ChainService) Rescan(options ...RescanOption) error {
 	// all the options that've been passed in.
 	ro := defaultRescanOptions()
 	ro.endBlock = &waddrmgr.BlockStamp{
-		Hash:   *s.chainParams.GenesisHash,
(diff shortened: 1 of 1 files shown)


rescan.go
```diff
@@ -157,7 +157,7 @@ func (s *ChainService) Rescan(options ...RescanOption) error {
 	// all the options that've been passed in.
 	ro := defaultRescanOptions()
 	ro.endBlock = &waddrmgr.BlockStamp{
-		Hash:   *s.chainParams.GenesisHash,
+		Hash:   chainhash.Hash{},
```
Roasbeef: What material impact does this have? Seems it has the ability to feed an invalid hash to the end user?

aakselrod: Resolved in chat. Merging.

PR #19, opened 2017-10-03
ChainService: use default port in `addStringToNetAddr` when not specified

aakselrod:
This PR makes `addStringToNetAddr` a method on `ChainService`. This allows the function to now have access to the service's `ChainParams()` method, which allows it to default to a port if the address passed doesn't specify a port.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -687,7 +687,7 @@ func NewChainService(cfg Config) (*ChainService, error) {
 				}
 
 				addrString := addrmgr.NetAddressKey(addr.NetAddress())
-				return addrStringToNetAddr(addrString)
+				return s.addrStringToNetAddr(addrString)
 			}
 
 			return nil, errors.New("no valid connect address")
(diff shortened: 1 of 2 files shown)


neutrino.go
```diff
@@ -945,13 +946,19 @@ cleanup:
...
-func addrStringToNetAddr(addr string) (net.Addr, error) {
+// addrStringToNetAddr takes an address in the form of 'host:port' or 'host'
+// and returns a net.Addr which maps to the original address with any host
+// names resolved to IP addresses and a default port added, if not specified,
+// from the ChainService's network parameters.
+func (s *ChainService) addrStringToNetAddr(addr string) (net.Addr, error) {
 	host, strPort, err := net.SplitHostPort(addr)
 	if err != nil {
-		return nil, err
+		if strings.HasSuffix(err.Error(), "missing port in address") {
```
Roasbeef: Rather than examining the string, I think we can check directly against the [`AddrError`](https://golang.org/pkg/net/#AddrError) type.

aakselrod: I've tested this; it works and I've updated the code. Thanks!

PR #20, opened 2017-10-12
multi: update to BIP-compliant btcd

aakselrod:
This PR updates the `neutrino` code to be able to correctly use the `btcd` code updated to comply with the cfilter BIP.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -82,7 +82,7 @@ type cfheadersMsg struct {
 type processCFHeadersMsg struct {
 	earliestNode *headerNode
 	stopHash     chainhash.Hash
-	extended     bool
+	filterType   wire.FilterType
 }
 
 // donePeerMsg signifies a newly disconnected peer to the block handler.
@@ -552,11 +552,11 @@ func (b *blockManager) startSync(peers *list.List) {
 		}
 		startHash := startHeader.BlockHash()
 		endHeight := i
-		fType := false
+		fType := wire.GCSFilterRegular
 		if endHeight < bestRegHeight {
 			// We're catching extended headers up to regular headers
 			// so we set the header type to extended.
-			fType = true
+			fType = wire.GCSFilterExtended
 			if endHeight+wire.MaxCFHeadersPerMsg > bestRegHeight {
 				endHeight = bestRegHeight
 			} else {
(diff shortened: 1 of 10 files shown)


blockmanager.go
```diff
@@ -82,7 +82,7 @@ type cfheadersMsg struct {
 type processCFHeadersMsg struct {
 	earliestNode *headerNode
 	stopHash     chainhash.Hash
-	extended     bool
+	filterType   uint8
```
Roasbeef: Perhaps we should introduce an enum here? We can define it within `btcsuite.wire`. That way it cna be re-used across all the contexts where it would be referenced: `btcd`, `neutrino`, `btcutil`, etc. 

The upside of this is that we also get rid of all the seemingly magic numbers, so we can say:
```
filterType == wire.GCSFilterRegular  (or w/e other name)
```

instead of:
```
filterType == 1
```

sync_test.go
```diff
@@ -1024,16 +1024,14 @@ func waitForSync(t *testing.T, svc *neutrino.ChainService,
 				"header for %d (%s) from DB", i, hash)
 		}
 		knownBasicHeader, err =
-			correctSyncNode.Node.GetCFilterHeader(&hash,
-				false)
+			correctSyncNode.Node.GetCFilterHeader(&hash, 0)
```
Roasbeef: Minor nit suggestion: modify this line such that the assignment is on the same line. We can then wrap the two arguments, so something like this:
```go
knownBasicHeader, err = correctSyncNode.Node.GetCFilterHeader(
    &hash, 0,
)
```

This also applies to the same case right below this line.

blockmanager.go
aakselrod: I've updated this to use `wire.FilterType`. We might consider eventually moving to that type throughout instead of `filterdb.FilterType` as well, but that's not urgent IMO.

sync_test.go
aakselrod: I've reformatted this bit of code.

aakselrod:
The build won't pass until we merge https://github.com/Roasbeef/btcutil/pull/5 and https://github.com/Roasbeef/btcd/pull/8 (which itself required a glide update), and update glide to point to those.

Issue #39, opened 2018-01-03
Implement a batched version of GetUtxo


simonhorlick:
An application that requires many concurrent calls to GetUtxo will find that it requests the same blocks from the network multiple times. A batched version of `GetUtxo` could solve this and would also reduce the amount of IO by not repeatedly fetching the same filters.

Implementation plan:
Create a filter consisting of all outpoints in the query.
Figure out which blocks might match by requesting all relevant filters.
Grab those blocks from the network and check the transactions.

halseth:
Caching blocks could also help

simonhorlick:
I also noticed that getting cfilters one-by-one is really killing performance, especially in the expected case where the neutrino server is far away. I want to also try grabbing them in parallel.

Roasbeef:
@simonhorlick we plan to modify the message that neutrino uses to fetch filters to allow _batch_ fetching. This would mean when syncing from offline, we can pipeline all the messages from node to client. Also we can add prefetching heuristics to speculatively fetch filters within reason.

simonhorlick:
Great, that sounds like it would help a lot.

PR #40, opened 2018-01-09
Block Caching

simonhorlick:
This PR is a series of patches that aim to improve performance of neutrino by caching blocks. The original issue is #39.


blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+// If a newer block (by height) is inserted and the cache is already full,
+// then the oldest block will be evicted to make room.
+type MostRecentBlockCache struct {
+	db *bolt.DB
+
+	// A BlockHeaderStore allows translation between block heights and hashes.
+	blockHeaders *headerfs.BlockHeaderStore
+
+	// The maximum number of blocks to store in the cache.
+	capacity int
```
Roasbeef: Why not base the cache off of total bytes allocated instead? This would allow for a greater degree of control of resource usage.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// The number of cache hits and misses. Must be accessed atomically.
+	hits   uint64
```
Roasbeef: These should be moved to the top of the struct in order to avoid issues accessing them atomically 32-bit platforms.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+)
+
+const (
+	dbName           = "blockcache.db"
+	dbFilePermission = 0600
+)
+
+// BlockCache is an interface which represents an object that is capable of
+// storing and retrieving blocks according to their corresponding block hash.
+//
```
Roasbeef: Extra new line here.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// New creates a new instance of a MostRecentBlockCache given a path to store
+// the data.
+func New(dbPath string, capacity int, blockHeaders *headerfs.BlockHeaderStore) (*MostRecentBlockCache, error) {
```
Roasbeef: Lines should respect 80 char columns on a best-effort basis.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *MostRecentBlockCache) PutBlock(block *wire.MsgBlock) error {
+	// The key for this block is its height in the blockchain.
+	// This allows us to evict the oldest entry easily.
+	hash := block.BlockHash()
+	height, err := c.blockHeaders.HeightFromHash(&hash)
+	if err != nil {
+		return err
+	}
+
+	key := []byte(strconv.FormatUint(uint64(height), 10))
```
Roasbeef: This is inefficient, instead a raw serialization of the height should be used instead. See the [`binary/encoding`](https://golang.org/pkg/encoding/binary/) package.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// PutBlock attempts to insert the block into the cache. If the cache is at
+// capacity it will first attempt to evict the oldest block, if the given block
+// is older than all blocks currently in the cache, then no action happens.
+func (c *MostRecentBlockCache) PutBlock(block *wire.MsgBlock) error {
+	// The key for this block is its height in the blockchain.
+	// This allows us to evict the oldest entry easily.
+	hash := block.BlockHash()
+	height, err := c.blockHeaders.HeightFromHash(&hash)
```
Roasbeef: If the cache is switched to be based off of total size, we can actually avoid this additional query all together. As if we switch to total size, the height isn't relevant. 

There's also the question of if this cache should be a proper LRU or MRU cache. By proper I mean that it ignores the block height, and uses a linked list to keep track of the most/least frequently accessed elements, and uses that for eviction. I think this would generate better performance, as many times if a peer is flapping with several channels (for w/e reason), we'll keep fetching those series of blocks. As is now, when a new block is added, we may delete this block that's frequently used erroneously in favor of a newer block.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Insert the block, evicting the oldest block if the capacity has been
+	// reached.
+	err = c.db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(blockCacheBucket)
+
+		isFull := bucket.Stats().KeyN >= c.capacity
```
Roasbeef: If moving to size based, this can be removed (and also the logic below modified a bit).

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// PrintStats prints the cache hit rate to standard output.
+func (c *MostRecentBlockCache) PrintStats() {
```
Roasbeef: Rather than printing directly to `stdout`, this should return some sort of stats struct. This struct can then be used by the caller to either examine the stats (as it's unable to do so currently), and also format the stats as they chose.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *MostRecentBlockCache) FetchBlock(hash *chainhash.Hash) (*wire.MsgBlock,
+	error) {
+	// Look up the height from the hash, then query by height and double
+	// check we got the right block.
+	height, err := c.blockHeaders.HeightFromHash(hash)
+	if err != nil {
+		return nil, err
+	}
+
+	key := []byte(strconv.FormatUint(uint64(height), 10))
```
Roasbeef: Similar comment here about using a compact encoding.

Roasbeef: Actually, when moving to a size based cache, this goes away all together as the key will likely be the hash of the block.

blockcache/blockcache.go
```diff
@@ -0,0 +1,224 @@
...
+	// Serialize the block.
+	w := bytes.NewBuffer(make([]byte, 0, block.SerializeSize()))
+	err = block.Serialize(w)
+	if err != nil {
+		return err
+	}
+
+	// Insert the block, evicting the oldest block if the capacity has been
+	// reached.
+	err = c.db.Update(func(tx *bolt.Tx) error {
```
Roasbeef: All `Update` calls in this PR can safely be modified to use `Batch`. The latter will attempt to coalesce multiple transactions under a single transaction on a best-effort basis.

blockcache/blockcache.go
simonhorlick: Once we've checked a block to determine whether an outpoint is spent, we shouldn't really need to check that block again unless we're searching for a different outpoint, right? In that case older blocks become progressively less useful as it's unlikely that nodes would create a channel then announce it much later, requiring us to check old blocks again. I have no data to back this up, but that's where I was coming from.

simonhorlick:
FYI I've addressed all comments except for implementing a proper LRU cache, which I haven't had time to get to yet.

simonhorlick:
This PR currently caches blocks in boltdb. As iOS really doesn't like to mmap things this causes memory exhaustion pretty quickly.

Roasbeef:
Blocks should never be written to disk IMO. Only kept in an in memory cache.

simonhorlick:
In general I agree, but for mobile the app is likely to be killed very frequently so we’ll probably need some sort of persistence. There’s also the issue I outlined above with having very limited memory on mobile in general.

Roasbeef:
Persistence here is unnecessary.

On Jul 8, 2018 12:39 AM, "Simon Horlick" <notifications@github.com> wrote:

In general I agree, but for mobile the app is likely to be killed very
frequently so we’ll probably need some sort of persistence. There’s also
the issue I outlined above with having very limited memory on mobile in
general.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/lightninglabs/neutrino/pull/40#issuecomment-403269207>,
or mute the thread

Roasbeef:
Superseded by #88.

PR #42, opened 2018-01-24
GetUtxo batching

simonhorlick:
This change adds the ability to batch calls to GetUtxo so that a single scan can search for multiple outpoints. If a scan is in progress when a new call to GetUtxo happens, we check whether it can safely be added to the current batch, if not it will be included in the next batch.

The aim of this change is to reduce the number of times the same block is fetched from the network and to reduce IO when checking filters.

(Original issue is #39 - I split out this change from the block caching as it's pretty unrelated)


rescan.go
```diff
@@ -885,127 +885,32 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
...
 
-		// Fetch the previous header so we can continue our walk
-		// backwards.
-		header, err := s.BlockHeaders.FetchHeaderByHeight(uint32(curStamp.Height))
-		if err != nil {
-			return nil, err
-		}
-		curStamp.Hash = header.BlockHash()
-	}
+	wg.Add(1)
```
stevenroose: This very much relies on the code 100% of the time executing the Result function. Which it seems like it does, but that promise is easily forgotten/broken. I would have gone for a `chan struct{}` that is closed in the result function, so that you can abort waiting when the service is closing:
```go
select {
case <-s.quit:
	return nil, errors.New("service shutting down")
case <-resultChan:
}
```

neutrino.go
```diff
@@ -733,6 +734,9 @@ func NewChainService(cfg Config) (*ChainService, error) {
 		})
 	}
 
+	s.utxoScanner = NewUtxoScanner(&s)
+	s.utxoScanner.Start()
```
stevenroose: Since this starts a goroutine, I guess this should be called in `ChainService.Start` so that a waitgroup item can be added to make sure the scanner is finished when the service is shutting down.

utxoscanner.go
```diff
@@ -0,0 +1,387 @@
...
+
+// GetUtxoRequest is a request to scan for OutPoint from the height StartHeight.
+type GetUtxoRequest struct {
+	OutPoint    *wire.OutPoint
+	StartHeight uint32
+	Result      func(*SpendReport, error)
+}
+
+// Interface exposes the necessary methods for interacting with the blockchain.
+type Interface interface {
```
stevenroose: While this is easier, the method names look strange since they are tied to the ChainService, making it little different than just adding a `*ChainService` in here..

Did you consider passing methods in a `Config` struct? Sorta like this:
```go
type UtxoScannerConfig struct {
	GetBlock           func(chainhash.Hash, ...QueryOption) (*btcutil.Block, error)
	GetBlockHash       func(int64) (*chainhash.Hash, error)
	BestSnapshot       func() (*waddrmgr.BlockStamp, error)
	GetCFilter         func(blockHash chainhash.Hash, filterType wire.FilterType, options ...QueryOption) (*gcs.Filter, error)
	BlockFilterMatches func(ro *rescanOptions, blockHash *chainhash.Hash) (bool, error)
}
```

Could potentially also make the testing easier..

utxoscanner.go
```diff
@@ -0,0 +1,387 @@
...
+			log.Debugf("Fetching block at height %d (%s)", height,
+				hash.String())
+			block, err := s.chainClient.GetBlockFromNetwork(*hash)
+			if err != nil {
+				return requests, err
+			}
+
+			log.Debugf("Got block %d (%s)", height, hash.String())
+
+			if fetch {
```
stevenroose: Always true.

utxoscanner.go
```diff
@@ -0,0 +1,387 @@
...
+					request.Result(nil, err)
+				}
+			}
+		}
+	}()
+}
+
+// getAtHeight returns all GetUtxo requests that have starting height of the
+// given height.
+func (s *UtxoScanner) getAtHeight(height uint32) []*GetUtxoRequest {
```
stevenroose: This name confused me a little, it also seems like it could have been inside the `runBatch` method.. How about `dequeueAtHeight`?

utxoscanner.go
```diff
@@ -0,0 +1,387 @@
...
+				s.cv.L.Unlock()
+				return
+			}
+
+			// Re-queue previously skipped requests for next batch.
+			for _, request := range s.nextBatch {
+				heap.Push(&s.pq, request)
+			}
+
+			s.nextBatch = nil
```
cfromknecht: we should consider using
```golang
s.nextBatch = s.nextBatch[:0]
```
to reuse the backing array

Roasbeef:
[mainnet ready neutrino](https://trello.com/c/aqc3HfMv/20-mainnet-ready-neutrino)

utxoscanner.go
cfromknecht: on second thought, ignore what i said :P 
it might be useful to release this slice so that this can be garbage collected and downsized automatically. otherwise we may retain more space than necessary after the initial rescan.

simonhorlick:
See PR#57

PR #51, opened 2018-02-22
Resolved: TODO(roasbeef): similar interface for headerfs?

tamasblummer:
introduced BlockHeaderDB and FilterHeaderDB interfaces that expose the previously exported functions.


tamasblummer:
Can't do any better.  go fmt run on all files, but travis still complaining.

stevenroose:
`gofmt -s` (which is the same as `go fmt`)

PR #55, opened 2018-04-01
rescan: match against the raw script for an addr, not its datapush

Roasbeef:
In this commit, we update neutrino to match a recent change to the BIP
itself. Rather than looking for the push datas in a script, we'll
actually instead just look for the script itself. This is due to the
fact that we'll now index the entire script rather then each individual
push.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -169,7 +169,12 @@ func (s *ChainService) Rescan(options ...RescanOption) error {
 	// If we have something to watch, create a watch list. The watch list
 	// can be composed of a set of scripts, outpoints, and txids.
 	for _, addr := range ro.watchAddrs {
-		ro.watchList = append(ro.watchList, addr.ScriptAddress())
+		script, err := txscript.PayToAddrScript(addr)
+		if err != nil {
+			return err
+		}
+
+		ro.watchList = append(ro.watchList, script)
 	}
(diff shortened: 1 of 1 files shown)


rescan.go
```diff
@@ -712,8 +738,8 @@ type Rescan struct {
 }
 
 // NewRescan returns a rescan object that runs in another goroutine and has an
-// updatable filter. It returns the long-running rescan object, and a channel
-// which returns any error on termination of the rescan process.
+// updateable filter. It returns the long-running rescan object, and a channel
+// which returns any error on termination of the rescan prcess.
 func (s *ChainService) NewRescan(options ...RescanOption) Rescan {
```
aakselrod: This adds a new typo.

aakselrod:
Verified tests succeed when run against https://github.com/Roasbeef/btcutil/pull/6 (with a `btcd` compiled with the patch).

aakselrod:
Verified `lnd` tests pass against this PR and the `btcutil` PR when run with a `btcd` also built against the `btcutil` PR. Tested:

* `lnwallet`
* `chainntnfs`
* `chainview`

rescan.go
Roasbeef: Fixed.

PR #57, opened 2018-05-04
GetUtxo Batching Proposal #2

cfromknecht:
This PR extends @simonhorlick's original utxo batching PR (#42) with the following items:
 - cache outpoint filter entries, only ever allocate one entry/outpoint
 - batched discovery of initial txns, worst case # hashes/block is now linear
 - reuses backing array for filter entries across entire rescan
 - handles newfound blocks added during rescan
 - do-once Start/Stop with rescan breakout
 - extended unit tests


utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+}
+
+// Returns the GetUtxoRequest with the lowest block height. If no elements are
+// available, then block until one is added.
+func (s *UtxoScanner) peek() *GetUtxoRequest {
+	s.cv.L.Lock()
+	defer s.cv.L.Unlock()
+
+	// Block until the queue is no longer empty.
```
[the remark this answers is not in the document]
simonhorlick: The API for Wait requires the lock to be held:

```
// Wait atomically unlocks c.L and suspends execution
// of the calling goroutine. After later resuming execution,
// Wait locks c.L before returning.
```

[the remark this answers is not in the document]
halseth: Ah, I forgot that this was not a regular mutex. Thanks!

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+			initialTx[outPoint] = tx
+		}
+
+		// Filter the block for any spends using our set of outpoints.
+		spends := findSpends(block.MsgBlock(), height, outpoints)
+		for outPoint, spend := range spends {
+			// Find the request this spend relates to.
+			request, ok := requests[outPoint]
+			if !ok {
+				continue
```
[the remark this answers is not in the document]
cfromknecht: Fixed!

neutrino.go
```diff
@@ -991,6 +1010,7 @@ out:
 	s.connManager.Stop()
 	s.blockManager.Stop()
 	s.addrManager.Stop()
+	s.utxoScanner.Stop()
```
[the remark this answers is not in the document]
cfromknecht: Fixed!

utxoscanner.go
```diff
@@ -0,0 +1,517 @@
...
+		txidReverseIndex[req.OutPoint.Hash] = append(
+			txidReverseIndex[req.OutPoint.Hash], req,
+		)
+	}
+
+	// Iterate over the transactions in this block, hashing each and
+	// querying our reverse index to see if any requests depend on the txn.
+	initialTxns := make(map[wire.OutPoint]*SpendReport)
+	for _, tx := range block.Transactions {
+		txidReqs, ok := txidReverseIndex[tx.TxHash()]
```
Roasbeef: What's the rationale behind this logic? The previous outpoint will appear in the _input_ of each of the transactions. Therefore, it shouldn't be the case that a transaction matches the reverse filter unless it's the transaction that _created_ the output.

utxoscanner.go
```diff
@@ -0,0 +1,517 @@
...
+		case <-s.quit:
+			return spillRequests(), ErrShuttingDown
+		default:
+		}
+
+		log.Debugf("Got block height=%d hash=%s", height, hash)
+
+		// Grab the txns that created the outputs requested at this
+		// block height.
+		txnsInBlock := findInitialTransactions(block.MsgBlock(), reqs)
```
Roasbeef: Why is is necessary that we even locate the initial set of transactions? To ensure that the output was actually even _created_ in the history of the chain?

utxoscanner.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	log.Debugf("Finished batch, %d unspent outpoints", len(requests))
+
+	for _, request := range requests {
+		tx, ok := initialTx[*request.OutPoint]
+		if ok {
+			request.Result(tx, nil)
+		} else {
+			// A nil SpendReport indicates the output was not found.
```
Roasbeef: It seems there isn't a distinction between the output being not found, and not _spent_.

cfromknecht: The above clause handles the case where the output is not spent, as it will return the initial txn. The above clause can also pass nil `tx`.

utxoscanner.go
cfromknecht: Yep, in the current impl we walk back and return the still-unspent txn if no spend was detected.

utxoscanner.go
cfromknecht: Yeh, this method searches for the transactions that create the original output on chain. The method description could use some more specifics, will update.

utxoscanner.go
```diff
@@ -0,0 +1,409 @@
...
+			heap.Push(&s.pq, request)
+		}
+		s.nextBatch = nil
+
+		req := s.pq.Peek()
+		s.cv.L.Unlock()
+
+		// Exit early if a nil request is returned, as this indicates we
+		// are shutting down the utxoscanner.
+		if req == nil {
```
halseth: Can this really occur? Looks like in the case of shutdown, we'll signal above and return checking the `s.quit` channel. 

The only point at which this returns `nil`, seems to be when the heap is empty, and since we have exited the loop above the heap is non-empty and we are holding the lock.

utxoscanner.go
```diff
@@ -0,0 +1,409 @@
...
+
+			// If still no match is found, we have no reason to
+			// fetch this block, and can continue to next height.
+			if !match {
+				continue
+			}
+		}
+
+		// At this point, we've determined that we either (1) have new
+		// requests which we need the block to scan for originating
```
halseth: I don't think it is obvious that `startHeight` is assumed to be the height at which the utxo is created.

batch_spend_reporter.go
```diff
@@ -0,0 +1,261 @@
...
+	// Otherwise, we must reconcile any requests for which the txid did not
+	// exist in this block. A nil spend report is saved for every initial
+	// txn that could not be found, otherwise the result is copied from scan
+	// above. The copied values can include valid initial txns, as well as
+	// nil spend report if the output index was invalid.
+	for _, req := range newReqs {
+		tx, ok := initialTxns[*req.OutPoint]
+		switch {
+		case !ok:
+			log.Errorf("Failed to find outpoint %s -- "+
```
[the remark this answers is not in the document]
cfromknecht: Currently in lnd, we always pass in the shortchanid's `BlockHeight` as the starting height. With this method we are typically interested in getting the UTXO itself, tho I think in the future we can modify it have a persistent height hint cache simiar to lightningnetwork/lnd#1278.

Part of the solution might be to add a bool for each request specifying if we want the initial output, or if we are only interested in watching for spends from that point forward?

neutrino.go
```diff
@@ -783,6 +784,13 @@ func NewChainService(cfg Config) (*ChainService, error) {
 		})
 	}
 
+	s.utxoScanner = NewUtxoScanner(&UtxoScannerConfig{
+		BestSnapshot:       s.BestSnapshot,
+		GetBlockHash:       s.GetBlockHash,
+		BlockFilterMatches: s.blockFilterMatches,
```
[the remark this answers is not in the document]
cfromknecht: Should this be a follow up once batching is finished? Or is there something you have in mind that's currently actionable?

utxoscanner.go
```diff
@@ -0,0 +1,413 @@
...
+		hash, err := s.cfg.GetBlockHash(int64(height))
+		if err != nil {
+			return reporter.SpillRequests(), err
+		}
+
+		// If there are any new requests that can safely be added to this batch,
+		// then try and fetch them.
+		newReqs := s.dequeueAtHeight(height)
+
+		// If an outpoint is created in this block, then fetch it regardless.
```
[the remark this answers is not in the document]
cfromknecht: `GetUtxo` returns the unspent outpoint in the `SpendReport` if the original output is still unspent at the end of the scan.

rescan.go
```diff
@@ -906,133 +906,47 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
...
-			// If we found the transaction that created the output,
-			// then it's not spent and we can return the TxOut.
-			for _, tx := range block.Transactions() {
-				if *(tx.Hash()) == ro.watchOutPoints[0].Hash {
-					outputs := tx.MsgTx().TxOut
-					return &SpendReport{
-						Output: outputs[ro.watchOutPoints[0].Index],
-					}, nil
-				}
+	s.utxoScanner.Enqueue(&GetUtxoRequest{
```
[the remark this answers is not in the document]
cfromknecht: Fixed.

rescan.go
```diff
@@ -906,133 +906,47 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
...
+		OutPoint:    &ro.watchOutPoints[0],
+		StartHeight: uint32(ro.startBlock.Height),
+		Result: func(report *SpendReport, err error) {
+			select {
+			case result <- reportAndError{report, err}:
+			default:
```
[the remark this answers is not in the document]
cfromknecht: moved the population of the channels inside the request object, so caller doesn't have to do anything besides call `Result()`

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+	OutPoint *wire.OutPoint
+
+	// StartHeight is the height we should begin scanning for spends.
+	StartHeight uint32
+
+	// Result is callback returning either a spend report or an error.
+	//
+	// NOTE: The spend report WILL be nil if the target outpoint could not
+	// be found or was not spent above the start height.
+	Result func(*SpendReport, error)
```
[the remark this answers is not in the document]
cfromknecht: Fixed.

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+	quit chan struct{}
+}
+
+// NewUtxoScanner creates a new instance of UtxoScanner using the given chain
+// interface.
+func NewUtxoScanner(cfg *UtxoScannerConfig) UtxoScanner {
+	return UtxoScanner{
+		cfg:  cfg,
+		pq:   make(PriorityQueue, 0),
+		cv:   sync.NewCond(&sync.Mutex{}),
```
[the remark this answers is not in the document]
cfromknecht: Fixed.

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+
+	// Result is callback returning either a spend report or an error.
+	//
+	// NOTE: The spend report WILL be nil if the target outpoint could not
+	// be found or was not spent above the start height.
+	Result func(*SpendReport, error)
+}
+
+// UtxoScannerConfig exposes configurable methods for interacting with the blockchain.
+type UtxoScannerConfig struct {
```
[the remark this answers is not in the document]
cfromknecht: Currently yes they are all fns, though I think it's likely that this will be amended with other non-fn variables. I've found the ability to piecemeal the functions super convenient for testing and mocking out functions

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+	wg   sync.WaitGroup
+	quit chan struct{}
+}
+
+// NewUtxoScanner creates a new instance of UtxoScanner using the given chain
+// interface.
+func NewUtxoScanner(cfg *UtxoScannerConfig) UtxoScanner {
+	return UtxoScanner{
+		cfg:  cfg,
+		pq:   make(PriorityQueue, 0),
```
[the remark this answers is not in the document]
cfromknecht: Fixed.

utxoscanner.go
```diff
@@ -0,0 +1,515 @@
...
+
+	// Insert the request into the queue and signal any threads that might be
+	// waiting for new elements.
+	heap.Push(&s.pq, req)
+
+	s.cv.L.Unlock()
+	s.cv.Signal()
+}
+
+// NOTE: This method MUST be spawned as a goroutine.
```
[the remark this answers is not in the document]
cfromknecht: Fixed

batch_spend_reporter.go
```diff
@@ -0,0 +1,261 @@
...
+// delivered signaling that no spend was detected. If the original output could
+// not be found, a nil spend report is returned.
+func (b *batchSpendReporter) NotifyUnspentAndUnfound() {
+	log.Debugf("Finished batch, %d unspent outpoints", len(b.requests))
+
+	for outpoint, requests := range b.requests {
+		// A nil SpendReport indicates the output was not found.
+		tx, ok := b.initialTxns[outpoint]
+		if !ok {
+			log.Warnf("Unknown initial txn for getuxo request %v",
```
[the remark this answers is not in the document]
cfromknecht: Nope, we expect there to be _some_ entry in `initialTxns`, even if it's nil. This is just here to warn us in case there is an issue.

cfromknecht:
Thanks @halseth :)

Requiring the start height to be the height of the block containing the UTXO is an artifact of mimicking the existing behavior or `GetUtxo`. Within lnd, we always pass in this height correctly, and sometimes require the original output.

https://github.com/lightningnetwork/lnd/blob/master/routing/router.go#L997
https://github.com/lightningnetwork/lnd/blob/master/routing/router.go#L1132

I agree that not having this requirement is more useful, though I think we should push that to a later PR, since this is intended to work as a drop-in replacement. Changing the semantics would invalidate current assumptions in

https://github.com/lightningnetwork/lnd/blob/master/lnwallet/btcwallet/blockchain.go#L43

batch_spend_reporter.go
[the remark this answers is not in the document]
cfromknecht: This case could also be triggered if the request contains a txid that wasn't in the correct block

batch_spend_reporter.go
```diff
@@ -0,0 +1,265 @@
...
+			entry := builder.OutPointToFilterEntry(outpoint)
+			b.outpoints[outpoint] = entry
+			b.filterEntries = append(b.filterEntries, entry)
+		}
+	}
+}
+
+// findInitialTransactions returns a SpendReport for the UTXO, or nil if it does
+// not exist in this block.
+//
```
halseth: extra line

batch_spend_reporter.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		b.notifyRequests(&outpoint, requests, tx, nil)
+	}
+}
+
+// FailRemaining will return an error to all remaining requests in the event we
+// experience a critical rescan error. The error is threaded through to allow
+// the syntax:
+//     return reporter.FailRemaining(err)
```
halseth: ?

utxoscanner.go
```diff
@@ -0,0 +1,410 @@
...
+			select {
+			case <-s.quit:
+				s.cv.L.Unlock()
+				return
+			default:
+			}
+		}
+
+		// Re-queue previously skipped requests for next batch.
+		for _, request := range s.nextBatch {
```
halseth: seems like it could be a problem if a batch is being processed, and only one new request comes in and is added to the `nextBatch`. When the batch processing is over, we won't start the next batch because the pq is empty?

Solution would be to just move this loop before the `IsEmpty` check above I think

cfromknecht: Ahh i see what you're saying, would making the condition:
`for s.pq.IsEmpty() && len(s.nextBatch) == 0 {`
solve this?

utxoscanner.go
```diff
@@ -0,0 +1,410 @@
...
+	mu sync.Mutex
+
+	quit chan struct{}
+}
+
+// deliver tries to deliver the report or error to any subscribers. If
+// resultErrChan cannot accept a new update, this method will not block.
+func (r *GetUtxoRequest) deliver(report *SpendReport, err error) {
+	select {
+	case r.resultErrChan <- reportAndError{report, err}:
```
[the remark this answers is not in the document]
cfromknecht: Yeah, this is definitely more defensive than is necessary if we are 100% certain there are no bugs :) With the previous logic, failing to deliver a ntfn would only block that GetUtxo call. With the batched version, the entire pipeline would grind to a halt if such a bug was introduced.

I've added a warning to the `default` case, so that we can be alerted to any occurrences in the wild.

utxoscanner.go
halseth: Yeah, I think so. Or moving the loop pushing elements from `nextBatch` to the heap, as mentioned.

halseth: And I think the latter is cleanest tbh 😬

batch_spend_reporter.go
```diff
@@ -0,0 +1,268 @@
...
+func (b *batchSpendReporter) FailRemaining(err error) error {
+	for outpoint, requests := range b.requests {
+		b.notifyRequests(&outpoint, requests, nil, err)
+	}
+	return err
+}
+
+// notifyRequests delivers the same final response to the given requests, and
+// cleans up any remaingin state for the outpoint.
```
simonhorlick: Typo "remaingin" -> "remaining"

utxoscanner.go
cfromknecht: If we move the loop before, i think we would have the same issue. Say `nextBatch` is empty when we try to copy elements, and we only wait for the pq to be non-empty. Elements can be added to `nextBatch` without us recognizing.

I think the condition on the pq _and_ `nextBatch` w/ loop after handles these cases though

halseth: I don't think this should be a problem, as elements are added to `nextBatch` only during ongoing scans, which is not happening at this point.

cfromknecht: Ahh you're right! I moved the nextbatch loop to the top as you suggested

cfromknecht:
Addressed most recent batch of review, PTAL

cfromknecht:
Comments in `batchManager` have been fixed 👌

neutrino.go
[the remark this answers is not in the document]
Roasbeef: Yeh IMO should be a follow up rather than modifying this PR to implement. When we implement batch utxo fetching for the routine "catch up to tip" behavior, then we can modify this at that time as well.

utxoscanner.go
```diff
@@ -0,0 +1,415 @@
...
+	mu sync.Mutex
+
+	quit chan struct{}
+}
+
+// deliver tries to deliver the report or error to any subscribers. If
+// resultChan cannot accept a new update, this method will not block.
+func (r *GetUtxoRequest) deliver(report *SpendReport, err error) {
+	select {
+	case r.resultChan <- &getUtxoResult{report, err}:
```
Roasbeef: From my reading above, loos like it's possible  for `deliver` to be called twice? As in when/if we detect a spend, but then at the end (I think?) when we call `notifyAllUnspent`.

Roasbeef: Ah nvm, see the default below :p

utxoscanner.go
```diff
@@ -0,0 +1,415 @@
...
+// Result is callback returning either a spend report or an error.
+func (r *GetUtxoRequest) Result() (*SpendReport, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	select {
+	case result := <-r.resultChan:
+		// Cache the first result returned, in case we have multiple
+		// readers calling Result.
+		if r.result == nil {
```
Roasbeef: Shouldn't we also check the internal result variable at the top before attempting to select? Otherwise, a caller can just get stuck here (if they're late).

cfromknecht: I don’t think it’s necessary since an exclusive lock is held throughout by the first caller, released, then acquired by any potential second calls to `Result`

cfromknecht: So `resultChan` should be closed by the time the lock is released

cfromknecht: So `resultChan` should be closed by the time the lock is released

cfromknecht:
Fixed compile error introduced during rebase. All green now, PTAL

PR #58, opened 2018-05-09
neutrino: use functional option for config passed to NewChainService

parakeety:
- add functional options for `Config` passed to `NewChainService`
- change sync_test.go accordingly

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -1,5 +1,4 @@
 // NOTE: THIS API IS UNSTABLE RIGHT NOW.
-// TODO: Add functional options to ChainService instantiation.
 
 package neutrino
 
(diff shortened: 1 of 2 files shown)


sync_test.go
```diff
@@ -356,21 +356,11 @@ func TestSetup(t *testing.T) {
...
-			h2.P2PAddress(),
-			h1.P2PAddress(),
-		},
-	}
 
 	neutrino.MaxPeers = 3
 	neutrino.BanDuration = 5 * time.Second
 	neutrino.WaitForMoreCFHeaders = time.Second
-	svc, err := neutrino.NewChainService(config)
+	svc, err := neutrino.NewChainService(neutrino.DataDir(tempDir), neutrino.Database(db), neutrino.ChainParams(modParams), neutrino.AddPeers([]string{ h3.P2PAddress(), h2.P2PAddress(),h1.P2PAddress()}))
```
Roasbeef: Style nit: we typically attempt to wrap all lines to 80 columns on a best effort basis.

parakeety:
I fixed the styling. I'll send a PR to lnd 👍

neutrino.go
```diff
@@ -477,6 +476,62 @@ type Config struct {
 	NameResolver func(host string) ([]net.IP, error)
 }
 
+// DataDir is a functional option that allows callers of NewChainService to set the
+// data directory for storing header information.
+func DataDir(dataDir string) func(*Config) {
+	return func(c *Config) {
+		c.DataDir = dataDir
```
halseth: When adding functional options, I think we should unexport the set fields?

parakeety: Good call. It's better to keep scope as small as possible. I'll change the fields to private

PR #59, opened 2018-05-24
build: point to latest btcsuite packages

Roasbeef:

PR #61, opened 2018-05-31
Fix stall detection-caused sync and query issues

aakselrod:
This PR, based on #59, updates two additional places, and those which depend on them:

* The `queryPeers()` function is modified to re-check the connected peers before attempting to query them. This prevents queries failing due to peers disconnecting, since it allows to wait some time for new or reconnected peers. This refactor also reduces the number of goroutines used per query, which improves query speed. The per-peer query function is also factored out, allowing it to be used in other schedulers.

* The `sendGetcfheaders()` function is modified to wait for a minimum number of peers to be connected prior to broadcasting a request. This works *most* of the time in an environment of rapidly-changing peers such as the sync reorg tests; however, an update currently in development will replace this code with query-based code that's optimized for handling this case correctly.

In addition, tests are updated to provide more error information and to support the new code; some small concurrency fixes are also added.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -126,7 +126,7 @@ type spMsg struct {
 	msg wire.Message
 }
 
-// spMsgSubscription sends all messages from a peer ove a channel, allowing
+// spMsgSubscription sends all messages from a peer over a channel, allowing
 // pluggable filtering of the messages.
 type spMsgSubscription struct {
 	msgChan  chan<- spMsg
--- a/neutrino.go
+++ b/neutrino.go
@@ -120,6 +120,19 @@ type cfhRequest struct {
 	stopHash   chainhash.Hash
 }
 
+// spMsg represents a message over the wire from a specific peer.
+type spMsg struct {
+	sp  *ServerPeer
+	msg wire.Message
+}
+
+// spMsgSubscription sends all messages from a peer ove a channel, allowing
+// pluggable filtering of the messages.
+type spMsgSubscription struct {
+	msgChan  chan<- spMsg
+	quitChan <-chan struct{}
+}
+
 // ServerPeer extends the peer to maintain state shared by the server and the
 // blockmanager.
 type ServerPeer struct {
@@ -498,9 +511,7 @@ func (sp *ServerPeer) OnRead(_ *peer.Peer, bytesRead int, msg wire.Message,
 	sp.mtxSubscribers.RLock()
 	defer sp.mtxSubscribers.RUnlock()
 	for subscription := range sp.recvSubscribers {
-		subscription.wg.Add(1)
 		go func(subscription spMsgSubscription) {
-			defer subscription.wg.Done()
 			select {
 			case <-subscription.quitChan:
 			case subscription.msgChan <- spMsg{
--- a/neutrino.go
+++ b/neutrino.go
@@ -12,8 +12,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/lightninglabs/neutrino/filterdb"
-	"github.com/lightninglabs/neutrino/headerfs"
 	"github.com/btcsuite/btcd/addrmgr"
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/chaincfg"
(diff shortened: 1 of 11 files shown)


neutrino.go
```diff
@@ -120,6 +120,19 @@ type cfhRequest struct {
 	stopHash   chainhash.Hash
 }
 
+// spMsg represents a message over the wire from a specific peer.
+type spMsg struct {
+	sp  *ServerPeer
+	msg wire.Message
+}
+
+// spMsgSubscription sends all messages from a peer ove a channel, allowing
```
Roasbeef: ove -> over

neutrino.go
```diff
@@ -498,9 +511,7 @@ func (sp *ServerPeer) OnRead(_ *peer.Peer, bytesRead int, msg wire.Message,
 	sp.mtxSubscribers.RLock()
 	defer sp.mtxSubscribers.RUnlock()
 	for subscription := range sp.recvSubscribers {
-		subscription.wg.Add(1)
```
Roasbeef: Why remove the wait group in this case?

query.go
```diff
@@ -85,15 +109,34 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+	select {
+	case <-sentChan:
+	case <-quit:
+	}
 
-type spMsgSubscription struct {
-	msgChan  chan<- spMsg
-	quitChan <-chan struct{}
-	wg       *sync.WaitGroup
+	// Wait for a quit signal.
```
Roasbeef: What's the use of the select if it falls through anyway? Perhaps should also pass in the quit channel of the external server if it's available?

query.go
```diff
@@ -120,190 +163,44 @@ func (s *ChainService) queryPeers(
...
-		// channel get garbage collected.
-		subwg.Wait()
-	}()
-
 	// Loop for any messages sent to us via our subscription channel and
 	// check them for whether they satisfy the query. Break the loop if
 	// it's time to quit.
-	timeout := time.After(time.Duration(len(peers)+1) *
-		qo.timeout * time.Duration(qo.numRetries))
+	peerTimeout := time.Tick(qo.timeout)
```
Roasbeef: A `time.Ticker` should be used here, as otherwise, we can't actually shutdown the ticker and it'll go on forever with a goroutine leak.

query.go
```diff
@@ -314,8 +211,36 @@ checkResponses:
...
+		// query. Time to select a new peer and query it.
+		case <-peerTimeout:
+			select {
+			case peerQuit <- struct{}{}:
+			default:
+			}
+
+			curPeer = nil
+			for _, curPeer = range s.Peers() {
+				if curPeer != nil && curPeer.Connected() &&
```
Roasbeef: Perhaps we should try to _always_ not use the peer we used prior? As if the peer had issues just a bit ago, likely maybe it still won't be able to respond properly.

neutrino.go
aakselrod: It was causing a data race due to the incorrect way it was being used by the subscription. Those goroutines don't need to shut down cleanly after the quit channel is closed, so I just removed it.

query.go
aakselrod: Ah, thanks, that was just a leftover artifact. I had it passing in the external quit channels originally but I had 2 more already (used differently by different schedulers) and it seems like it'd make more sense to make the caller responsible for a single quit channel if we are going to factor out the scheduler into a functional option. If I could select on a slice of channels, that'd be another option.

query.go
aakselrod: Is that different from just setting the number of retries to 1?

query.go
aakselrod: Thanks, will fix.

neutrino.go
aakselrod: Thanks, will fix.

query.go
Roasbeef: Hmm, I guess not.

query.go
aakselrod: I just looked at the code again. The reason that second quit is there is to delay the unsubscribe. But since we have the timeout in the calling function which sends over/closes that channel, we can just subscribe and unsubscribe and send query messages right in the same function, without starting a new goroutine. The change seems to test well, and now we've eliminated yet another goroutine.

aakselrod:
The next PR with cfheaders sync parallelization and conflict resolution will have faster tx broadcast (query broadcast in general becomes a first-class citizen as it's also required for checkpoint downloading).

Roasbeef:
Committed as ea85047f01c7ba80d11538ede6d54e5eb9b2f91b.

PR #62, opened 2018-06-01
imports: sorts imports sequences

qshuai:
golang: canonical imports sort

### DIFF
--- a/blocklogger.go
+++ b/blocklogger.go
@@ -4,8 +4,8 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


qshuai:
Ok, git it. I think the code change between my commit and the current commit is the main reason. My commit is only to adjust imports sequence.

qshuai:
@coveralls I see @Roasbeef has updated import style [201026a](https://github.com/lightninglabs/neutrino/commit/201026aa3d86f0905f8b68adc007bc958bd3c031). So this commit is timeout.

PR #63, opened 2018-06-08
[wip] mainnet readiness

aakselrod:


neutrino.go
```diff
@@ -445,7 +445,7 @@ func (sp *ServerPeer) OnReject(_ *peer.Peer, msg *wire.MsgReject) {
 func (sp *ServerPeer) OnCFHeaders(p *peer.Peer, msg *wire.MsgCFHeaders) {
 	log.Tracef("Got cfheaders message with %d items from %s",
 		len(msg.FilterHashes), p.Addr())
-	sp.server.blockManager.QueueCFHeaders(msg, sp)
+	//sp.server.blockManager.QueueCFHeaders(msg, sp)
```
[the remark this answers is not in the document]
aakselrod: Right, I had to unwire the existing requests for and processing of cfheaders messages because I'm wiring in the new stuff, especially w.r.t. storage. If I try to process the cfheaders through both paths, the store ends up getting messed up.

blockmanager.go
```diff
@@ -1200,19 +1691,23 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
...
+	}
 }
 
 // sendGetcfheaders sends a getcfheaders message to all peers with the
 // specified start block and hash block and for the specified type of header.
 func (b *blockManager) sendGetcfheaders(startHeight uint32, endBlock *chainhash.Hash,
 	count int, filterType wire.FilterType) {
 	// Send getcfheaders to each peer based on these headers.
-	cfhReq := cfhRequest{
+	/* cfhReq := cfhRequest{
```
aakselrod: Disconnected sending the `getcfheaders` request to peers here as well so it doesn't interfere with the query-based wiring.

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+				if !sp.Connected() {
+					return
+				}
+				lastFailed = handleQuery
+			case m := <-matchSignal:
+				// We got a match signal. If it matches our
+				// query, we can mark this query a success.
+				if m == queryMsgs[handleQuery] {
+					atomic.StoreUint32(
+						&queryStates[handleQuery], 2)
```
Roasbeef: IMO all these magic numbers here should be moved into constants, preferably enum-like variables.

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+	defer func() {
+		for _, quitChan := range peerQuits {
+			close(quitChan)
+		}
+	}()
+
+	for {
+		// Update our view of peers, starting new workers for new peers
+		// and removing disconnected/banned peers.
+		for _, peer := range s.Peers() {
```
Roasbeef: Alternatively, so add a bit more synchronization such that we _only_ get connected peers ever. Or perhaps, grab the mutex ourselves during this loop?

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+			if _, ok := peerQuits[peer]; !ok && peer.Connected() {
```
Roasbeef: I always get a bit nervous when we use a pointer as a key to map. Alternatively, we can attempt to do (ip, port) -> peer perhaps?

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+		for peer, quitChan := range peerQuits {
+			if !peer.Connected() {
+				close(quitChan)
+				close(matchSignals[peer])
+				delete(peerQuits, peer)
+				delete(matchSignals, peer)
+			}
+		}
+	}
+
```
Roasbeef: Style nit: extra new line.

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+//
+// TODO(aakselrod): support for more than one in-flight query per peer to
+// reduce effects of latency.
+func (s *ChainService) queryBatch(
+	// queryMsgs is a slice of queries for which the caller wants responses.
+	queryMsgs []wire.Message,
+
+	// checkResponse is called for every received message to see if it
+	// answers the query message. It should return true if so.
+	checkResponse func(sp *ServerPeer, query wire.Message,
```
Roasbeef: Don't see this called in this method.

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+			for i := uint8(0); i < qo.numRetries; i++ {
+				timeout := time.After(qo.timeout)
+				sp.QueueMessageWithEncoding(queryMsg,
+					nil, qo.encoding)
+				select {
+				case <-quit:
+					return
+				case <-peerQuit:
+					return
+				case <-timeout:
```
Roasbeef: What use is the timeout here? I guess it should be the case that the peer will exit if the caller detects that the correct response was received?

blockmanager.go
```diff
@@ -319,6 +324,492 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, sp *ServerPeer) {
...
+	// the block headers.
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		b.getAllCFHeaders(regCheckpoints, b.server.RegFilterHeaders,
+			wire.GCSFilterRegular)
+		wg.Done()
+	}()
+
+	go func() {
```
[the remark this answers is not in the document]
Roasbeef: In my branch, I've just deleted all references. But yeah, it would make sense to instead just have everything layered as a map, so future changes don't require that much modification.

blockmanager.go
```diff
@@ -319,6 +324,492 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, sp *ServerPeer) {
...
+			return nil, fmt.Errorf("mismatch between filter " +
+				"headers expected to be the same")
+		}
+	}
+
+	// For each header, go through and check whether all headers messages
+	// have the same filter hash. If we find a difference, get the block
+	// and calculate the filter.
+	for i := 0; i < wire.MaxCFHeadersPerMsg; i++ {
+		hash = chainhash.Hash{}
```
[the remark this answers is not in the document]
Roasbeef: Ah nvm then :P

blockmanager.go
```diff
@@ -319,6 +324,492 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, sp *ServerPeer) {
...
+	}
+
+	if len(checkpoints) == 0 {
+		return nil, fmt.Errorf("no peer is serving good cfheaders")
+	}
+
+	// Now we get all of the mismatched CFHeaders from peers, and check
+	// which ones are valid.
+	startHeight := uint32(heightDiff) * wire.CFCheckptInterval
+	headers := b.getCFHeaders(startHeight, fType)
```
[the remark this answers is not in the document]
Roasbeef: 👍

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+		matchSignal <-chan wire.Message) {
+
+		// Subscribe to messages from the peer.
+		sp.subscribeRecvMsg(subscription)
+		defer sp.unsubscribeRecvMsgs(subscription)
+
+		// Track the last query our peer failed to answer and skip over
+		// it for the next attempt. This helps prevent most instances
+		// of the same peer being asked for the same query every time.
+		lastFailed, firstUnfinished, handleQuery := -1, 0, -1
```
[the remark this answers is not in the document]
Roasbeef: 👍

query.go
```diff
@@ -103,14 +122,339 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
...
+	// queryStates holds an atomically updated state for each query. The
+	// states are:
+	// 0 - waiting to be submitted to a peer
+	// 1 - submitted to a peer, waiting for reply
+	// 2 - valid reply received
+	// State transitions are:
+	// 0->1 - send query to peer
+	// 1->0 - query timeout with no acceptable response
+	// 1->2 - acceptable response to query received
+	queryStates := make([]uint32, len(queryMsgs))
```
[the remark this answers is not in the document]
Roasbeef: Ah, yeh agreed.

query.go
Roasbeef: 👍

.gitignore
```diff
@@ -19,3 +19,12 @@ vendor/
 #GoLand config files
 .idea
 *.DS_Store
+
+# vim swap files
+*.swp
```
cfromknecht: YASSS

Roasbeef: kek

sync_test.go
```diff
@@ -60,7 +60,7 @@ var (
 	// "bd":	OnBlockDisconnected
 	// "fd":	OnFilteredBlockDisconnected
 	wantLog = func() (log []byte) {
-		for i := 796; i <= 800; i++ {
+		for i := 1096; i <= 1100; i++ {
```
Roasbeef: Curious as to why shifted by 300 blocks now?

Roasbeef:
Not a blocker for this PR, but once this is in, ideally we also follow up and modify the test to sync against multiple peer (enough to full our outbound set). This would also involve randomly killing peers during the syncing process to exercise the various edge cases in detecting dead peers.

blockmanager.go
```diff
@@ -347,21 +357,100 @@ func (b *blockManager) cfHandler() {
...
+				fType)
+
+			// Catch the filter headers up to the tip of the
+			// block header store and then stay caught up with
+			// each signal while watching for reorgs. We ignore
+			// checkpoints because at this point, we're past
+			// any valid checkpoints.
+			ticker := time.NewTicker(QueryTimeout)
+			defer ticker.Stop()
+			for {
```
Roasbeef: Seems like this won't actually exit? Or is the point that this runs in the background indefinitely after the initial set of checkpoints has been calculated?

blockmanager.go
```diff
@@ -347,21 +357,100 @@ func (b *blockManager) cfHandler() {
...
+		return fmt.Errorf("cfheaders already caught up to blocks")
+	}
+
+	// Query all peers for the responses.
+	headers := b.getCFHeadersForAllPeers(startHeight, fType)
+	if len(headers) == 0 {
+		return fmt.Errorf("couldn't get cfheaders from peers")
+	}
+
+	// For each header, go through and check whether all headers messages
```
Roasbeef: Why can't we just go checkpoints all the way? Or is the idea that this will only be a few blocks at most that were added after we got the full set of checkpoints, but before this?

blockmanager.go
```diff
@@ -493,6 +625,89 @@ func (b *blockManager) getAllCFHeaders(checkpoints []*chainhash.Hash,
 	)
 }
 
+// writeCFHeadersMsg writes a cfheaders message to the specified store. It
+// assumes that everything is being written in order. The hints are required
+// to store the correct block heights for the filters.
```
Roasbeef: What hints?

blockmanager.go
```diff
@@ -493,6 +625,89 @@ func (b *blockManager) getAllCFHeaders(checkpoints []*chainhash.Hash,
...
+			FilterHash: lastHeader,
+		}
+		headerBatch = append(headerBatch, fHeader)
+	}
+
+	// Add the block hashes by walking backwards through the headers.
+	// Also, cache the block headers for notification dispatch.
+	processedHeaders := make([]*wire.BlockHeader, len(headerBatch))
+	curHeader, height, err := b.server.BlockHeaders.FetchHeader(
+		&msg.StopHash)
```
Roasbeef: Ignoring the error here.

Roasbeef: Not a blocker right now, but ideally we also _batch fetch_ these headers just to reduce the read I/O here. They can really all be fetched in one sweep. So at the very least, can you leave a TODO here? Thanks!

blockmanager.go
```diff
@@ -493,6 +625,89 @@ func (b *blockManager) getAllCFHeaders(checkpoints []*chainhash.Hash,
...
+// false if not.
+func verifyCheckpoint(checkpoint *chainhash.Hash,
+	cfheaders *wire.MsgCFHeaders) bool {
+
+	lastHeader := cfheaders.PrevFilterHeader
+	for _, hash := range cfheaders.FilterHashes {
+		lastHeader = chainhash.DoubleHashH(append(hash[:],
+			lastHeader[:]...))
+	}
+	return lastHeader == *checkpoint
```
Roasbeef: 👍

blockmanager.go
```diff
@@ -689,6 +804,127 @@ func (b *blockManager) resolveConflict(
...
+		// header at this height doesn't match the calculated
+		// filter hash.
+		_, maxHeight, err := store.ChainTip()
+		if err != nil {
+			return err
+		}
+
+		if u32i <= maxHeight {
+			/*curHeader, err := store.FetchHeaderByHeight(
+				u32i)
```
Roasbeef: Delete the commented out code?

blockmanager.go
```diff
@@ -697,12 +933,16 @@ func (b *blockManager) getCFHeadersForAllPeers(height uint32,
...
-	// Get the header we expect at the last height represented by the
-	// checkpoint number passed to us.
-	stopHeader, err := b.server.BlockHeaders.FetchHeaderByHeight(
-		height + wire.MaxCFHeadersPerMsg - 1)
-	if err != nil {
-		return headers
+	// Get the header we expect at either the tip of the block header store
+	// or at the end of the maximum-size response message, whichever is
+	// larger.
+	stopHeader, stopHeight, err := b.server.BlockHeaders.ChainTip()
```
Roasbeef: Ignoring the error here.

sync_test.go
```diff
@@ -275,10 +275,10 @@ var testCases = []*testCase{
 		name: "test blocks and filters in random order",
 		test: testRandomBlocks,
 	},
-	&testCase{
+	/*&testCase{
 		name: "check long-running rescan results",
 		test: testRescanResults,
-	},
+	},*/
```
Roasbeef: Reminder to uncomment before presenting the "final" version of this PR.

blockmanager.go
```diff
@@ -319,6 +324,492 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, sp *ServerPeer) {
...
+			}
+			if checkpoint == (chainhash.Hash{}) {
+				checkpoint = *checkpoints[i]
+			}
+			if checkpoint != *checkpoints[i] {
+				return i, nil
+			}
+		}
+		if uint32((i+1)*wire.CFCheckptInterval) <= storeTip {
+			header, err := headerStore.FetchHeaderByHeight(
```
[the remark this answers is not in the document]
aakselrod: Right, this code is run on every start up to catch up any checkpoints that have happened since we were offline. In addition, this plays into the TODO above for counteracting eclipse attacks after restarting and getting honest peers.

query.go
```diff
@@ -103,14 +129,385 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
 	}
 }
 
+type queryState uint32
```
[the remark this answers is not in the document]
aakselrod: Updated comments.

sync_test.go
aakselrod: As we discussed in chat, this is to exercise the checkpoint code.

PR #72, opened 2018-07-12
Use lru cache for filters.

### LINKED ISSUE #65: neutrino: never write filters to disk unless signaled, instead utilize LRU cache
Currently we _always_ write _every_ filter we need to fetch to disk. This results in high-ish disk and I/O utilization when booting back up after a long dormant period, or aggressive usage of the `GetUTXO` call. For "normal" wallets, they'll almost never need to re-match a filter once it's been fetched unless they're doing something like a key import. As a result, we should only actually keep a _limited_ (and configurable) set of filters in memory. 

# Steps To Completion 

- [ ] Modify the [`GetCfilter`](https://github.com/lightninglabs/neutrino/blob/master/query.go#L334) method to _only_ write to disk if specified by the caller. If we don't write the filter to disk, then we should write it to the cache. 

- [ ] Create a new generic (as in `[]byte`) LRU cache. This can double for usage to cache blocks, as well as filters. It may be the case that a more advanced eviction strategy will be more optimal, but we'll at least start with this just to pick the low hanging fruit. 

- [ ] Modify the `GetCfilter` method to first check the cache, then disk if a filter is requested.

rawtxapp:
Resolves #65.

First attempt at the filter cache.
A few things:
- ended up writing a very basic LRU instead of including the lib I pointed out (though it would have been a very lightweight dependency...),
- parse the queryOptions twice to see if persistToDisk was set, seems ugly, but otherwise, it would be necessary to parse it in GetCFilter and change the `queryPeers` signature to take a final queryOption, but then all the callers of `queryPeers` will need to parse it too.
- regular and extended filters share the same cache, should they have different caches ?
- default cache size is 20.

### DIFF
--- a/cache/cacheable_filter.go
+++ b/cache/cacheable_filter.go
@@ -1,0 +1,18 @@
+package cache
+
+import "github.com/btcsuite/btcutil/gcs"
+
+// CacheableFilter is a wrapper around Filter type which provides a Size method
+// used by the cache to target certain memory usage.
+type CacheableFilter struct {
+	*gcs.Filter
+}
+
+// Size returns size of this filter in bytes.
+func (c *CacheableFilter) Size() (uint64, error) {
+	f, err := c.Filter.NBytes()
+	if err != nil {
+		return 0, err
+	}
+	return uint64(len(f)), nil
+}
--- a/cache/lru/lru.go
+++ b/cache/lru/lru.go
@@ -1,0 +1,162 @@
+package lru
+
+import (
+	"container/list"
+	"fmt"
+	"sync"
+
+	"github.com/lightninglabs/neutrino/cache"
+)
+
+// elementMap is an alias for a map from a generic interface to a list.Element.
+type elementMap map[interface{}]*list.Element
+
+// entry represents a (key,value) pair entry in the Cache. The Cache's list
+// stores entries which let us get the cache key when an entry is evicted.
+type entry struct {
+	key   interface{}
+	value cache.Value
+}
+
+// Cache provides a generic thread-safe lru cache that can be used for
+// storing filters, blocks, etc.
+type Cache struct {
+	// capacity represents how much this cache can hold. It could be number
+	// of elements or a number of bytes, decided by the cache.Value's Size.
+	capacity uint64
+
+	// size represents the size of all the elements currenty in the cache.
+	size uint64
+
+	// ll is a doubly linked list which keeps track of recency of used
+	// elements by moving them to the front.
+	ll *list.List
+
+	// cache is a generic cache which allows us to find an elements position
+	// in the ll list from a given key.
+	cache elementMap
+
+	// mtx is used to make sure the Cache is thread-safe.
+	mtx sync.RWMutex
+}
+
+// NewCache return a cache with specified capacity, the cache's size can't
+// exceed that given capacity.
+func NewCache(capacity uint64) *Cache {
+	return &Cache{
+		capacity: capacity,
+		ll:       list.New(),
+		cache:    make(elementMap),
+	}
+}
+
+// evict will evict as many elements as necessary to make enough space for a new
+// element with size needed to be inserted.
+func (c *Cache) evict(needed uint64) error {
+	if needed > c.capacity {
+		return fmt.Errorf("can't evict %v elements in size, since"+
+			"capacity is %v", needed, c.capacity)
+	}
+
+	for c.capacity-c.size < needed {
+		// We still need to evict some more elements.
+		if c.ll.Len() == 0 {
+			// We should never reach here.
+			return fmt.Errorf("all elements got evicted, yet "+
+				"still need to evict %v, likelihood of error "+
+				"during size calculation",
+				needed-(c.capacity-c.size))
+		}
+
+		// Find the least recently used item.
+		if elr := c.ll.Back(); elr != nil {
+			// Determine lru item's size.
+			ce := elr.Value.(*entry)
+			es, err := ce.value.Size()
+			if err != nil {
+				return fmt.Errorf("couldn't determine size of "+
+					"existing cache value %v", err)
+			}
+
+			// Account for that element's removal in evicted and
+			// cache size.
+			c.size -= es
+
+			// Remove the element from the cache.
+			c.ll.Remove(elr)
+			delete(c.cache, ce.key)
+		}
+	}
+
+	return nil
+}
+
+// Put inserts a given (key,value) pair into the cache, if the key already
+// exists, it will replace value and update it to be most recent item in cache.
+func (c *Cache) Put(key interface{}, value cache.Value) error {
+	vs, err := value.Size()
+	if err != nil {
+		return fmt.Errorf("couldn't determine size of cache value: %v",
+			err)
+	}
+
+	if vs > c.capacity {
+		return fmt.Errorf("can't insert entry of size %v into cache "+
+			"with capacity %v", vs, c.capacity)
+	}
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	// If the element already exists, remove it and decrease cache's size.
+	el, ok := c.cache[key]
+	if ok {
+		es, err := el.Value.(*entry).value.Size()
+		if err != nil {
+			return fmt.Errorf("couldn't determine size of existing"+
+				"cache value %v", err)
+		}
+		c.ll.Remove(el)
+		c.size -= es
+	}
+
+	// Then we need to make sure we have enough space for the element, evict
+	// elements if we need more space.
+	if err := c.evict(vs); err != nil {
+		return err
+	}
+
+	// We have made enough space in the cache, so just insert it.
+	el = c.ll.PushFront(&entry{key, value})
+	c.cache[key] = el
+	c.size += vs
+
+	return nil
+}
+
+// Get will return value for a given key, making the element the most recently
+// accessed item in the process. Will return nil if the key isn't found.
+func (c *Cache) Get(key interface{}) (cache.Value, error) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	el, ok := c.cache[key]
+	if !ok {
+		// Element not found in the cache.
+		return nil, cache.ErrElementNotFound
+	}
+
+	// When the cache needs to evict a element to make space for another
+	// one, it starts eviction from the back, so by moving this element to
+	// the front, it's eviction is delayed because it's recently accessed.
+	c.ll.MoveToFront(el)
+	return el.Value.(*entry).value, nil
+}
+
+// Len returns number of elements in the cache.
+func (c *Cache) Len() int {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	return c.ll.Len()
+}
--- a/cache/lru/lru_test.go
+++ b/cache/lru/lru_test.go
@@ -1,0 +1,308 @@
+package lru
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/lightninglabs/neutrino/cache"
+)
+
+func assertEqual(t *testing.T, a interface{}, b interface{}, message string) {
+	if a == b {
+		return
+	}
+	if len(message) == 0 {
+		message = fmt.Sprintf("%v != %v", a, b)
+	}
+	t.Fatal(message)
+}
+
+// sizeable is a simple struct that represents an element of arbitrary size
+// which holds a simple integer.
+type sizeable struct {
+	value int
+	size  uint64
+}
+
+// Size implements the CacheEntry interface on sizeable struct.
+func (s *sizeable) Size() (uint64, error) {
+	return s.size, nil
+}
+
+// getSizeableValue is a helper method used for converting the cache.Value
+// interface to sizeable struct and extracting the value from it.
+func getSizeableValue(generic cache.Value, _ error) int {
+	return generic.(*sizeable).value
+}
+
+// TestEmptyCacheSizeZero will check that an empty cache has a size of 0.
+func TestEmptyCacheSizeZero(t *testing.T) {
+	t.Parallel()
+	c := NewCache(10)
+	assertEqual(t, c.Len(), 0, "")
+}
+
+// TestCacheNeverExceedsSize inserts many filters into the cache and verifies
+// at each step that the cache never exceeds it's initial size.
+func TestCacheNeverExceedsSize(t *testing.T) {
+	t.Parallel()
+	c := NewCache(2)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	assertEqual(t, c.Len(), 2, "")
+	for i := 0; i < 10; i++ {
+		c.Put(i, &sizeable{value: i, size: 1})
+		assertEqual(t, c.Len(), 2, "")
+	}
+}
+
+// TestCacheAlwaysHasLastAccessedItems will check that the last items that
+// were put in the cache are always available, it will also check the eviction
+// behavior when items put in the cache exceeds cache capacity.
+func TestCacheAlwaysHasLastAccessedItems(t *testing.T) {
+	t.Parallel()
+	c := NewCache(2)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	two := getSizeableValue(c.Get(2))
+	one := getSizeableValue(c.Get(1))
+	assertEqual(t, two, 2, "")
+	assertEqual(t, one, 1, "")
+
+	c = NewCache(2)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	c.Put(3, &sizeable{value: 3, size: 1})
+	oneEntry, _ := c.Get(1)
+	two = getSizeableValue(c.Get(2))
+	three := getSizeableValue(c.Get(3))
+	assertEqual(t, oneEntry, nil, "")
+	assertEqual(t, two, 2, "")
+	assertEqual(t, three, 3, "")
+
+	c = NewCache(2)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	c.Get(1)
+	c.Put(3, &sizeable{value: 3, size: 1})
+	one = getSizeableValue(c.Get(1))
+	twoEntry, _ := c.Get(2)
+	three = getSizeableValue(c.Get(3))
+	assertEqual(t, one, 1, "")
+	assertEqual(t, twoEntry, nil, "")
+	assertEqual(t, three, 3, "")
+}
+
+// TestElementSizeCapacityEvictsEverything tests that Cache evicts everything
+// from cache when an element with size=capacity is inserted.
+func TestElementSizeCapacityEvictsEverything(t *testing.T) {
+	t.Parallel()
+	c := NewCache(3)
+
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	c.Put(3, &sizeable{value: 3, size: 1})
+
+	// Insert element with size=capacity of cache, should evict everything.
+	c.Put(4, &sizeable{value: 4, size: 3})
+	assertEqual(t, c.Len(), 1, "")
+	assertEqual(t, len(c.cache), 1, "")
+	four := getSizeableValue(c.Get(4))
+	assertEqual(t, four, 4, "")
+
+	c = NewCache(6)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 2})
+	c.Put(3, &sizeable{value: 3, size: 3})
+	assertEqual(t, c.size, uint64(6), "")
+
+	// Insert element with size=capacity of cache.
+	c.Put(4, &sizeable{value: 4, size: 6})
+	assertEqual(t, c.Len(), 1, "")
+	assertEqual(t, len(c.cache), 1, "")
+	four = getSizeableValue(c.Get(4))
+	assertEqual(t, four, 4, "")
+}
+
+// TestCacheFailsInsertionSizeBiggerCapacity tests that the cache fails the
+// put operation when the element's size is bigger than it's capacity.
+func TestCacheFailsInsertionSizeBiggerCapacity(t *testing.T) {
+	t.Parallel()
+	c := NewCache(2)
+
+	err := c.Put(1, &sizeable{value: 1, size: 3})
+	if err == nil {
+		t.Fatal("shouldn't be able to put elements larger than cache")
+	}
+	assertEqual(t, c.Len(), 0, "")
+}
+
+// TestManySmallElementCanInsertAfterBigEviction tests that when a big element
+// is evicted from the Cache, multiple smaller ones can be inserted without an
+// eviction taking place.
+func TestManySmallElementCanInsertAfterBigEviction(t *testing.T) {
+	t.Parallel()
+	c := NewCache(3)
+
+	err := c.Put(1, &sizeable{value: 1, size: 3})
+	if err != nil {
+		t.Fatal("couldn't insert element")
+	}
+
+	assertEqual(t, c.Len(), 1, "")
+
+	c.Put(2, &sizeable{value: 2, size: 1})
+	two := getSizeableValue(c.Get(2))
+	oneEntry, _ := c.Get(1)
+	assertEqual(t, c.Len(), 1, "")
+	assertEqual(t, two, 2, "")
+	assertEqual(t, oneEntry, nil, "")
+
+	c.Put(3, &sizeable{value: 3, size: 1})
+	assertEqual(t, c.Len(), 2, "")
+
+	c.Put(4, &sizeable{value: 4, size: 1})
+	assertEqual(t, c.Len(), 3, "")
+
+	two = getSizeableValue(c.Get(2))
+	three := getSizeableValue(c.Get(3))
+	four := getSizeableValue(c.Get(4))
+	assertEqual(t, two, 2, "")
+	assertEqual(t, three, 3, "")
+	assertEqual(t, four, 4, "")
+}
+
+// TestReplacingElementValueSmallerSize tests that if an existing element is
+// replaced with a value of size smaller, that the size shrinks and we can
+// insert without an eviction taking place.
+func TestReplacingElementValueSmallerSize(t *testing.T) {
+	t.Parallel()
+	c := NewCache(2)
+
+	c.Put(1, &sizeable{value: 1, size: 2})
+
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+	one := getSizeableValue(c.Get(1))
+	two := getSizeableValue(c.Get(2))
+	assertEqual(t, one, 1, "")
+	assertEqual(t, two, 2, "")
+	assertEqual(t, c.Len(), 2, "")
+}
+
+// TestReplacingElementValueBiggerSize tests that if an existing element is
+// replaced with a value of size bigger, that it evicts accordingly.
+func TestReplacingElementValueBiggerSize(t *testing.T) {
+	t.Parallel()
+	c := NewCache(2)
+
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 1})
+
+	c.Put(1, &sizeable{value: 3, size: 2})
+	assertEqual(t, c.Len(), 1, "")
+	one := getSizeableValue(c.Get(1))
+	assertEqual(t, one, 3, "")
+}
+
+// TestConcurrencySimple is a very simple test that checks concurrent access to
+// the lru cache. When running the test, "-race" option should be passed to
+// "go test" command.
+func TestConcurrencySimple(t *testing.T) {
+	t.Parallel()
+	c := NewCache(5)
+	var wg sync.WaitGroup
+
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			err := c.Put(i, &sizeable{value: i, size: 1})
+			if err != nil {
+				t.Fatal(err)
+			}
+		}(i)
+	}
+
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			_, err := c.Get(i)
+			if err != nil && err != cache.ErrElementNotFound {
+				t.Fatal(err)
+			}
+		}(i)
+	}
+
+	wg.Wait()
+}
+
+// TestConcurrencySmallCache is a test that checks concurrent access to the
+// lru cache when the cache is smaller than the number of elements we want to
+// put and retrieve. When running the test, "-race" option should be passed to
+// "go test" command.
+func TestConcurrencySmallCache(t *testing.T) {
+	t.Parallel()
+	c := NewCache(5)
+	var wg sync.WaitGroup
+
+	for i := 0; i < 20; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			err := c.Put(i, &sizeable{value: i, size: 1})
+			if err != nil {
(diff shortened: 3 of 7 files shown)


rawtxapp:
Rebased on the latest master. Also noticed that there's a change in PR that gets rid of extended filters, should I remove the filter type from cache key and just make it blockhash ?

lru/lru.go
```diff
@@ -0,0 +1,79 @@
...
+
+type elementMap map[interface{}]*list.Element
+
+// list stores entries that let us get the cache key when an entry is evicted.
+type entry struct {
+	key   interface{}
+	value interface{}
+}
+
+type LRUCache struct {
```
halseth: missin godoc

lru/lru.go
```diff
@@ -0,0 +1,79 @@
...
+	if c.ll.Len() > c.size {
+		if elr := c.ll.Back(); elr != nil {
+			c.ll.Remove(elr)
+			delete(c.cache, elr.Value.(*entry).key)
+		}
+	}
+}
+
+// Get will return value for a given key, making the element the most recently
+// accessed item in the process.
```
halseth: not that nil will be returned in case it is not found

lru/lru_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if a == b {
+		return
+	}
+	if len(message) == 0 {
+		message = fmt.Sprintf("%v != %v", a, b)
+	}
+	t.Fatal(message)
+}
+
+func TestEmptyCacheSizeZero(t *testing.T) {
```
halseth: missing godocs on all these tests

neutrino.go
```diff
@@ -66,6 +67,10 @@ var (
 	// DisableDNSSeed disables getting initial addresses for Bitcoin nodes
 	// from DNS.
 	DisableDNSSeed = false
+
+	// DefaultFilterCacheSize is the number of filters neutrino will keep
+	// in memory if no size is specified in the neutrino.Config.
+	DefaultFilterCacheSize = 20
```
halseth: Seems low? I would suggest 200, but again that's just taking a number out of the air

query_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	filter, err := gcs.BuildGCSFilter(20, key, elements)
+	if err != nil {
+		return nil, err
+	}
+
+	return filter, nil
+}
+
+func TestFilterCacheStorage(t *testing.T) {
```
halseth: missing doc

rawtxapp:
Thanks for the review, updated to reflect the comments.

query.go
```diff
@@ -638,6 +643,21 @@ checkResponses:
 	}
 }
 
+func (s *ChainService) getFilterFromCache(blockHash *chainhash.Hash,
+	filterType filterdb.FilterType) *gcs.Filter {
+	cacheKey := filterCacheKey{blockHash: blockHash, filterType: filterType}
+	if cachedFilter, ok := s.FilterCache.Get(cacheKey).(*gcs.Filter); ok {
```
halseth: No need to check `ok`, can return the value directly I think.

query.go
```diff
@@ -659,8 +679,13 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
 	}
 
-	// First check the database to see if we already have this filter. If
+	// First check the cache to see if we already have this filter. If
 	// so, then we can return it an exit early.
+	filter := s.getFilterFromCache(&blockHash, dbFilterType)
+	if filter != nil {
+		return filter, nil
+	}
+	// If not in cache, check if it's in database, returning early if yes.
```
halseth: nit: missing newline above.

halseth:
@rawtxapp Any updates?

query.go
rawtxapp: do you mean just:
`return s.FilterCache.Get(cacheKey).(*gcs.Filter)`
if I do that, the test fails with:
`panic: interface conversion: interface {} is nil, not *gcs.Filter`.

query.go
```diff
@@ -63,6 +63,12 @@ type queryOptions struct {
 	// doneChan lets the query signal the caller when it's done, in case
 	// it's run in a goroutine.
 	doneChan chan<- struct{}
+
+	// persistToDisk indicates whether the filter should also be written
+	// to disk in addition to the memory cache.
```
Roasbeef: Should be wrapped to 80 characters all around.

query.go
```diff
@@ -129,6 +135,14 @@ func DoneChan(doneChan chan<- struct{}) QueryOption {
 	}
 }
 
+// PersistToDisk allows the caller to tell that the filter should be kept
+// on disk once it's found.
+func PersistToDisk(persistToDisk bool) QueryOption {
```
Roasbeef: Why does this need a parameter? The default should be false, so users would only use `PersistToDisk()`.

lru/lru.go
```diff
@@ -0,0 +1,100 @@
...
+	Get(key interface{}) interface{}
+
+	// Len returns number of elements in the cache.
+	Len() int
+}
+
+// elementMap is an alias for a map from a generic interface to a list.Element.
+type elementMap map[interface{}]*list.Element
+
+// entry represents a (key,value) pair entry in the Cache.
```
Roasbeef: Similar 80 character comments all throughout the PR.

neutrino.go
```diff
@@ -66,6 +67,10 @@ var (
 	// DisableDNSSeed disables getting initial addresses for Bitcoin nodes
 	// from DNS.
 	DisableDNSSeed = false
+
+	// DefaultFilterCacheSize is the number of filters neutrino will keep
```
Roasbeef: Perhaps we should instead target a size in memory usage? 

This would allow more precise tuning of the cache, as the size of each filter varies based on the composition of the block. In order to do this, we may need to instead create a new interface and use that as the input to the cache interface for each of the methods. The cache would store this type `T`. `T` has a single method, just to compute the size of the element serialized in bytes. This will allow the cache to prune based on size rather than number of elements.

query.go
```diff
@@ -71,6 +71,11 @@ type queryOptions struct {
 	persistToDisk bool
 }
 
+type filterCacheKey struct {
```
Roasbeef: Missing `godoc` comments.

query.go
```diff
@@ -638,6 +643,21 @@ checkResponses:
 	}
 }
 
+func (s *ChainService) getFilterFromCache(blockHash *chainhash.Hash,
```
Roasbeef: Missing `godoc` comments. Also the method could use a bit more "air" for breathability.

query.go
```diff
@@ -638,6 +643,21 @@ checkResponses:
...
+	filterType filterdb.FilterType) *gcs.Filter {
+	cacheKey := filterCacheKey{blockHash: blockHash, filterType: filterType}
+	if cachedFilter, ok := s.FilterCache.Get(cacheKey).(*gcs.Filter); ok {
+		return cachedFilter
+	}
+	return nil
+}
+
+func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
```
Roasbeef: Missing `godoc` comments.

lru/lru.go
```diff
@@ -0,0 +1,100 @@
+package lru
+
+import "container/list"
+import "sync"
+
+// Cache represents a generic LRU cache with a very simple interface.
+type Cache interface {
+	// Put stores the given (key,value) pair, replacing existing value if key
+	// already exists.
+	Put(key, value interface{})
```
Roasbeef: Operations should be able to fail, hence they should all return an `error`. This allows future implementations to do things like I/O which may fail.

query_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func genRandomBlockHash() *chainhash.Hash {
+	var seed [32]byte
+	rand.Read(seed[:])
+	hash := chainhash.Hash(seed)
+	return &hash
+}
+
+// Taken from filterdb/db_test.go
```
Roasbeef: Missing proper `godoc`.

neutrino.go
rawtxapp: yes, that makes a lot of sense.

so essentially the `size` in LRUCache becomes number of bytes instead of elements.

when a new element is inserted, we check that it's size isn't bigger than the size of the cache, otherwise evict as many filters as necessary to be able to put this one in (of course there is the issue that we could evict many smaller filters instead of a single big filter, but different tuning could be used later I guess).

one thing I don't really understand is (I haven't written a lot of go code...) why do we need a separate interface with the `ComputeSize` method ? wouldn't len(filter.filterData) work ? or filter could have a `Size` method ?

Roasbeef:
With the interface we don't need to leak the details of what's being stored
to the cache. We also eventually have plans to use this to store blocks as
well.

Roasbeef:
```golang
type CacheEntry interface {
    ComputeSize() uint64
}
```

lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// Put inserts a given (key,value) pair into the cache, if the key already
+// exists, it will replace value and update it to be most recent item in cache.
+func (c *LRUCache) Put(key, value interface{}) error {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	// key already exists, replace value.
```
halseth: nit: Start sentences with capital letter.

lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+// stores entries which let us get the cache key when an entry is evicted.
+type entry struct {
+	key   interface{}
+	value interface{}
+}
+
+// LRUCache provides a generic thread-safe lru cache that can be used for
+// storing filters, blocks, etc.
+type LRUCache struct {
+	// size is the capacity of the cache (ie not the number of elements in it).
```
halseth: should be named `capacity` to avoid the confusion.

lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Get will return value for a given key, making the element the most recently
+// accessed item in the process. Will return nil if the key isn't found.
+func (c *LRUCache) Get(key interface{}) (interface{}, error) {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	if el, ok := c.cache[key]; ok {
+		c.ll.MoveToFront(el)
```
halseth: suggestion: make happy-path the last returned, i.e. check `!ok` here and return nil in that case (and include a comment) :)

lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+		return el.Value.(*entry).value, nil
+	}
+	return nil, nil
+}
+
+// Len returns number of elements in the cache.
+func (c *LRUCache) Len() int {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
```
halseth: nit: extra nee line after defer

neutrino.go
halseth: A `ComputeSize` interface is preferred since the cache can then be used for arbitrary types :) I agree that `Size` is a better name.

query.go
```diff
@@ -638,6 +644,31 @@ checkResponses:
 	}
 }
 
+// getFilterFromCache returns a filter from ChainService's FilterCache if it
+// exists, returning nil and error if it doesn't.
+func (s *ChainService) getFilterFromCache(
```
halseth: split line like this:
```go
func (s *ChainService) getFilterFromCache(blockHash *chainhash.Hash, 
        filterType filterdb.FilterType) (*gcs.Filter, error)
```
and add a new line before `cacheKey` maybe?

query.go
```diff
@@ -638,6 +644,31 @@ checkResponses:
...
+	blockHash *chainhash.Hash, filterType filterdb.FilterType) (*gcs.Filter, error) {
+	cacheKey := filterCacheKey{blockHash: blockHash, filterType: filterType}
+
+	filter, err := s.FilterCache.Get(cacheKey)
+	if err != nil {
+		return nil, err
+	}
+
+	if cachedFilter, ok := filter.(*gcs.Filter); ok {
```
halseth: again suggestion to do early returns in case of error cases

query.go
```diff
@@ -638,6 +644,31 @@ checkResponses:
...
+		return cachedFilter, nil
+	}
+
+	return nil, fmt.Errorf("cache doesn't contain filter")
+}
+
+// putFilterToCache inserts a given filter in ChainService's FilterCache.
+func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
+	filterType filterdb.FilterType, filter *gcs.Filter) error {
+	cacheKey := filterCacheKey{blockHash: blockHash, filterType: filterType}
```
halseth: add a newline

query.go
```diff
@@ -661,9 +692,15 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 	getHeader := s.RegFilterHeaders.FetchHeader
 	dbFilterType := filterdb.RegularFilter
 
-	// First check the database to see if we already have this filter. If
+	// First check the cache to see if we already have this filter. If
 	// so, then we can return it an exit early.
-	filter, err := s.FilterDB.FetchFilter(&blockHash, dbFilterType)
+	filter, err := s.getFilterFromCache(&blockHash, dbFilterType)
+	if err == nil && filter != nil {
```
halseth: We should return if the errors are unexpected. (as in `ErrFilterNotFound` is an expected error from `FetchFilter` below). Same for `FilterBD.FetchFilter`.

query.go
```diff
@@ -758,15 +795,22 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
 
-	// If we've found a filter, write it to the database for next time.
 	if filter != nil {
-		err := s.FilterDB.PutFilter(&blockHash, filter, dbFilterType)
-		if err != nil {
-			return nil, err
-		}
+		// If we found a filter, put it in the cache and persistToDisk if
+		// the caller requested it.
+		s.putFilterToCache(&blockHash, dbFilterType, filter)
```
halseth: check returned error

query_test.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+
+	return filter, nil
+}
+
+// TestFilterCacheStorage verifies the interaction of filters with the
+// lru cache. It verifies that the cache returns the correct filters
+// for a given block and that it evicts older filters.
+func TestFilterCacheStorage(t *testing.T) {
```
halseth: Good test! 👍

query.go
rawtxapp: added this as a separate commit for now, wouldn't failing `GetCFilter` for an error in `FilterDB` or `Cache` (which are not *that* critical) lead to smaller bugs in those 2 systems becoming bigger bugs in neutrino ?

query.go
rawtxapp: what do you suggest to do on error ?

query.go
rawtxapp: Well the thing is that GetCFilter's goal is to get a filter by whatever means, the local db and cache is just a way of making it faster, I don't think any failure in them is critical, sure it would make getting filters slower, but the other option would be not getting anything.

For example, I could imagine iOS killing the app in the middle of a run and corrupting filterdb, that doesn't mean GetCFilter should stop working, maybe we could log it instead ?

Either way, I left the code returning the error if it's anything else than ErrElementNotFound and ErrFilterNotFound, but I think this is less than ideal.

query.go
rawtxapp: Like I said above, I don't think filterdb and cache are critical components. By this point, we already have a filter, why fail the call ? Insertion into cache could fail for different reasons. For example, what if the filter is bigger than the cache's size (unlikely, but still...), that would mean that GetCFilter would never be able to return that filter.

query_test.go
rawtxapp: This test is a bit outdated since the Size changes, I will rewrite it with additional cases once we agree on the interface for Size, etc.

rawtxapp:
Thanks for the suggestions. I moved lru to `cache/lru` and cache is now in `cache`, I was able to remove all the prefixes golang didn't like (ex: `CacheValue` now becomes `cache.Value`, etc).

I also did the `CachableFilter` wrapping, I think this is probably better than giving a `getSize` function to the constructor since this will be type safe.

Also replied to the comments about error handling.

neutrino.go
```diff
@@ -68,9 +68,9 @@ var (
 	// from DNS.
 	DisableDNSSeed = false
 
-	// DefaultFilterCacheSize is the number of filters neutrino will keep
-	// in memory if no size is specified in the neutrino.Config.
-	DefaultFilterCacheSize = 200
+	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino will
+	// keep in memory if no size is specified in the neutrino.Config.
+	DefaultFilterCacheSize uint64 = 4096
```
halseth: How many filters fit in 4 KB?

cache/lru/lru.go
```diff
@@ -0,0 +1,164 @@
...
+// exceed that given capacity.
+func NewCache(capacity uint64) *Cache {
+	return &Cache{
+		capacity: capacity,
+		ll:       list.New(),
+		cache:    make(elementMap),
+	}
+}
+
+// evict will evict as many elements as necessary where combined size of
```
halseth: This comment is hard to understand.

cache/lru/lru.go
```diff
@@ -0,0 +1,164 @@
...
+// elements worth size 1 and evict(2) was called, it would remove 2 in total.
+func (c *Cache) evict(sizeTotal uint64) error {
+	if sizeTotal > c.capacity {
+		return fmt.Errorf(
+			"can't evict %v elements in size, since capacity is %v",
+			sizeTotal, c.capacity)
+	}
+
+	evicted := uint64(0)
+	for evicted < sizeTotal {
```
halseth: sprinkle some comments in here :)

cacheable_filter.go
```diff
@@ -0,0 +1,18 @@
+package neutrino
```
halseth: I think it makes more sense to add this to the `cache` package.

query.go
halseth: Good point. I'm fine with just logging the error, if we make sure we are not relying on being able to add to the cache :)

halseth:
Also needs a rebase!

(After the comments are addressed I think this is a good point to squash commits together again for easier review) 😄

cache/lru/lru.go
rawtxapp: rewrote it, let me know if it's still unclear.

rawtxapp:
Rebased, and rewrote the commits in the same original order/size they were in before `Size`ing. Let me know what you think. Thanks!

cache/cacheable_filter.go
```diff
@@ -0,0 +1,18 @@
...
+
+// CacheableFilter is a wrapper around Filter type which provides a Size method
+// used by the cache to target certain memory usage.
+type CacheableFilter struct {
+	*gcs.Filter
+}
+
+// Size returns size of this filter in bytes.
+func (c *CacheableFilter) Size() (uint64, error) {
+	f, err := c.Filter.NBytes()
```
halseth: I'm not sure if we should use `Bytes`, `PBytes `, `NBytes` or `NPBytes` here, or something else entirely. One thing to think about is that these all serializes the `Filter`, which might be expensive, but I think it's better to optimize this at a later point if it turns out to be a bottleneck.

I'm leaning towards just using `Bytes` for simplicity.

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+	if vs > c.capacity {
+		return fmt.Errorf(
+			"can't insert entry of size %v into cache with capacity %v",
+			vs, c.capacity)
+	}
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	// If the filter already exists, remove it and decrease cache's size.
```
halseth: s/filter/element

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+		es, err := el.Value.(*entry).value.Size()
+		if err != nil {
+			return fmt.Errorf(
+				"couldn't determine size of existing cache value %v", err)
+		}
+		c.ll.Remove(el)
+		c.size -= es
+	}
+
+	// Then we need to make sure we have enough space for the filter, evict
```
halseth: here and other places: remove "filter" reference.

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+	// filters if we need more space.
+	if c.size+vs > c.capacity {
```
halseth: `evict` should be change to take the size of the filter to insert. That way the `if` statement can be removed and this can be simplified to
```go
if err := c.evict(vs); err != nil {
```

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+		ll:       list.New(),
+		cache:    make(elementMap),
+	}
+}
+
+// evict will evict as many elements as necessary to make enough space for a new
+// filter to be inserted. sizeTotal is the amount of space evict needs to make.
+// It could end up evicting more than sizeTotal if the size of the least
+// recently used element to evict was bigger than the size needed to evict.
+func (c *Cache) evict(sizeTotal uint64) error {
```
halseth: Letting `evict` take the size of the element to insert (see comment below), it would go something like this:
```go
func (c *Cache) evict(needed uint64) error {
    room := c.capacity - c.size
    for room < needed {
        ... remove one element ...
    }
```

cache/lru/lru_test.go
```diff
@@ -0,0 +1,196 @@
...
+	one := getSizeableValue(c.Get(1))
+	two := getSizeableValue(c.Get(2))
+	assertEqual(t, one, 1, "")
+	assertEqual(t, two, 2, "")
+	assertEqual(t, c.Len(), 2, "")
+}
+
+// TestReplacingElementValueBiggerSize tests that if an existing element is
+// replaced with a value of size bigger, that it evicts accordingly.
+func TestReplacingElementValueBiggerSize(t *testing.T) {
```
halseth: Good set of tests! 👍

neutrino.go
```diff
@@ -66,6 +67,10 @@ var (
 	// DisableDNSSeed disables getting initial addresses for Bitcoin nodes
 	// from DNS.
 	DisableDNSSeed = false
+
+	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino will
+	// keep in memory if no size is specified in the neutrino.Config.
+	DefaultFilterCacheSize uint64 = 4096 * 1000
```
halseth: Would be nice to include some back of the envelope calculation of how many filters will fit.

neutrino.go
```diff
@@ -586,6 +596,12 @@ func NewChainService(cfg Config) (*ChainService, error) {
 		return nil, err
 	}
 
+	filterCacheSize := DefaultFilterCacheSize
+	if cfg.FilterCacheSize != 0 {
+		filterCacheSize = cfg.FilterCacheSize
+	}
+	s.FilterCache = lru.NewCache(uint64(filterCacheSize))
```
halseth: No reason to cast to `uint64`.

query.go
```diff
@@ -662,6 +670,34 @@ checkResponses:
...
+	filterType filterdb.FilterType) (*gcs.Filter, error) {
+
+	cacheKey := filterCacheKey{blockHash: blockHash, filterType: filterType}
+
+	filterValue, err := s.FilterCache.Get(cacheKey)
+	if err != nil {
+		return nil, err
+	}
+
+	cachedFilter, ok := filterValue.(*cache.CacheableFilter)
```
halseth: Just return `filterValue.(*cache.CacheableFilter).Filter` at this point. If the cache contains something that is not a filter we should panic.

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+	"container/list"
+	"fmt"
+	"sync"
+
+	"github.com/lightninglabs/neutrino/cache"
+)
+
+var (
+	// ErrElementNotFound is returned when element isn't found in the cache.
+	ErrElementNotFound = fmt.Errorf("unable to find element")
```
halseth: this should be defined in the `cache` package.

cache/lru/lru.go
```diff
@@ -0,0 +1,173 @@
...
+	el, ok := c.cache[key]
+	// Element not found in the cache.
+	if !ok {
+		return nil, ErrElementNotFound
+	}
+
+	// When the cache needs to evict a filter to make space for another one, it
+	// starts eviction from the back, so by moving this filter to the front,
+	// it's eviction is delayed because it's recently accessed.
+	c.ll.MoveToFront(el)
```
halseth: We must actually acquire a write lock before calling this (I encountered this running race condition tests).

cache/cacheable_filter.go
Roasbeef: FWIW, it's serialized on the wire as `NBytes`.

neutrino.go
rawtxapp: Sure, sounds like a good idea, though I don't have any good numbers on the average filter sizes, etc. @Roasbeef do you have some stats I could use ?

cache/lru/lru.go
rawtxapp: great catch! also added a test case to cover it.

cache/cacheable_filter.go
halseth: I'm fine with sticking to `NBytes` as it it just an estimate on mem usage anyway.

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+			"can't evict %v elements in size, since capacity is %v",
+			needed, c.capacity)
+	}
+
+	room := c.capacity - c.size
+	for room < needed {
+		// We still need to evict some more elements.
+		if c.ll.Len() == 0 {
+			// We should never reach here.
+			return fmt.Errorf(
```
halseth: nit: Wrap lines as 
```go
return fmt.Errorf("all elements got evicted, yet "+
        "still need to evict %v, likelihood of error "+
        "during size calculation", needed-room)
```

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+				needed-room)
+		}
+
+		// Find the least recently used item.
+		if elr := c.ll.Back(); elr != nil {
+			// Determine lru item's size.
+			ce := elr.Value.(*entry)
+			es, err := ce.value.Size()
+			if err != nil {
+				return fmt.Errorf(
```
halseth: same, wrap line

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+					"couldn't determine size of existing cache value %v", err)
+			}
+
+			// Account for that element's removal in evicted and cache size.
```
halseth: wrap all lines at 80 chars.

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+// stores entries which let us get the cache key when an entry is evicted.
+type entry struct {
+	key   interface{}
+	value cache.Value
+}
+
+// Cache provides a generic thread-safe lru cache that can be used for
+// storing filters, blocks, etc.
+type Cache struct {
+	// capacity represents how much this cache can hold. It could be number of
```
halseth: some of these lines are >80 chars.

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+			ce := elr.Value.(*entry)
+			es, err := ce.value.Size()
+			if err != nil {
+				return fmt.Errorf(
+					"couldn't determine size of existing cache value %v", err)
+			}
+
+			// Account for that element's removal in evicted and cache size.
+			c.size -= es
+			room = c.capacity - c.size
```
halseth: nit: this can be simplified to making the loop
```go
// We'll evict elements until the cache has room for an element 
// of size "needed".
for c.capacity - c.size < needed {
```

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Put inserts a given (key,value) pair into the cache, if the key already
+// exists, it will replace value and update it to be most recent item in cache.
+func (c *Cache) Put(key interface{}, value cache.Value) error {
+	vs, err := value.Size()
+	if err != nil {
+		return fmt.Errorf("couldn't determine size of cache value: %v", err)
+	}
+	if vs > c.capacity {
+		return fmt.Errorf(
```
halseth: same with wrapping and keeping lines < 80 chars here.

cache/lru/lru.go
```diff
@@ -0,0 +1,163 @@
...
+	return nil
+}
+
+// Put inserts a given (key,value) pair into the cache, if the key already
+// exists, it will replace value and update it to be most recent item in cache.
+func (c *Cache) Put(key interface{}, value cache.Value) error {
+	vs, err := value.Size()
+	if err != nil {
+		return fmt.Errorf("couldn't determine size of cache value: %v", err)
+	}
```
halseth: nit: newline after closing bracket

cache/lru/lru_test.go
```diff
@@ -0,0 +1,224 @@
...
+	assertEqual(t, len(c.cache), 1, "")
+	four := getSizeableValue(c.Get(4))
+	assertEqual(t, four, 4, "")
+
+	c = NewCache(6)
+	c.Put(1, &sizeable{value: 1, size: 1})
+	c.Put(2, &sizeable{value: 2, size: 2})
+	c.Put(3, &sizeable{value: 3, size: 3})
+	assertEqual(t, c.size, uint64(6), "")
+	// Insert element with size=capacity of cache.
```
halseth: nit: newline before comments

cache/lru/lru_test.go
```diff
@@ -0,0 +1,224 @@
...
+	one := getSizeableValue(c.Get(1))
+	assertEqual(t, one, 3, "")
+}
+
+// TestConcurrency is a very simple test that checks concurrent access to the
+// lru cache. When running the test, "-race" option should be passed to
+// "go test" command.
+func TestConcurrency(t *testing.T) {
+	t.Parallel()
+	c := NewCache(5)
```
halseth: Maybe add a few more concurrency test cases below. Suggestions:
* a cache of capacity 5, putting and getting 20 elements
* a cache of capacity 100, putting and getting 50 elements

cache/lru/lru_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	for i := 0; i < 5; i++ {
+		go func(i int) {
```
halseth: Add all goroutines to a waitgroup, then wait for them all to be done before finishing the test.

cache/lru/lru_test.go
```diff
@@ -0,0 +1,224 @@
...
+			c.Put(i, &sizeable{value: i, size: 1})
```
halseth: Check that no error is returned.

cache/lru/lru_test.go
```diff
@@ -0,0 +1,224 @@
...
+		}(i)
+	}
+
+	for i := 0; i < 5; i++ {
+		go func(i int) {
+			c.Get(i)
```
halseth: Check that no other error than `ErrElementNotFound` is returned.

cache/lru/lru.go
rawtxapp: hmm I'm not sure, this is 78 chars, and none of the lines below go over 80 chars, the longest is the comment `// evict will evict...` and it's exactly 80 chars, am I missing something ?

halseth: tabs are counted as 8 spaces

cache/lru/lru.go
```diff
@@ -0,0 +1,160 @@
...
+		ll:       list.New(),
+		cache:    make(elementMap),
+	}
+}
+
+// evict will evict as many elements as necessary to make enough space for a new
+// element with size needed to be inserted.
+func (c *Cache) evict(needed uint64) error {
+	if needed > c.capacity {
+		return fmt.Errorf(
```
halseth: wrapping

cache/lru/lru.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Get will return value for a given key, making the element the most recently
+// accessed item in the process. Will return nil if the key isn't found.
+func (c *Cache) Get(key interface{}) (cache.Value, error) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	el, ok := c.cache[key]
+	// Element not found in the cache.
```
halseth: nit: move inside `if`

cache/lru/lru.go
rawtxapp: oh I see, I had 4 space for tabs, makes sense now.

neutrino.go
Roasbeef: Not blocking for this PR, as I can add on later, but last I checked, the median filter size was ~14KB.

PR #74, opened 2018-07-14
multi: update usage of filters to match latest version of BIP 158 

Roasbeef:
In this PR, we update the rescan and syncing logic to adhere to the new format of filters defined within BIP 158. The major changes are in the area of the rescan abstraction as the filters no longer include the previous output, so we're forced to introduce a new `InputWithScript` which packages the previous output script (that we match on), and also the outpoint (that we'll use to locate the matching transaction). 

The final major change is a re-write of the logic we use to determine which peer is presenting a faulty history. In the past, we were able to simply fetch the block, reconstruct the filter and ban the peers which report incorrect filter hashes. Since the filters now include the previous output scripts, this is no longer an option. Instead, we're now forced to download all filters from each peer, and then ensure that the filters contain _most_ of the items they should. I say most as we can only verify that the filters contain the set of output scripts found within a block. In order to return to the strong veification that was possible in the past, we'll need to introduce a new variant of block messages that also include the et of previous output scripts.


rescan.go
```diff
@@ -907,14 +892,13 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
...
 	// As this is meant to fetch UTXO's, the options MUST specify at least
-	// a single outpoint.
-	if len(ro.watchOutPoints) != 1 {
-		return nil, fmt.Errorf("must pass exactly one OutPoint")
-	}
-	watchList := [][]byte{
-		builder.OutPointToFilterEntry(ro.watchOutPoints[0]),
-		ro.watchOutPoints[0].Hash[:],
+	// a single input.
+	if len(ro.watchInputs) != 1 {
```
wpaulino: Nit: comment should be updated to no longer say "at least".

rescan.go
```diff
@@ -936,6 +920,8 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
 			ro.startBlock.Hash = chainhash.Hash{}
 		}
 	}
+
+	// Alternatively, our earliest possible block by actually be a height.
```
wpaulino: nit: s/by/might/

Issue #73, opened 2018-07-12
Build errors from BuildBasicFilter


dmp1ce:
I just checked out out master, got dependencies and tried to build with `go build`. I received these errors:

```
> go build
# github.com/lightninglabs/neutrino/filterdb
../../lightninglabs/neutrino/filterdb/db.go:102:47: not enough arguments in call to builder.BuildBasicFilter
        have (*wire.MsgBlock)
        want (*wire.MsgBlock, [][]byte)
../../lightninglabs/neutrino/filterdb/db.go:119:21: undefined: builder.BuildExtFilter
../../lightninglabs/neutrino/filterdb/db.go:211:34: not enough arguments in call to gcs.FromNBytes
        have (number, []byte)
        want (uint8, uint64, []byte)
# github.com/lightninglabs/neutrino/headerfs
../../lightninglabs/neutrino/headerfs/store.go:529:48: not enough arguments in call to builder.BuildBasicFilter
        have (*wire.MsgBlock)
        want (*wire.MsgBlock, [][]byte)
../../lightninglabs/neutrino/headerfs/store.go:545:22: undefined: builder.BuildExtFilter
```

Did btcutil have an update?

keymone:
you need to install and run glide to fetch correct dependencies into `vendor/`

btw, is there any interest to switch from glide to dep? that seems to be the new canonical tool.

Roasbeef:
Yep you need to use glide. 

Also yep we plan to switch to dep, just hasn't been done yet. PR's accepted!

PR #75, opened 2018-07-14
Convert from glide to dep

keymone:
From discussion in #73 

- dep init automatically converts glide.yaml
- use dep in gotest.sh
- use HEAD gometalinter (v1 and v2 don't install?)


halseth:
We now use `go modules`.

Issue #68, opened 2018-07-10
rescan: update filter fetching interfaces to allow for batch filter fetching


Roasbeef:
The finalized version of `BIP 157` [allows nodes to query for a _batch_ of filters in a single message](https://github.com/lightninglabs/neutrino/blob/36d9a509ff4ea1e916a73c6f1c2b784c1faf11ed/query.go#L319). This is useful as if we're catching up from a few thousand blocks, we can request them all in a _single_ message, and get the responses pipelined, rather than the back and forth of repeatedly fetching _each_ filter individually. 

Currently, this isn't exposed to the caller, nor utilized internally for rescan (or `GetUtxo`) purposes.

# Steps To Completion 

- [ ] Modify the `GetCfilters` method to expose both a start height and stop hash just as in the p2p message. To start with, one can convert all the existing calls to use the height of the filter to be requested, and use the stop hash of that very same header. 

- [ ] Where relevant ([in the rescan loop for example](https://github.com/lightninglabs/neutrino/blob/master/rescan.go#L288)) start fetching filters in batch, rather than fetching them individually. One way to do this is at the top of the loop we can check if our starting height is very far behind the height of the main node. If so, we can likely fetch the filters using a single batch query.

- [ ] Update the `GetUTXO` method to also fetch the filters in batch. This is likely where we'll see an even bigger savings as outside of the regular rescan loop, we may be starting far-ish in the past. 

The above changes will likely also affect `btcsuite/btcwallet` and a similar set of optimizations during it's initial chain sync, and also the HD seed import rescan can take advantage of this optimization.

rawtxapp:
This looks straightforward, I'd like to try to do it if no one is working on it.

If yes, what do you think of using a "predicted cache" inside `GetCFilter`, so something like:
- if filter 50 is requested, the function could fetch filters 30-to-70 or possibly checkpoint-to-checkpoint and store them in a cache inside itself, so that on the next call to it, it could return immediately, presumably the marginal cost of fetching extra filters is very low.
- if the caller wants the flexibility, it could use queryoptions to set start height and stop hash, or it could potentially just hint towards which direction it's scanning and let GetCFilter take care of caching (for example: I'm scanning backwards, give me filter 50, GetCFilter would fetch filters 20-to-50).

That way, the logic in the callers wouldn't need to change and GetCFilter could try different batch sizes to optimize network/CPU/memory. But you know better, what do you think ?

Roasbeef:
[...]

PR #76, opened 2018-07-16
early draft: batch filter fetch

rawtxapp:
This is just a draft to see if it's headed in the right direction.

A few questions I have is:
- should the return type from GetCFilter be an array of filters or a map of hash->filter ?
- should the same query be dispatched to one peer at a time until answered or broken up into multiple pieces and sent to different peers ? for example, say caller asks for filters 100-to-200, do we send that request to a peer, or should we do 100-to-150 from 1 peer and 150-to-200 from another peer ?
- I noticed that there are some functions are quite large, getcfilter itself is >100lines, it's relatively easy to follow, but difficult to unit-test individual parts, just wondering if integration tests are considered good enough ?

github context: https://github.com/lightninglabs/neutrino/issues/68

### DIFF
--- a/query.go
+++ b/query.go
@@ -696,16 +696,17 @@ func (s *ChainService) GetCFilter(filterType wire.FilterType,
 	// In addition to fetching the block header, we'll fetch the filter
 	// headers (for this particular filter type) from the database. These
 	// are required in order to verify the authenticity of the filter.
-	curHeader, err := getHeader(&stopHash)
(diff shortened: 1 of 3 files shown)


rawtxapp:
On a second thought, I think the feedback back and forth would probably take a lot of time on this change since everybody is quite busy. So I will close this and instead focus on bugfixes and smaller changes. Thanks.

PR #78, opened 2018-07-26
rescan: fix bug that could cause rescan loop to miss blocks

Roasbeef:


blockmanager.go
```diff
@@ -1744,6 +1744,19 @@ func (b *blockManager) FilterHeaderTip() uint32 {
...
 }
 
+// SynchronizeFilterHeaders allows the caller to execute a function closure
+// that depends on synchronization with the current set of filter headers. This
+// allows the caller to execute an action that depends on the current filter
+// header state, thereby ensuring that the state would shift from underneath
+// them. Each execution of the closure will have the current filter header tip
+// passed in to ensue that the caller gets a consistent view.
+func (b *blockManager) SynchronizeFilterHeaders(f func(uint32) error) error {
+	b.newHeadersMtx.RLock()
```
halseth: should be `newFilterHeadersMtx`?

Roasbeef: Yep! switched over and things broke, working on getting it back to a workable state...

blocksubscriptions.go
```diff
@@ -85,22 +90,71 @@ func sendMsgToSubscriber(sub *blockSubscription, bm *blockMessage) {
 }
 
 // subscribeBlockMsg handles adding block subscriptions to the ChainService.
-// TODO(aakselrod): move this to its own package and refactor so that we're
-// not modifying an object held by the caller.
-func (s *ChainService) subscribeBlockMsg(onConnectBasic, onConnectExt,
-	quit <-chan struct{}) *blockSubscription {
+// The best known height to the caller should be passed in, such that we can a
```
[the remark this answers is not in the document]
Roasbeef: Fixed!

blocksubscriptions.go
```diff
@@ -35,33 +40,50 @@ type blockSubscription struct {
 
 // sendSubscribedMsg sends all block subscribers a message if they request this
 // type.
+//
 // TODO(aakselrod): Refactor so we're able to handle more message types in new
 // package.
 func (s *ChainService) sendSubscribedMsg(bm *blockMessage) {
-	var subChan chan<- wire.BlockHeader
+
```
halseth: nit: could use `defer` to unlock

blocksubscriptions.go
```diff
@@ -35,33 +40,50 @@ type blockSubscription struct {
...
-			case <-sub.intQuit:
-			}
-		}
+		sendMsgToSubscriber(sub, bm)
 	}
 	s.mtxSubscribers.RUnlock()
 }
 
+// sendMsgToSubscriber is a helper function that sends the target message to
+// the subscription client over the proper channel based on the type of the new
```
halseth: godoc says it "sends [..] over the proper channel", but it is always sent on the `notifyBlocks` channel.  

Actually looks like the `selectChan` method that handles this block message later will handle the case where the sub channel is `nil`, so this confusing method could be simplified significantly. no blocker for this pr though, as this is equivalent to what was already here.

blocksubscriptions.go
```diff
@@ -27,75 +35,138 @@ type blockMessage struct {
...
-	return &subscription
+
+	// At this point, we'll now check to see if we need to deliver any
+	// backlog notifications as its possible that while the caller is
+	// requesting right after a new set of blocks has been connected.
+	err := s.blockManager.SynchronizeFilterHeaders(func(filterHeaderTip uint32) error {
+		s.mtxSubscribers.Lock()
+		defer s.mtxSubscribers.Unlock()
+
+		s.blockSubscribers[&subscription] = struct{}{}
```
cfromknecht: Does this need to be done inside synchronize filter headers? Is it possible to do:
 1. create `subscription`
 2. spawn `subscriptionHandler`
3.` SychnronizeFilterHeaders`
4. If successful:
  4.1 acquire `mtxSubscribers` and add to `blockSubscribers`

halseth: I think so, but I believe you would need to acquire `mtxSubscribers` inside `SychnronizeFilterHeaders`, since else you risk  that a new block comes in before you are able to add the new subscription.

Roasbeef: Yep what @halseth said.

PR #81, opened 2018-08-07
utxoscanner: prevent send on close channel + update glide

cfromknecht:
I did a big no-no, and realized that this channel
close is not necessary since we already cache the
result.


cfromknecht:
I've also amended the PR to point to latest btcd, which fixes some syncing issues that cause the tests to timeout

PR #83, opened 2018-08-09
Fix lnd crash when restarting it

yaslama:


Roasbeef:
Any explanation? Or a commit message body?

yaslama:
Before the lines which are changed in this commot, there is a test "if startHeight > curHeight+1" which returns true, thus there is only one case when startHeight - curHeight - 1 is not negative: when startHeight==curHeight+1, that is when we already have the headers up to the 1000 boundary (1000n where n>=0). And this case doesn't need any specific handling.
But the problem occurs when we have already for instance headers for 1000n+500 bocks, and we restart lnd, it will receive a batch with headers from 1000n to 1001n and offset will be negative which causes r.FilterHashes[offset:] to crash.
In order to handle this case (where we already have some headers in the interval), we need to have an offset of curHeight + 1 - startHeight which is the index of the first header we need from this batch.
```
 2018-08-08 13:51:55.339 [INF] BTCN: Fetching set of checkpointed cfheaders filters from height=1381157, hash=1d4eccbea1ec3f47ee82335af64a3382e372fab0166ac0f4892c6dca79349028
 2018-08-08 13:51:55.339 [INF] BTCN: Starting to query for cfheaders from checkpoint_interval=1381
 2018-08-08 13:51:55.339 [INF] BTCN: Attempting to query for 1 cfheader batches
 panic: runtime error: slice bounds out of range
 
 goroutine 34 [running]:
 /vendor/github.com/lightninglabs/neutrino.(*blockManager).getCheckpointedCFHeaders.func1(0x44200f0d20, 0x6f271a4ce0, 0x44203ca0c0, 0x6f271a4aa0, 0x443168bf80, 0x0)
 	/vendor/github.com/lightninglabs/neutrino/blockmanager.go:832 +0xa44
 /vendor/github.com/lightninglabs/neutrino.(*ChainService).queryBatch(0x44200f2c80, 0x4431808000, 0x1, 0x566, 0x4420289e30, 0x4420294540, 0x0, 0x0, 0x0)
 	/vendor/github.com/lightninglabs/neutrino/query.go:379 +0x7f4
 /vendor/github.com/lightninglabs/neutrino.(*blockManager).getCheckpointedCFHeaders(0x44201a0480, 0x44317d8000, 0x566, 0x566, 0x44205be068, 0x500)
 	/vendor/github.com/lightninglabs/neutrino/blockmanager.go:740 +0x6a8
 /vendor/github.com/lightninglabs/neutrino.(*blockManager).cfHandler.func2(0x44312538b0, 0x44201a0480, 0x151672, 0x4431640b00, 0x0, 0x6f27194680)
 	/vendor/github.com/lightninglabs/neutrino/blockmanager.go:475 +0x338
 created by /vendor/github.com/lightninglabs/neutrino.(*blockManager).cfHandler
 	/vendor/github.com/lightninglabs/neutrino/blockmanager.go:415 +0x338
```

aakselrod:
utACK for now, will test after another sync issue.

aakselrod:
Sync tests pass.

PR #85, opened 2018-08-14
LRU block cache

halseth:
This PR builds on top of #72, adding an in-memory block cache for block received from peers.


Roasbeef:
Superseded by #88.

PR #87, opened 2018-08-21
utxoscanner: break get utxo

cfromknecht:
This PR adds the ability to cancel GetUtxo
requests via two means:
 * a configurable cancel channel
 * automatically when shutting down the utxoscanner

The first is accomplished by selecting on an extra
cancel, which returns ErrGetUtxoCancelled if signaled
before the result has arrived.

The second iterates through all requests which have
been queued but not started, and cancels them all
with ErrGetUtxoCancelled. This should enable to
dependent systems to break out of long-running
queries, especially during shutdown.

### DIFF
--- a/utxoscanner.go
+++ b/utxoscanner.go
@@ -55,7 +55,7 @@ func (r *GetUtxoRequest) deliver(report *SpendReport, err error) {
 }
 
 // Result is callback returning either a spend report or an error.
-func (r *GetUtxoRequest) Result() (*SpendReport, error) {
+func (r *GetUtxoRequest) Result(cancel <-chan struct{}) (*SpendReport, error) {
 	r.mu.Lock()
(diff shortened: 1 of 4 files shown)


utxoscanner.go
```diff
@@ -144,14 +147,24 @@ func (s *UtxoScanner) Stop() error {
...
 		select {
 		case <-s.shutdown:
-			return nil
+			break batchShutdown
 		case <-time.After(50 * time.Millisecond):
 			s.cv.Signal()
 		}
 	}
+
+	// Cancel all pending pending get utxo requests that were not pulled
```
halseth: pending pending?

cfromknecht: Fixed

cfromknecht:
Comments addressed, PTAL

PR #92, opened 2018-08-28
Fix rescan update deadlock

halseth:
This PR fixes an issue within the current `neutrino` rescan logic, which would cause a send on the `updateChan` to block indefinitely. 

This was caused by a caller of `Update` killing the current rescan at the same time, making the update never being read from the channel.

This PR fixes that by ensuring an error is returned if the rescan exits while the update is being enqueued, moving the responsibility to handle this case to the caller.

NOTE: The caller in this case is `btcwallet`, which must be updated to handle the returned error.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -987,7 +987,7 @@ txOutLoop:
 // replacement for the btcd rescan and notification functionality used in
 // wallets. It only contains information about whether a goroutine is running.
 type Rescan struct {
-	running    uint32
+	running    chan struct{}
 	updateChan chan *updateOptions
 
 	options []RescanOption
@@ -1005,7 +1005,7 @@ type Rescan struct {
 // which returns any error on termination of the rescan process.
 func (s *ChainService) NewRescan(options ...RescanOption) *Rescan {
 	return &Rescan{
-		running:    1,
+		running:    make(chan struct{}),
 		options:    options,
 		updateChan: make(chan *updateOptions),
 		chain:      s,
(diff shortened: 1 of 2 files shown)


halseth:
Travis is currently failing, the added test seems to interact with the existing.

rescan.go
```diff
@@ -277,20 +277,36 @@ func (s *ChainService) Rescan(options ...RescanOption) error {
...
+			default:
+			}
 		}
+		s.blockManager.newFilterHeadersMtx.Unlock()
+		close(done)
+	}()
+
+	// Now wait for either filter headers are updated, or we are quitting.
+	select {
+	case <-done:
```
cfromknecht: @Roasbeef pointed out that we should also be pulling in updates while waiting for the sync to finish. Rn the sync is in the bg, but sends on `updateChan` will still block. Suggested that they can be queued up, and then applied after the sync completes

rescan.go
```diff
@@ -1085,8 +1109,15 @@ func DisableDisconnectedNtfns(disabled bool) UpdateOption {
...
-	if running != 1 {
+
+	uo := defaultUpdateOptions()
+	for _, option := range options {
+		option(uo)
+	}
+
+	select {
+	case r.updateChan <- uo:
+	case <-r.running:
```
cfromknecht: also select on `ro.quit` in case caller specified?

rescan.go
halseth: Good idea! Will add

rescan.go
halseth: Yep :)

halseth:
Comments addressed, PTAL!

Roasbeef:
Needs a rebase. With https://github.com/btcsuite/btcwallet/pull/535 and this, I don't think that https://github.com/btcsuite/btcwallet/pull/533 is actually needed.

PR #98, opened 2018-09-17
Catch up filter headers and block headers in parallel

halseth:
Catch up filter headers and block headers in parallel. 

Also includes a fix for rollback of filter headers, and a proper fix to what was attempted in #95.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -123,14 +123,14 @@ type blockManager struct {
 	// switch over.
 	newHeadersSignal *sync.Cond
 
-	// filterHeaderTip will be set to the current filter header tip at all
-	// times.  Callers MUST hold the lock below each time they read/write
-	// from this field.
+	// filterHeaderTip will be set to the height of the current filter
+	// header tip at all times.  Callers MUST hold the lock below each time
+	// they read/write from this field.
 	filterHeaderTip uint32
 
 	// filterHeaderTipHash will be set to the current block hash of the
-	// fitler header tip at all times.  Callers MUST hold the lock below
-	// each time they read/write from this field.
+	// block at height filterHeaderTip at all times.  Callers MUST hold the
+	// lock below each time they read/write from this field.
 	filterHeaderTipHash chainhash.Hash
 
 	// newFilterHeadersMtx is the mutex that should be held when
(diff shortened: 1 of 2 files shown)


blockmanager.go
```diff
@@ -395,15 +395,23 @@ func (b *blockManager) cfHandler() {
 		b.wg.Done()
 	}()
 
-	log.Infof("Waiting for block headers to sync, then will start " +
-		"cfheaders sync...")
+	b.newFilterHeadersMtx.RLock()
+	filterHeaderTip := b.filterHeaderTip
+	filterHeaderTipHash := b.filterHeaderTipHash
```
Roasbeef: I don't think these local variables are actually needed as this is the only goroutine that can mutate these variables.

halseth: Removed.

blockmanager.go
```diff
@@ -402,15 +402,21 @@ func (b *blockManager) cfHandler() {
...
+	log.Infof("Waiting for block headers to sync, then will start "+
+		"cfheaders sync from height %v...", b.filterHeaderTip)
 
-	// We'll wait until the main header sync is mostly finished before we
-	// actually start to sync the set of cfheaders. We do this to speed up
-	// the sync, as the check pointed sync is faster, then fetching each
-	// header from each peer during the normal "at tip" syncing.
 	b.newHeadersSignal.L.Lock()
-	for !b.IsCurrent() {
+	for !(b.IsCurrent() || b.filterHeaderTip+wire.CFCheckptInterval <= b.headerTip) {
```
cfromknecht: already discussed irl, but it might be worth reversing the order (again) to avoid db hits amap

halseth: reversed.

halseth:
Now waits 100 CF intervalse before starting catch up.

halseth:
Reverted back to waiting only one interval. Instead the latest commit will fetch the filter checkpoints up to the latest block checkpoint, avoiding doing this fetch every time the block headers advance. This means that we will reuse these checkpoints until `blockeaders > chain_checkpoint`, at which point we will fetch new checkpoints.

Roasbeef:
Tested this locally with multiple syncs from both a remote and local node, didn't run into any major issues. Ensured that it was able to proceed after forced restarts, and seemed to handle the started no problem. However, I think there're a few places where a quit signal isn't being properly threaded through, as I noticed after some attempts, I was unable to kill the process.

halseth:
When do you experience you are unable to kill the process? 

Haven't been able to reproduce it myself, but a hunch I had was that one of the loops  where we are waiting for the `newHeadersSignal` would stall. However, in the block manager stop method we will repeatedly signal until all goroutines are shut down, so it should be handling that case.

This PR shouldn't change the `quit` behaviour in any way, but a goroutine dump in the unable-to-kill case should expose the culprit.

halseth:
Rebased.

cfromknecht:
@halseth perhaps the source of the inability to quit while syncing is a result of using `Sleep`? Might suggest selecting on `time.After` and `quit` to ensure we can exit

PR #99, opened 2018-09-25
blockmanager: remove queryResponse length check

halseth:
queryResponses was recently changed from being a slice to a map.
Previously a length check was performed to make sure we didn't index out
of bounds, but after the change to using a map this is no longer
correct.

This would be apparent doing header download from multiple peers, as
responses would come asynchronously, and the headers wouldn't be written
because of this check.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -881,23 +881,17 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 				// last written set of cfheaders.
 				curHeight = checkPointIndex * wire.CFCheckptInterval
 
-				// Break if we've gotten to the end of the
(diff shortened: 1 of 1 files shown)


blockmanager.go
```diff
@@ -897,7 +891,7 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 
 				// We have another response to write, so delete
 				// it from the cache and write it.
-				queryResponses[checkPointIndex] = nil
+				delete(queryResponses, checkPointIndex)
```
cfromknecht: if we're actually removing the items, what do you think about having `queryResponses` check the `ok` value directly?

halseth: Agreed. Done.

Issue #100, opened 2018-10-08
initial sync, testnet, panic: attempt to write cfheaders out of order


AndrewSamokhvalov:
lnd version: Version 0.5.0-beta commit=a3edcf9cd537ea37a502efc91d262e7017a99e6b
btcd version: 2a560b2036bee5e3679ec2133eb6520b2f195213

```
2018-10-08 11:06:19.826 [INF] BTCN: Starting to query for cfheaders from checkpoint_interval=0
2018-10-08 11:06:19.839 [INF] BTCN: Attempting to query for 1432 cfheader batches
2018-10-08 11:06:19.839 [DBG] BTCN: Sending getcfheaders to 207.154.224.115:18333 (outbound)
2018-10-08 11:06:22.841 [DBG] BTCN: Sending getcfheaders to 207.154.224.115:18333 (outbound)
2018-10-08 11:06:23.008 [DBG] BTCN: Received cfheaders from 207.154.224.115:18333 (outbound)
2018-10-08 11:06:23.010 [DBG] BTCN: Got cfheaders from height=1 to height=1001
2018-10-08 11:06:23.014 [DBG] BTCN: Writing filter headers up to height=1000
2018-10-08 11:06:23.019 [INF] BTCN: Verified 1 filter header in the last 4m52.34s (height 1, 2011-02-03 02:22:08 +0300 MSK)
2018-10-08 11:06:23.019 [DBG] BTCN: Sending getcfheaders to 207.154.224.115:18333 (outbound)
2018-10-08 11:06:23.056 [DBG] BTCN: Received cfheaders from 207.154.224.115:18333 (outbound)
2018-10-08 11:06:25.352 [DBG] BTCN: Received cfheaders from 207.154.224.115:18333 (outbound)
2018-10-08 11:06:25.354 [DBG] BTCN: Got cfheaders from height=1001 to height=2001
panic: couldn't write cfheaders msg: attempt to write cfheaders out of order

goroutine 54 [running]:
github.com/lightningnetwork/lnd/vendor/github.com/lightninglabs/neutrino.(*blockManager).getCheckpointedCFHeaders.func1(0xc4200e0960, 0x1dbeea0, 0xc42048e3f0, 0x1dbec60, 0xc43102a900, 0x501)
	/Users/andrey/GoProjects/src/github.com/lightningnetwork/lnd/vendor/github.com/lightninglabs/neutrino/blockmanager.go:871 +0xb2d
[...]

PR #103, opened 2018-10-17
Add more query batch logging

halseth:
Adds more information to the logs in an attempt to pinpoint #100.

Also adds a new test that ensures the blockmanager can handle responses out of order, and partially written filter header intervals.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -765,7 +765,8 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 	// latest known checkpoint.
 	curHeader, curHeight, err := store.ChainTip()
 	if err != nil {
-		panic("getting chaintip from store")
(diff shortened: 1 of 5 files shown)


halseth:
comments addressed.

blockmanager.go
```diff
@@ -813,14 +814,8 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
 		if err != nil {
-			// Try to recover this.
-			select {
-			case <-b.quit:
-				return
-			case <-time.After(QueryTimeout):
-				currentInterval--
-				continue
-			}
+			panic(fmt.Sprintf("failed getting block header at "+
```
Roasbeef: This works for now, but we'll need to be able to recover and handle this in the future for the mainnet release of neutrino.

PR #111, opened 2018-11-09
OnVersion now supports returning a Reject Message to peer (which will…

jadeblaquiere:
… be disconnected after Reject message is sent)

see : https://github.com/btcsuite/btcd/commit/7b103e243413d3af03257649ca2185ae78f8eb6f

Instead of simply disconnecting, compose a disconnect message so the peer is given a reason for the disconnect.


halseth:
Needs a rebase!

jadeblaquiere:
Not sure what's going on with Travis builds... " fatal: unable to access 'https://go.googlesource.com/crypto/': The requested URL returned error: 502"

This seems to be an environmental failure...

PR #121, opened 2019-01-14
Ensure at least one peer is connected before starting cfheader sync.

roeierez:
This PR is a suggested way to reduce about 3 seconds (with default configuration) from neutrino startup time on some cases.
From within BlockManager the function cfHandler is called right on start regardless if there is at least one peer connected.
Since this function query the peers it causes the first query to fail with high probability and retry after a waiting period of QueryTimeout interval (3 seconds by default).
I have tested the startup on mobile devices and found that the first peer connection is usually established in less than 1 second, which creates a waste of 2-3 seconds.
My suggestion is to start the cfHandler after there is at least one peer, making sure the query has a target.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -250,7 +250,7 @@ func newBlockManager(s *ChainService) (*blockManager, error) {
 }
 
 // Start begins the core block handler which processes block and inv messages.
-func (b *blockManager) Start() {
+func (b *blockManager) Start(firstPeerConnect <-chan struct{}) {
 	// Already started?
 	if atomic.AddInt32(&b.started, 1) != 1 {
 		return
@@ -259,7 +259,22 @@ func (b *blockManager) Start() {
 	log.Trace("Starting block manager")
 	b.wg.Add(2)
 	go b.blockHandler()
-	go b.cfHandler()
+	go func() {
+		defer b.wg.Done()
+
+		log.Debug("Waiting for peer connection...")
+
+		// Before starting the cfHandler we want to make sure we are connected
+		// with at least one peer.
+		select {
+		case <-firstPeerConnect:
+		case <-b.quit:
+			return
+		}
+
+		log.Debug("Peer connected, starting cfHandler.")
+		b.cfHandler()
+	}()
 }
 
 // Stop gracefully shuts down the block manager by stopping all asynchronous
@@ -407,12 +422,7 @@ func (b *blockManager) handleDonePeerMsg(peers *list.List, sp *ServerPeer) {
 // run as a goroutine. It requests and processes cfheaders messages in a
 // separate goroutine from the peer handlers.
 func (b *blockManager) cfHandler() {
-	// If a loop ends with a quit, we want to signal that the goroutine is
-	// done.
-	defer func() {
-		log.Trace("Committed filter header handler done")
-		b.wg.Done()
-	}()
+	defer log.Trace("Committed filter header handler done")
 
 	var (
 		// allCFCheckpoints is a map from our peers to the list of
--- a/neutrino.go
+++ b/neutrino.go
@@ -530,6 +530,7 @@ type ChainService struct {
 	donePeers         chan *ServerPeer
 	banPeers          chan *ServerPeer
 	query             chan interface{}
+	firstPeerConnect  chan struct{}
 	peerHeightsUpdate chan updatePeerHeightsMsg
 	wg                sync.WaitGroup
 	quit              chan struct{}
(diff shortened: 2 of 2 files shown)


neutrino.go
```diff
@@ -1115,6 +1117,11 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
 		state.outboundPeers[sp.ID()] = sp
 	}
 
+	//close firstPeerConnect channel so blockManager will be notified
+	if s.firstPeerConnect != nil {
+		close(s.firstPeerConnect)
+		s.firstPeerConnect = nil
+	}
```
aakselrod:
```suggestion
	}
	
```

aakselrod:
Other than a small whitespace issue, this LGTM.

neutrino.go
roeierez: fixed

roeierez:
squashed.

neutrino.go
```diff
@@ -1115,6 +1117,12 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
 		state.outboundPeers[sp.ID()] = sp
 	}
 
+	// Close firstPeerConnect channel so blockManager will be notified
```
halseth: nit: period at end.

blockmanager.go
```diff
@@ -259,7 +259,16 @@ func (b *blockManager) Start() {
 	log.Trace("Starting block manager")
 	b.wg.Add(2)
 	go b.blockHandler()
-	go b.cfHandler()
+	go func() {
+		defer b.wg.Done()
+
+		select {
+		case <-firstPeerConnect:
```
halseth: this should have a comment.

blockmanager.go
```diff
@@ -411,7 +420,6 @@ func (b *blockManager) cfHandler() {
 	// done.
 	defer func() {
```
halseth: function closure not needed anymore

blockmanager.go
```diff
@@ -259,7 +259,16 @@ func (b *blockManager) Start() {
 	log.Trace("Starting block manager")
 	b.wg.Add(2)
 	go b.blockHandler()
-	go b.cfHandler()
+	go func() {
+		defer b.wg.Done()
+
+		select {
+		case <-firstPeerConnect:
+			b.cfHandler()
```
halseth: I would also add debug logs `"Waiting for peer connection..."` and `"Peer connected, starting cfHandler..."`

roeierez: good suggestion, added.

blockmanager.go
roeierez: added.

blockmanager.go
roeierez: I removed the outdated comment.
I thought of leaving the rest for leaving the  logging behavior as it was. Do you suggest to remove that?

neutrino.go
roeierez: fixed

blockmanager.go
```diff
@@ -259,7 +259,21 @@ func (b *blockManager) Start() {
 	log.Trace("Starting block manager")
 	b.wg.Add(2)
 	go b.blockHandler()
-	go b.cfHandler()
+	go func() {
+		defer b.wg.Done()
+
+		log.Debug("Waiting for peer connection...")
+
+		// Before starting the cfHandler we want to make sure we are connected
```
halseth: line looks long?

blockmanager.go
halseth: Oh, is just meant that this cane be `defer log.Trace("Committed filter header handler done")` now :)

blockmanager.go
roeierez: 79 characters, wrapping should start at 80 right?

blockmanager.go
roeierez: sure thing! fixed.

blockmanager.go
```diff
@@ -259,7 +259,21 @@ func (b *blockManager) Start() {
...
+		// with at least one peer.
+		select {
+		case <-firstPeerConnect:
+			log.Debug("Peer connected, starting cfHandler...")
```
halseth: nit: `log.Debug("Peer connected, starting cfHandler.")`

blockmanager.go
```diff
@@ -259,7 +259,21 @@ func (b *blockManager) Start() {
...
+			b.cfHandler()
```
halseth: style nit: this select can be written

```golang
select {
	case <-firstPeerConnect:
	case <-b.quit:
		return
	}
}

log.Debug("Peer connected, starting cfHandler...")
b.cfHandler()
```

blockmanager.go
roeierez: fixed

blockmanager.go
roeierez: fixed

PR #126, opened 2019-01-31
neutrino: pass firstPeerConnect as arg to blockManager constructor

Roasbeef:
In this commit, we perform a small refactoring to pass in the
`firstPeerConnect` channel as an argument to the constructor of the
blockManager rather than passing it via the start method.

### DIFF
--- a/blockmanager_test.go
+++ b/blockmanager_test.go
@@ -70,7 +70,7 @@ func setupBlockManager() (*blockManager, headerfs.BlockHeaderStore,
 	}
 
 	// Set up a blockManager with the chain service we defined.
-	bm, err := newBlockManager(cs)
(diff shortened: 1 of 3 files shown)


blockmanager_test.go
```diff
@@ -70,7 +70,7 @@ func setupBlockManager() (*blockManager, headerfs.BlockHeaderStore,
 	}
 
 	// Set up a blockManager with the chain service we defined.
-	bm, err := newBlockManager(cs)
+	bm, err := newBlockManager(cs, nil)
```
wpaulino: We don't yet use `blockManager.Start()` anywhere within the tests, but if we do, this will block startup. Should probably add a `nil` check there to avoid this.

PR #129, opened 2019-02-17
blockmanager: fix OP_RETURN bug

Roasbeef:
In this commit, we fix a bug in the way we attempt to verify the filters
during conflict resolution. A bug was fixed in the `btcutil` library
that would add some OP_RETURN items to the filter incorrectly. This was
remedied in the library, but this check was never updated itself. We put
all libraries in line by skipping _all_ OP_RETURN outputs when we
attempt to verify the integrity of a filter.


Roasbeef:
Travis can't find our coveralls repo for some reason, but the tests pass locally.

PR #131, opened 2019-03-02
fix travis

wakiyamap:
I think Travis will function normally in both cases, but I think that this one is easier to understand.


wakiyamap:
thanks!!

PR #133, opened 2019-03-14
Cache key fix

halseth:
Using a block hash pointer in the filter cache key would lead to the key
not being equal even when the block hash matched.

### DIFF
--- a/cache/cacheable_filter.go
+++ b/cache/cacheable_filter.go
@@ -8,7 +8,7 @@ import (
 
 // filterCacheKey represents the key used to access filters in the FilterCache.
 type FilterCacheKey struct {
-	BlockHash  *chainhash.Hash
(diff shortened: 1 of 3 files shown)


cache/cacheable_filter.go
```diff
@@ -1,6 +1,16 @@
 package cache
 
-import "github.com/btcsuite/btcutil/gcs"
+import (
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcutil/gcs"
+	"github.com/lightninglabs/neutrino/filterdb"
+)
+
+// filterCacheKey represents the key used to access filters in the FilterCache.
```
wpaulino: Nit: should be `FilterCacheKey` now.

PR #134, opened 2019-03-15
Batch filter fetching

halseth:
This PR implements batch filter fetching. This is achieved by giving the callers of `GetCFilter` the option to make a "optimistic batch query", where the filters following the requested filters are also fetched, and added to the filter cache.

I ran this on mainnet to find cache size needed to keep a whole filter batch, and a size of 30 MB could hold at minimum 1483 elements (tested on filters at height > 515110), which should give us some wiggle room.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -10,8 +10,9 @@ var (
 // Cache represents a generic cache.
 type Cache interface {
 	// Put stores the given (key,value) pair, replacing existing value if
-	// key already exists.
-	Put(key interface{}, value Value) error
+	// key already exists. The return value indicates whether items had to
+	// be evicted to make room for the new element.
+	Put(key interface{}, value Value) (bool, error)
 
 	// Get returns the value for a given key.
 	Get(key interface{}) (Value, error)
--- a/headerfs/file.go
+++ b/headerfs/file.go
@@ -3,6 +3,7 @@ package headerfs
 import (
 	"bytes"
 	"fmt"
+	"os"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
@@ -49,9 +50,10 @@ func (h *headerStore) readRaw(seekDist uint64) ([]byte, error) {
 	return rawHeader[:], nil
 }
 
-// readHeaderRange will attempt to fetch a series of headers within the target
-// height range. This method batches a set of reads into a single system call
-// thereby increasing performance when reading a set of contiguous headers.
+// readHeaderRange will attempt to fetch a series of block headers within the
+// target height range. This method batches a set of reads into a single system
+// call thereby increasing performance when reading a set of contiguous
+// headers.
 //
 // NOTE: The end height is _inclusive_ so we'll fetch all headers from the
 // startHeight up to the end height, including the final header.
@@ -59,38 +61,17 @@ func (h *blockHeaderStore) readHeaderRange(startHeight uint32,
 	endHeight uint32) ([]wire.BlockHeader, error) {
 
 	// Based on the defined header type, we'll determine the number of
-	// bytes that we need to read past the sync point.
-	var headerSize uint32
-	switch h.indexType {
-	case Block:
-		headerSize = 80
-
-	case RegularFilter:
-		headerSize = 32
-
-	default:
-		return nil, fmt.Errorf("unknown index type: %v", h.indexType)
-	}
-
-	// Each header is 80 bytes, so using this information, we'll seek a
-	// distance to cover that height based on the size of block headers.
-	seekDistance := uint64(startHeight) * uint64(headerSize)
-
-	// Based on the number of headers in the range, we'll allocate a single
-	// slice that's able to hold the entire range of headers.
-	numHeaders := endHeight - startHeight + 1
-	rawHeaderBytes := make([]byte, headerSize*numHeaders)
-
-	// Now that we have our slice allocated, we'll read out the entire
-	// range of headers with a single system call.
-	_, err := h.file.ReadAt(rawHeaderBytes, int64(seekDistance))
+	// bytes that we need to read from the file.
+	headerReader, err := readHeadersFromFile(
+		h.file, BlockHeaderSize, startHeight, endHeight,
+	)
 	if err != nil {
 		return nil, err
 	}
 
 	// We'll now incrementally parse out the set of individual headers from
 	// our set of serialized contiguous raw headers.
-	headerReader := bytes.NewReader(rawHeaderBytes)
+	numHeaders := endHeight - startHeight + 1
 	headers := make([]wire.BlockHeader, 0, numHeaders)
 	for headerReader.Len() != 0 {
 		var nextHeader wire.BlockHeader
@@ -140,4 +121,28 @@ func (f *FilterHeaderStore) readHeader(height uint32) (*chainhash.Hash, error) {
 	}
 
 	return chainhash.NewHash(rawHeader)
+}
+
+// readHeadersFromFile reads a chunk of headers, each of size headerSize, from
+// the given file, from startHeight to endHeight.
+func readHeadersFromFile(f *os.File, headerSize, startHeight,
+	endHeight uint32) (*bytes.Reader, error) {
+
+	// Each header is headerSize bytes, so using this information, we'll
+	// seek a distance to cover that height based on the size the headers.
+	seekDistance := uint64(startHeight) * uint64(headerSize)
+
+	// Based on the number of headers in the range, we'll allocate a single
+	// slice that's able to hold the entire range of headers.
+	numHeaders := endHeight - startHeight + 1
+	rawHeaderBytes := make([]byte, headerSize*numHeaders)
+
+	// Now that we have our slice allocated, we'll read out the entire
+	// range of headers with a single system call.
+	_, err := f.ReadAt(rawHeaderBytes, int64(seekDistance))
+	if err != nil {
+		return nil, err
+	}
+
+	return bytes.NewReader(rawHeaderBytes), nil
 }
--- a/headerfs/file.go
+++ b/headerfs/file.go
@@ -123,6 +123,41 @@ func (f *FilterHeaderStore) readHeader(height uint32) (*chainhash.Hash, error) {
 	return chainhash.NewHash(rawHeader)
 }
 
+// readHeaderRange will attempt to fetch a series of filter headers within the
+// target height range. This method batches a set of reads into a single system
+// call thereby increasing performance when reading a set of contiguous
+// headers.
+//
+// NOTE: The end height is _inclusive_ so we'll fetch all headers from the
+// startHeight up to the end height, including the final header.
+func (f *FilterHeaderStore) readHeaderRange(startHeight uint32,
+	endHeight uint32) ([]chainhash.Hash, error) {
+
+	// Based on the defined header type, we'll determine the number of
+	// bytes that we need to read from the file.
+	headerReader, err := readHeadersFromFile(
+		f.file, RegularFilterHeaderSize, startHeight, endHeight,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// We'll now incrementally parse out the set of individual headers from
+	// our set of serialized contiguous raw headers.
+	numHeaders := endHeight - startHeight + 1
+	headers := make([]chainhash.Hash, 0, numHeaders)
+	for headerReader.Len() != 0 {
+		var nextHeader chainhash.Hash
+		if _, err := headerReader.Read(nextHeader[:]); err != nil {
+			return nil, err
+		}
+
+		headers = append(headers, nextHeader)
+	}
+
+	return headers, nil
+}
+
 // readHeadersFromFile reads a chunk of headers, each of size headerSize, from
 // the given file, from startHeight to endHeight.
 func readHeadersFromFile(f *os.File, headerSize, startHeight,
--- a/headerfs/index.go
+++ b/headerfs/index.go
@@ -58,6 +58,15 @@ const (
 	RegularFilter
 )
 
+const (
+	// BlockHeaderSize is the size in bytes of the Block header type.
+	BlockHeaderSize = 80
+
+	// RegularFilterHeaderSize is the size in bytes of the RegularFilter
+	// header type.
+	RegularFilterHeaderSize = 32
+)
+
 // headerIndex is an index stored within the database that allows for random
 // access into the on-disk header file. This, in conjunction with a flat file
 // of headers consists of header database. The keys have been specifically
--- a/neutrino.go
+++ b/neutrino.go
@@ -70,9 +70,13 @@ var (
 	// from DNS.
 	DisableDNSSeed = false
 
-	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino will
-	// keep in memory if no size is specified in the neutrino.Config.
-	DefaultFilterCacheSize uint64 = 4096 * 1000
+	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino
+	// will keep in memory if no size is specified in the neutrino.Config.
+	// Since we utilize the cache during batch filter fetching, it is
+	// beneficial if it is able to to keep a whole batch. The current batch
+	// size is 1000, so we default to 30 MB, which can fit about 1450 to
+	// 2300 mainnet filters.
+	DefaultFilterCacheSize uint64 = 3120 * 10 * 1000
 
 	// DefaultBlockCacheSize is the size (in bytes) of blocks neutrino will
 	// keep in memory if no size is specified in the neutrino.Config.
--- a/query.go
+++ b/query.go
@@ -25,6 +25,11 @@ var (
 	// query.
 	QueryTimeout = time.Second * 3
 
+	// QueryBatchTimout is the total time we'll wait for a batch fetch
+	// query to complete.
+	// TODO(halseth): instead use timeout since last received response?
+	QueryBatchTimeout = time.Second * 30
+
 	// QueryNumRetries specifies how many times to retry sending a query to
 	// each peer before we've concluded we aren't going to get a valid
 	// response. This allows to make up for missed messages in some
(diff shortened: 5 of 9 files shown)


query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	var ok bool
+	select {
 
-			log.Tracef("Wrote filter for block %s, type %d",
-				blockHash, filterType)
+	// We'll return immediately to the caller when the filter arrives.
+	case filter, ok = <-filterChan:
+		if !ok {
+			// TODO: return error?
+			return nil, nil
```
wpaulino: We can use an error chan to send any of the fatal errors back and use that here.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
 		}
+
+		return filter, nil
+
+	case <-s.quit:
+		// TODO: return error?
+		return nil, nil
```
wpaulino: Yeah can return something like `ErrChainServiceExiting`.

headerfs/file.go
```diff
@@ -49,50 +51,31 @@ func (h *headerStore) readRaw(seekDist uint64) ([]byte, error) {
...
 		return nil, err
 	}
 
 	// We'll now incrementally parse out the set of individual headers from
 	// our set of serialized contiguous raw headers.
-	headerReader := bytes.NewReader(rawHeaderBytes)
+	numHeaders := endHeight - startHeight + 1
 	headers := make([]wire.BlockHeader, 0, numHeaders)
-	for headerReader.Len() != 0 {
+	for i := uint32(0); i < numHeaders; i++ {
```
Roasbeef: Why this over the prior logic? Assuming all contents are written to disk in 80 byte chunks the prior works as well assuming the method still returns a ` bytes.Reader`.

headerfs/file.go
```diff
@@ -49,50 +51,31 @@ func (h *headerStore) readRaw(seekDist uint64) ([]byte, error) {
...
-	numHeaders := endHeight - startHeight + 1
-	rawHeaderBytes := make([]byte, headerSize*numHeaders)
-
-	// Now that we have our slice allocated, we'll read out the entire
-	// range of headers with a single system call.
-	_, err := h.file.ReadAt(rawHeaderBytes, int64(seekDistance))
+	// bytes that we need to read from the file.
+	var headerSize uint32 = BlockHeaderSize
+	headerReader, err := readHeadersFromFile(
+		h.file, headerSize, startHeight, endHeight,
```
Roasbeef: This ignores the filter header case.

cache/cache.go
```diff
@@ -10,8 +10,9 @@ var (
 // Cache represents a generic cache.
 type Cache interface {
 	// Put stores the given (key,value) pair, replacing existing value if
-	// key already exists.
-	Put(key interface{}, value Value) error
+	// key already exists. The return value indicates whether items had to
```
Roasbeef: Commit body could use some details w.r.t the rationale for this change.

neutrino.go
```diff
@@ -70,9 +70,13 @@ var (
...
 	DisableDNSSeed = false
 
-	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino will
-	// keep in memory if no size is specified in the neutrino.Config.
-	DefaultFilterCacheSize uint64 = 4096 * 1000
+	// DefaultFilterCacheSize is the size (in bytes) of filters neutrino
+	// will keep in memory if no size is specified in the neutrino.Config.
+	// Since we utilize the cache during batch filter fetching, it is
+	// beneficial if it is able to to keep a whole batch. The current batch
+	// size is 1000, so we default to 30 MB, which can room about 1450 to
```
Roasbeef: can room -> can fit

Roasbeef: or "has room for"

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+
+		s.queryPeers(
+			// Send a wire.MsgGetCFilters
+			wire.NewMsgGetCFilters(filterType, startHeight, stopHash),
+
+			// Check responses and if we get one that matches, end
+			// the query early.
+			func(sp *ServerPeer, resp wire.Message, quit chan<- struct{}) {
+				switch response := resp.(type) {
+				// We're only interested in "cfilter" messages.
```
Roasbeef: This should definitely be extracted into a distinct function.

query.go
```diff
@@ -749,14 +749,8 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 		return nil, fmt.Errorf("unknown filter type: %v", filterType)
 	}
 
-	// Only get one CFilter at a time to avoid redundancy from mutliple
```
Roasbeef: Without this mutex, we'll have multiple concurrent requests for the same set of filters.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	if len(filterHeaders) != int(numFilters)+1 {
+		return nil, fmt.Errorf("expected %d filter headers, got %d",
+			numFilters+1, len(filterHeaders))
+	}
+
+	// We create a height index such that we can easily index into our
+	// header slices for a given block hash in the received response,
+	// without consulting the database. This also keeps track of for which
+	// blocks we are still awaiting a response. We start at index=1, as 0
+	// is for the block startHeight-1, only needed for validation.
```
Roasbeef: > which is only needed for validation

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	stopHash, err := s.GetBlockHash(int64(stopHeight))
 	if err != nil {
-		return nil, fmt.Errorf("Couldn't get cfheader for block %s "+
-			"from database", blockHash)
+		return nil, fmt.Errorf("unable to get hash for "+
+			"stopHeight=%d: %v", stopHeight, err)
 	}
-	prevHeader, err := getHeader(&block.PrevBlock)
+
+	// In order to verify the authenticity of the received filters, we'll
```
Roasbeef: I'd say this logic along with the height index itself could be extracted to a method that returns the set of materials needed to generate the queries as well as validate the received filter headers, returning a struct of all the necessary items.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	}
+
+	// We'll immediately respond to the caller with the requested filter
+	// when it is received, so we make a channel to be notified when it's
+	// ready.
+	filterChan := make(chan *gcs.Filter, 1)
+
+	// We acquire the mutex ensuring we don't have several redundant
+	// CFilter queries running in parallel.
+	s.mtxCFilter.Lock()
```
Roasbeef: So if we have multiple goroutines trying to obtain the same set of filters, then at this point they've done a lot of unnecessary work w.r.t reading all the ancestors from disk, and obtaining the filter validation context.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
-				// found a filter, or we risk closing an
-				// already closed channel.
-				if filter != nil {
-					return
-				}
+	go func() {
+		defer s.mtxCFilter.Unlock()
+		defer close(filterChan)
+
+		s.queryPeers(
```
Roasbeef: Given that we may now be requesting a series of headers from our peers, we may want to consider using `queryBatch` here, as it'll spread out the load amongst all of our peers allowing the fastest ones to consume most of the queries.

headerfs/index.go
```diff
@@ -58,6 +58,15 @@ const (
 	RegularFilter
 )
 
+const (
+	// BlockHeaderSize is the size in bytes of the Block header type.
+	BlockHeaderSize = 80
```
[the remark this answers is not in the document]
halseth: Hm, I think it is just github showing them in a weird order. Should be fixed now :)

headerfs/file.go
halseth: This method is only used for blockheaders

headerfs/file.go
```diff
@@ -141,3 +124,25 @@ func (f *FilterHeaderStore) readHeader(height uint32) (*chainhash.Hash, error) {
 
 	return chainhash.NewHash(rawHeader)
 }
+
+func readHeadersFromFile(f *os.File, headerSize, startHeight,
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/134/commits/3773d1f67b01982e546cb5f0b88ffdb07ef738a5

headerfs/file.go
halseth: Reverted to returning `bytes.Reader`: https://github.com/lightninglabs/neutrino/pull/134/commits/324253acdbd2305da9bbda5b5de133bc8e480cf5

cache/cache.go
halseth: Done.

query.go
```diff
@@ -25,6 +25,11 @@ var (
 	// query.
 	QueryTimeout = time.Second * 3
 
+	//QueryBatchTimout is the total time we'll wait for a batch fetch query
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/134/commits/d2a3721b9e365f77b5ab692083651f0a925e7bdb

query.go
halseth: Done.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	// the hash of the previous block is needed for validation.
+	numFilters := stopHeight - startHeight + 1
+	blockHeaders, _, err := s.BlockHeaders.FetchHeaderAncestors(
+		numFilters, stopHash,
+	)
 	if err != nil {
-		return nil, fmt.Errorf("Couldn't get cfheader for block %s "+
-			"from database", blockHash)
+		return nil, fmt.Errorf("unable to get %d blockheader "+
+			"ancestors for stopHash=%d: %v", numFilters,
```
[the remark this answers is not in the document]
halseth: Done.

query.go
halseth: It is actually just a single query (GetCFilters), so not sure if `queryBatch` applies?

query.go
halseth: Just returning nil for now, to not change the behavior of `GetCFilter` from the perspective of the caller.

query.go
halseth: Same.

query.go
```diff
@@ -778,120 +772,272 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+		return nil, fmt.Errorf("expected %d filter headers, got %d",
+			numFilters+1, len(filterHeaders))
+	}
+
+	// We create a height index such that we can easily index into our
+	// header slices for a given block hash in the received response,
+	// without consulting the database. This also keeps track of for which
+	// blocks we are still awaiting a response. We start at index=1, as 0
+	// is for the block startHeight-1, only needed for validation.
+	heightIndex := make(map[chainhash.Hash]int)
```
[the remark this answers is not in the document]
halseth: Renamed to `headerIndex`.

neutrino.go
halseth: Done.

query.go
halseth: Done.

query.go
halseth: Done.

query.go
halseth: Fixed.

halseth:
Comments addressed, PTAL @Roasbeef @wpaulino

query.go
```diff
@@ -763,6 +763,118 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
 	return s.FilterCache.Put(cacheKey, &cache.CacheableFilter{Filter: filter})
 }
 
+// cfiltersQuery is a strcut that holds all the information necessary to
+// perform batch GetCFilters request, and handle the responses.
+type cfiltersQuery struct {
```
wpaulino: :+1:

query.go
```diff
@@ -785,43 +790,127 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
+		stopHeight = startHeight + wire.MaxGetCFiltersReqRange - 1
+
+		// We need a longer timeout, since we are going to receive more
+		// than a single response.
+		options = append(options, Timeout(QueryBatchTimeout))
+
+	// Reverse batch, fetch as many of the preceding filters as possible.
+	case reverseBatch:
+		stopHeight = int64(height)
+		startHeight = stopHeight - wire.MaxGetCFiltersReqRange + 1
```
wpaulino: Hmm I'm a bit worried about this. I guess the assumption is that by the time we get to do a reverse batch filter fetch, they've most likely already been fetched from a previous forward batch? Otherwise, let's say the reverse batch happens before anything else. On mobile, we expect users to be opening and closing the app multiple times a day, so this would be a pretty big bandwidth hit if we end up fetching 1000 filters unnecessarily every time. Not sure what the best solution is other than lowering the max range for a reverse batch.

query.go
```diff
@@ -785,43 +790,127 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
+	// [startHeight-1, stopHeight]. We go one below our startHeight since
+	// the hash of the previous block is needed for validation.
+	numFilters := uint32(stopHeight - startHeight + 1)
+	blockHeaders, _, err := s.BlockHeaders.FetchHeaderAncestors(
+		numFilters, stopHash,
+	)
 	if err != nil {
-		return nil, fmt.Errorf("Couldn't get cfheader for block %s "+
-			"from database", blockHash)
+		return nil, fmt.Errorf("unable to get %d blockheader "+
```
wpaulino: Nit: s/blockheader/block header/

query.go
```diff
@@ -785,43 +790,127 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
+	}
+
+	if len(filterHeaders) != int(numFilters)+1 {
+		return nil, fmt.Errorf("expected %d filter headers, got %d",
+			numFilters+1, len(filterHeaders))
 	}
 
+	// We create a header index such that we can easily index into our
+	// header slices for a given block hash in the received response,
+	// without consulting the database. This also keeps track of for which
```
wpaulino: Nit: s/for//

query.go
```diff
@@ -785,43 +790,127 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
+	// blocks we are still awaiting a response. We start at index=1, as 0
+	// is for the block startHeight-1, which is only needed for validation.
+	headerIndex := make(map[chainhash.Hash]int)
+	for i := 1; i < len(blockHeaders); i++ {
+		block := blockHeaders[i]
+		headerIndex[block.BlockHash()] = i
+	}
+
+	// We'll immediately respond to the caller with the requested filter
+	// when it is received, so we make a to notify on when it's ready.
```
wpaulino: Nit: make a channel*

query.go
```diff
@@ -785,43 +790,127 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
+		return nil, fmt.Errorf("expected %d filter headers, got %d",
+			numFilters+1, len(filterHeaders))
 	}
 
+	// We create a header index such that we can easily index into our
+	// header slices for a given block hash in the received response,
+	// without consulting the database. This also keeps track of for which
+	// blocks we are still awaiting a response. We start at index=1, as 0
+	// is for the block startHeight-1, which is only needed for validation.
+	headerIndex := make(map[chainhash.Hash]int)
```
wpaulino: Nit: provide length.

query.go
halseth: That's a good point, can totally happen that in most cases we will need only a few of the filters at tip.

I think however that it is a bit premature to be special casing this before we got some real data to back it up. Actually, maybe we should even leave reverse fetching out for now, until we know the impact of using it?

headerfs/file.go
```diff
@@ -49,48 +50,29 @@ func (h *headerStore) readRaw(seekDist uint64) ([]byte, error) {
...
-	// Based on the number of headers in the range, we'll allocate a single
-	// slice that's able to hold the entire range of headers.
-	numHeaders := endHeight - startHeight + 1
-	rawHeaderBytes := make([]byte, headerSize*numHeaders)
-
-	// Now that we have our slice allocated, we'll read out the entire
-	// range of headers with a single system call.
-	_, err := h.file.ReadAt(rawHeaderBytes, int64(seekDistance))
+	// bytes that we need to read from the file.
+	var headerSize uint32 = BlockHeaderSize
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/134/commits/89d3e621a3d68e464a8c268a735c83d72ec69936

headerfs/file.go
```diff
@@ -124,6 +124,42 @@ func (f *FilterHeaderStore) readHeader(height uint32) (*chainhash.Hash, error) {
...
+// headers.
+//
+// NOTE: The end height is _inclusive_ so we'll fetch all headers from the
+// startHeight up to the end height, including the final header.
+func (f *FilterHeaderStore) readHeaderRange(startHeight uint32,
+	endHeight uint32) ([]chainhash.Hash, error) {
+
+	// Based on the defined header type, we'll determine the number of
+	// bytes that we need to read from the file.
+	var headerSize uint32 = RegularFilterHeaderSize
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/134/commits/c368fdc8d055e38857b4e0db64807af7b6f6ec6e

query.go
```diff
@@ -763,6 +763,118 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
 	return s.FilterCache.Put(cacheKey, &cache.CacheableFilter{Filter: filter})
 }
 
+// cfiltersQuery is a strcut that holds all the information necessary to
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/134/commits/20e934b84f5e57b181d58497655e376d3aafbb46

query.go
halseth: Done: https://github.com/lightninglabs/neutrino/pull/134/commits/baf5ed86aa5e6b00f57c58b12132d94a5128e6e9

query.go
halseth: https://github.com/lightninglabs/neutrino/pull/134/commits/36d528abfa9d4c8ee4fa88f56340e77ba55bc1e0

query.go
halseth: Done: https://github.com/lightninglabs/neutrino/pull/134/commits/f275dd44a1f0ae9b131b32a1a8f6cf490afb03b6

query.go
halseth: 👍  

Done: https://github.com/lightninglabs/neutrino/pull/134/commits/2be9b2db66b0f6c39012b579b4d620d763ef4dac

query.go
wpaulino: I don't think we want to remove it. For Neutrino, we scan backwards when scanning for a historical confirmation. If we don't address the reverse batch range, then we basically depend on the order of `lnd`'s subsystems to ensure a spend registration happens before a confirmation registration, which doesn't seem like a reasonable option to me.

FWIW, currently it seems like spend registrations usually happen before confirmations due to starting the `ChannelArbitrator`s, but this might not always be the case in the future.

halseth: Solution for now seems to be for the caller to not request an optimistic batch if the range of filters to scan is known to be small?

PR #139, opened 2019-04-08
blockmanager: remove panic on failed WriteHeaders

halseth:
If headers were written while lnd was in the process of shutting down it
would panic with a "Database not open" error.


halseth:
@cfromknecht Yeah, you're right. That is indeed a better fix.

cfromknecht:
Fwiw I’m not against removing the panic as well :)

halseth:
Fixed by making sure `Neutrino` is stopped before closing the DB: 

'btcwallet': https://github.com/btcsuite/btcwallet/pull/610
'lnd': https://github.com/lightningnetwork/lnd/pull/2910

Roasbeef:
This stops the panic in the short term, will also take a look at the two related PRs to land those as well.

PR #140, opened 2019-04-10
Detect peers serving invalid filters

halseth:
This PR adds a set of extra consistency checks to downloaded filters, in an attempt to ban peers that are serving either old filter types, or filters that are inconsistent with the filter headers.

If we cannot determine directly from the filter headers that a peer is serving invalid filters we employ the following strategies in attempt to ban the smallest set of peers needed for the sync not to stall:

1. If a peers' filter doesn't match on a script that must match, we know the filter is invalid.
2. If a peers' filter matches on a script that _should not_ match, it is potentially invalid. In this case we ban peers that matches more such scripts than other peers.
3. If we cannot detect which filters are invalid from the block contents, we ban peers serving filters different from the majority of peers.

For the majority rule to kick in, we currently require the majority of peers serving the same filter to be **at least 6**. If this requirement is not met, we will need more peers for the sync to continue.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -756,33 +756,8 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 		if checkForCFHeaderMismatch(headers, i) {
 			targetHeight := startHeight + uint32(i)
 
-			log.Warnf("Detected cfheader mismatch at "+
-				"height=%v!!!", targetHeight)
-
-			// Get the block header for this height, along with the
-			// block as well.
-			header, err := b.server.BlockHeaders.FetchHeaderByHeight(
-				targetHeight,
-			)
-			if err != nil {
-				return err
-			}
-			block, err := b.server.GetBlock(header.BlockHash())
-			if err != nil {
-				return err
-			}
-
-			log.Warnf("Attempting to reconcile cfheader mismatch "+
-				"amongst %v peers", len(headers))
-
-			// We'll also fetch each of the filters from the peers
-			// that reported check points, as we may need this in
-			// order to determine which peers are faulty.
-			filtersFromPeers := b.fetchFilterFromAllPeers(
-				targetHeight, header.BlockHash(), fType,
-			)
-			badPeers, err := resolveCFHeaderMismatch(
-				block.MsgBlock(), fType, filtersFromPeers,
+			badPeers, err := b.detectBadPeers(
+				headers, targetHeight, uint32(i), fType,
 			)
 			if err != nil {
 				return err
@@ -1289,31 +1264,8 @@ func (b *blockManager) resolveConflict(
 			// block as well.
 			targetHeight := startHeight + uint32(i)
 
-			log.Warnf("Detected cfheader mismatch at "+
-				"height=%v!!!", targetHeight)
-
-			header, err := b.server.BlockHeaders.FetchHeaderByHeight(
-				targetHeight,
-			)
-			if err != nil {
-				return nil, err
-			}
-			block, err := b.server.GetBlock(header.BlockHash())
-			if err != nil {
-				return nil, err
-			}
-
-			log.Infof("Attempting to reconcile cfheader mismatch "+
-				"amongst %v peers", len(headers))
-
-			// We'll also fetch each of the filters from the peers
-			// that reported check points, as we may need this in
-			// order to determine which peers are faulty.
-			filtersFromPeers := b.fetchFilterFromAllPeers(
-				targetHeight, header.BlockHash(), fType,
-			)
-			badPeers, err := resolveCFHeaderMismatch(
-				block.MsgBlock(), fType, filtersFromPeers,
+			badPeers, err := b.detectBadPeers(
+				headers, targetHeight, uint32(i), fType,
 			)
 			if err != nil {
 				return nil, err
@@ -1396,6 +1348,38 @@ func checkForCFHeaderMismatch(headers map[string]*wire.MsgCFHeaders,
 	}
 
 	return false
+}
+
+// detectBadPeers fetches filters and the block at the given height to attempt
+// to detect which peers are serving bad filters.
+func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
+	targetHeight, filterIndex uint32,
+	fType wire.FilterType) ([]string, error) {
+
+	log.Warnf("Detected cfheader mismatch at height=%v!!!", targetHeight)
+
+	// Get the block header for this height, along with the block as well.
+	header, err := b.server.BlockHeaders.FetchHeaderByHeight(targetHeight)
+	if err != nil {
+		return nil, err
+	}
+	block, err := b.server.GetBlock(header.BlockHash())
+	if err != nil {
+		return nil, err
+	}
+
+	log.Warnf("Attempting to reconcile cfheader mismatch amongst %v peers",
+		len(headers))
+
+	// We'll also fetch each of the filters from the peers that reported
+	// check points, as we may need this in order to determine which peers
+	// are faulty.
+	filtersFromPeers := b.fetchFilterFromAllPeers(
+		targetHeight, header.BlockHash(), fType,
+	)
+	return resolveCFHeaderMismatch(
+		block.MsgBlock(), fType, filtersFromPeers,
+	)
 }
 
 // resolveCFHeaderMismatch will attempt to cross-reference each filter received
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -1414,68 +1414,64 @@ func resolveCFHeaderMismatch(block *wire.MsgBlock, fType wire.FilterType,
 		block.Header.BlockHash())
 
 	// Based on the type of filter, our verification algorithm will differ.
-	switch fType {
+	// Only regular filters are currently defined.
+	if fType != wire.GCSFilterRegular {
+		return nil, fmt.Errorf("unknown filter: %v", fType)
+	}
 
 	// With the current set of items that we can fetch from the p2p
 	// network, we're forced to only verify what we can at this point. So
 	// we'll just ensure that each of the filters returned
 	//
 	// TODO(roasbeef): update after BLOCK_WITH_PREV_OUTS is a thing
-	case wire.GCSFilterRegular:
 
-		// We'll now run through each peer and ensure that each output
-		// script is included in the filter that they responded with to
-		// our query.
-		for peerAddr, filter := range filtersFromPeers {
-		peerVerification:
+	// We'll now run through each peer and ensure that each output
+	// script is included in the filter that they responded with to
+	// our query.
+	for peerAddr, filter := range filtersFromPeers {
+		// We'll ensure that all the filters include every output
+		// script within the block.
+		//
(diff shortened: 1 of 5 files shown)


blockmanager_test.go
```diff
@@ -624,3 +626,46 @@ func TestBlockManagerInvalidInterval(t *testing.T) {
...
+	}
+
+	return b.Build()
+}
+
+// buildOldOldFilter creates a CFilter with all output scripts, including
+// OP_RETURNS.
+//
+// NOTE: this is not a valid filter, only for tests.
+func buildOldOldFilter(block *wire.MsgBlock) (*gcs.Filter, error) {
```
wpaulino: Lol could rename these to what the filter includes. So this'd be something like `buildAllPkScriptsFilter`.

blockmanager.go
```diff
@@ -1334,11 +1334,37 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
...
+	)
+
+	var badPeers []string
+	for peer, _ := range headers {
+		_, ok := filtersFromPeers[peer]
+
+		// If a peer did not respond, ban it immediately.
+		if !ok {
+			log.Warnf("Peer %v did not respond to filter "+
+				"request, considering bad.", peer)
```
wpaulino: Nit: we usually don't add periods to the end of log messages.

blockmanager.go
```diff
@@ -1334,11 +1334,37 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
...
+			badPeers = append(badPeers, peer)
+			continue
+		}
+	}
+
+	if len(badPeers) != 0 {
+		return badPeers, nil
```
wpaulino: Any reason we want to return here and not continue to determine if we have any other bad peers?

blockmanager.go
```diff
@@ -1357,14 +1357,26 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
...
 			continue
 		}
+
+		// If the peer is serving filters that isn't consistent with
+		// its filter hashes, ban it.
+		hash, err := builder.GetFilterHash(filter)
+		if err != nil {
+			return nil, err
+		}
+		if hash != *msg.FilterHashes[filterIndex] {
```
wpaulino: :+1:

blockmanager.go
```diff
@@ -1567,6 +1570,47 @@ func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
 		return potentialBans, nil
 	}
 
+	// If all peers where serving filters consistent with the block, we
+	// cannot know for sure which one is dishonest (since we don't have the
+	// prevouts to deterministically reconstruct the filter). In this
```
wpaulino: Not sure if I'm missing something, but since we have the full block, can't we just reconstruct the filter ourselves and check its hash? We can use ComputePkScript on each transaction input's sigScript/witness in order to retrieve the other output scripts needed. Might not always work if there's some type of non-standard script in the block though.

blockmanager.go
```diff
@@ -1567,6 +1570,47 @@ func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
...
+	best := 0
+	for _, filter := range filtersFromPeers {
+		hash, err := builder.GetFilterHash(filter)
+		if err != nil {
+			return nil, err
+		}
+
+		cnt := count[hash] + 1
+		count[hash] = cnt
+		if cnt > best {
```
wpaulino: Nit: can do 
```
count[hash]++
if count[hash] > best
```

blockmanager.go
```diff
@@ -1440,8 +1491,44 @@ func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
 	// check whether the store or the checkpoints we got from the network
 	// are correct.
 
-	// With the set of bad peers known, we'll collect a slice of all the
-	// faulty peers.
+	// Return the bad peers if we have already found some.
+	if len(badPeers) > 0 {
```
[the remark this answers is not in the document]
halseth: They will be caught in the next round of checks.

blockmanager.go
```diff
@@ -1353,8 +1358,8 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
 // and verify from the filter in question. We'll return all the peers that
 // returned what we believe to be an invalid filter.
 func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
-	fType wire.FilterType, filtersFromPeers map[string]*gcs.Filter) (
-	[]string, error) {
+	fType wire.FilterType, filtersFromPeers map[string]*gcs.Filter,
+	threshold int) ([]string, error) {
```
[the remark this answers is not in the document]
halseth: Done.

blockmanager.go
halseth: Idea is to return as early as possible, before starting expensive operations such as getting blocks.

blockmanager.go
```diff
@@ -1440,8 +1491,44 @@ func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
...
+
+	// Gather up the peers with the most OP_RETURN matches.
+	var potentialBans []string
+	for peer, cnt := range opReturnMatches {
+		if cnt == mostMatches {
+			potentialBans = append(potentialBans, peer)
+		}
+	}
+
+	// If we only a few peers had matching OP_RETURNS, we assume they are
```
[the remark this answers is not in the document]
halseth: Fixed

blockmanager_test.go
halseth: lol fixed

blockmanager.go
halseth: Fixed.

blockmanager.go
halseth: 👍

blockmanager.go
halseth: Posting my answer that I think got lost in chat logs:

> we can compute the pkscript for supported types, and make sure at least they match in the filter

I tried this on recent blocks on mainnet, and we fail to interpret signature scripts containing uncompressed pubkeys as P2PKH
instead we think they are P2SH, and fail to match them with the filter
@wpaulino any reason for the ComputePKScript to assume only compressed pubkeys?
I also see that we are assuming signatures to be 71 or 72 bytes, while they can be other lengths, again causing us to assume p2sh

wpaulino: No reason for it really. Would it help if those were to be checked for as well, or would we run into other unknown/unsupported scripts?

halseth: I would think we never would be able to reconstruct all scripts, as someone could always make a non-standard one.

wpaulino: Sure, I was more just wondering if it could be done for the majority of blocks if we address the `P2PKH` gaps.

halseth:
PTAL @wpaulino  @Roasbeef

blockmanager.go
```diff
@@ -1393,68 +1393,64 @@ func resolveCFHeaderMismatch(block *wire.MsgBlock, fType wire.FilterType,
...
+
+				if match {
+					continue
+				}
+
+				// If this filter doesn't match, then we'll
+				// mark this peer as bad and move on to the
+				// next peer.
+				badPeers[peerAddr] = struct{}{}
+				continue peerVerification
```
wpaulino: Seems like this should be `break` instead, otherwise we'll continue to check the block again?

blockmanager.go
```diff
@@ -1358,16 +1358,21 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
 	)
 	return resolveFilterMismatchFromBlock(
 		block.MsgBlock(), fType, filtersFromPeers,
+
+		// We'll need half of our peers to agree on filters.
+		(len(filtersFromPeers)+1)/2,
```
wpaulino: We'd want at least a minimum of 3 peers in order to detect a majority, so I think the minimum threshold should be 2.

blockmanager.go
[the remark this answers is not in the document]
wpaulino: Which round of checks?

blockmanager.go
```diff
@@ -1453,8 +1504,43 @@ func resolveFilterMismatchFromBlock(block *wire.MsgBlock,
...
+		if cnt == mostMatches {
+			potentialBans = append(potentialBans, peer)
+		}
+	}
+
+	// If only a few peers had matching OP_RETURNS, we assume they are bad.
+	numRemaining := len(filtersFromPeers) - len(potentialBans)
+	if len(potentialBans) > 0 && numRemaining >= threshold {
+		log.Warnf("Found %d peers serving filters with unexpected "+
+			"OP_RETURNS. %d peers remaining.", len(potentialBans),
```
wpaulino: Nit: logs usually don't include periods.

mock_store.go
```diff
@@ -39,11 +41,17 @@ func (m *mockBlockHeaderStore) LatestBlockLocator() (
...
+
 func (m *mockBlockHeaderStore) FetchHeaderByHeight(height uint32) (
 	*wire.BlockHeader, error) {
 
-	return nil, nil
+	if header, ok := m.heights[height]; ok {
+		return &header, nil
+	}
+
+	return nil, fmt.Errorf("not found")
```
wpaulino: Nit: return `ErrHeightNotFound`?

blockmanager.go
```diff
@@ -1343,11 +1343,37 @@ func (b *blockManager) detectBadPeers(headers map[string]*wire.MsgCFHeaders,
...
+	filtersFromPeers := b.fetchFilterFromAllPeers(
+		targetHeight, header.BlockHash(), fType,
+	)
+
+	var badPeers []string
+	for peer, _ := range headers {
+		_, ok := filtersFromPeers[peer]
+
+		// If a peer did not respond, ban it immediately.
+		if !ok {
```
wpaulino: This might be an issue for nodes with higher latency which would end up running into timeouts when fetching the filter.

blockmanager.go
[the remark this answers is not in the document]
halseth: We will re-attempt the sync after having kicked out at least one peer, and then we'll find the next bad one.

blockmanager.go
halseth: This is to ensure peers cannot avoid the filter check by just not responding to our query.

blockmanager.go
halseth: Good catch! I agree, fixed by moving label to outer loop: https://github.com/lightninglabs/neutrino/pull/140/commits/f5b07e94e6d7be48cd1c747fca1fae1e398ab9cf

blockmanager.go
halseth: Done

blockmanager.go
halseth: Fixed.

mock_store.go
halseth: Done.

blockmanager.go
wpaulino: Shouldn't be too big of an issue with the increased timeout introduced in #158.

Roasbeef: We'll definitely want to make this timeout dynamic in the future. We can also use traffic shaping utilities locally to emulate high latency links as well.

PR #141, opened 2019-04-12
query: query remaining peers when NumRetries is 0

wpaulino:
NumRetries signals that we should retry all of our peers a certain amount of times. Before this commit, if set to 0, we'd skip over all of our remaining peers because peerTries would return a 0 value if the peer had yet to be tried. We fix this by ensuring that we try all peers _once_ regardless of what NumRetries is set to.

### DIFF
--- a/query.go
+++ b/query.go
@@ -662,21 +662,31 @@ checkResponses:
 			}
 
 			queryPeer = nil
-			for _, curPeer := range s.Peers() {
-				if curPeer != nil && curPeer.Connected() &&
-					peerTries[curPeer.Addr()] < qo.numRetries {
-
-					curPeer := curPeer
-					queryPeer = curPeer
-
-					// Found a peer we can query.
-					peerTries[queryPeer.Addr()]++
-					queryPeer.subscribeRecvMsg(subscription)
-					queryPeer.QueueMessageWithEncoding(
-						queryMsg, nil, qo.encoding,
-					)
-					break
+			for _, peer := range s.Peers() {
+				// If the peer is no longer connected, we'll
+				// skip them.
+				if !peer.Connected() {
+					continue
 				}
+
+				// If we've yet to try this peer, we'll make
+				// sure to do so. If we've exceeded the number
+				// of tries we should retry this peer, then
+				// we'll skip them.
+				numTries, ok := peerTries[peer.Addr()]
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -662,21 +662,31 @@ checkResponses:
...
+				// If we've yet to try this peer, we'll make
+				// sure to do so. If we've exceeded the number
+				// of tries we should retry this peer, then
+				// we'll skip them.
+				numTries, ok := peerTries[peer.Addr()]
+				if ok && numTries >= qo.numRetries {
+					continue
 				}
+
+				queryPeer = peer
```
halseth: Is it safe to drop the `curPeer := curPeer`?

wpaulino: Should be, there are no goroutines within the loop.

query.go
```diff
@@ -662,21 +662,31 @@ checkResponses:
...
-					peerTries[queryPeer.Addr()]++
-					queryPeer.subscribeRecvMsg(subscription)
-					queryPeer.QueueMessageWithEncoding(
-						queryMsg, nil, qo.encoding,
-					)
-					break
+			for _, peer := range s.Peers() {
+				// If the peer is no longer connected, we'll
+				// skip them.
+				if !peer.Connected() {
```
cfromknecht: Missing `peer != nil` check

wpaulino: Why is the check needed?

cfromknecht: To prevent panics, check the history

wpaulino: Are you referring to https://github.com/lightninglabs/neutrino/commit/9a0709b914d3ea43113806d8b8aaf6f87e06bf44? That seems different -- we query for a peer given their address. The `nil` check makes sense there to me as the peer can no longer exist, but here `Peers` is just returning the currently active peers.

cfromknecht: Hmm, actually i think I’m recalling a different case where checking nilness is necessary. Why is used atm?

wpaulino: I think it was left over from some shadowing changes, see https://github.com/lightninglabs/neutrino/commit/fd24569c15b5960f51dc3f336954b8406a2beac1.

cfromknecht: Ahh okay, thanks for clarifying. Clear to me now that it is not needed

PR #143, opened 2019-04-15
neutrino: reject address suggestions for already connected peers

wpaulino:
We do this to ensure we don't have more than once active connection per peer. This is needed since the addrManager has no idea about the client's currently connected peers.

We also ensure we don't reconnect to banned peers -- there's no point in attempting to do so if we'll disconnect them anyway.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -1221,7 +1221,14 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
 	}
 
 	if sp.connReq != nil {
-		s.connManager.Disconnect(sp.connReq.ID())
+		// If the peer has been banned, we'll remove the connection
+		// request from the manager to ensure we don't reconnect again.
+		// Otherwise, we'll just simply disconnect.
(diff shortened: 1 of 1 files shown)


neutrino.go
```diff
@@ -697,12 +697,35 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	var newAddressFunc func() (net.Addr, error)
 	if s.chainParams.Net != chaincfg.SimNetParams.Net {
 		newAddressFunc = func() (net.Addr, error) {
+
+			// Gather our set of currently connected peers to avoid
+			// connecting to them again.
+			connectedPeers := make(map[string]struct{})
+			for _, peer := range s.Peers() {
+				peerAddr := addrmgr.NetAddressKey(peer.NA())
+				connectedPeers[peerAddr] = struct{}{}
```
Roasbeef: This isn't enough. We also need to check in the server once peers are actually connected.

wpaulino: Fixed.

PR #145, opened 2019-04-18
query: ensure we don't re-assign previously failed queries to peers

wpaulino:
This prevents us from getting stalled by ensuring that a failed query is re-assigned to a different peer. Failed queries can happen due to timeouts as well, but usually it's because the peer cannot satisfy our query, so we should try a different one.

### DIFF
--- a/query.go
+++ b/query.go
@@ -296,6 +296,11 @@ func queryChainServiceBatch(
 			mtxPeerStates.Unlock()
 		}()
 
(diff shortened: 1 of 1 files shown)


halseth:
Hm, thinking more about it. Could it be a problem that if a query times out, it will in some cases never be retried (in case we have only 1 or very few peers)?

Something like #161 a safer solution?

wpaulino:
Closing this to proceed with #161 instead.

PR #147, opened 2019-04-19
 banman: add BanStore interface with implementation

wpaulino:
In this PR, we introduce a new BanStore interface, along with a concrete implementation that will be backed by a boltdb instance, that will serve as a persistent ban store for peers. This prevents us from reconnecting to known bad peers until their respective ban expires.

### DIFF
--- a/banman/reason.go
+++ b/banman/reason.go
@@ -1,0 +1,43 @@
+package banman
+
+// Reason includes the different possible reasons which caused us to ban a peer.
+type Reason uint8
+
+// We prevent using `iota` to ensure the order does not have the value since
+// these are serialized within the database.
+const (
+	// ExcedeedBanThreshold signals that a peer exceeded its ban threshold.
+	ExceededBanThreshold Reason = 1
+
+	// NoCompactFilters signals that a peer was unable to serve us compact
+	// filters.
+	NoCompactFilters Reason = 2
+
+	// InvalidFilterHeader signals that a peer served us an invalid filter
+	// header.
+	InvalidFilterHeader Reason = 3
+
+	// InvalidFilterHeaderCheckpoint signals that a peer served us an
+	// invalid filter header checkpoint.
+	InvalidFilterHeaderCheckpoint Reason = 4
+)
+
+// String returns a human-readable description for the reason a peer was banned.
+func (r Reason) String() string {
+	switch r {
+	case ExceededBanThreshold:
+		return "peer exceeded ban threshold"
+
+	case NoCompactFilters:
+		return "peer was unable to serve compact filters"
+
+	case InvalidFilterHeader:
+		return "peer served invalid filter header"
+
+	case InvalidFilterHeaderCheckpoint:
+		return "peer served invalid filter header checkpoint"
+
+	default:
+		return "unknown reason"
+	}
+}
--- a/banman/store.go
+++ b/banman/store.go
@@ -1,0 +1,221 @@
+package banman
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/btcsuite/btcwallet/walletdb"
+)
+
+var (
+	// byteOrder is the preferred byte order in which we should write things
+	// to disk.
+	byteOrder = binary.BigEndian
+
+	// banStoreBucket is the top level bucket of the Store that will contain
+	// all relevant sub-buckets.
+	banStoreBucket = []byte("ban-store")
+
+	// banBucket is the main index in which we keep track of IP networks and
+	// their absolute expiration time.
+	//
+	// The key is the IP network host and the value is the absolute
+	// expiration time.
+	banBucket = []byte("ban-index")
+
+	// reasonBucket is an index in which we keep track of why an IP network
+	// was banned.
+	//
+	// The key is the IP network and the value is the Reason.
+	reasonBucket = []byte("reason-index")
+
+	// ErrCorruptedStore is an error returned when we attempt to locate any
+	// of the ban-related buckets in the database but are unable to.
+	ErrCorruptedStore = errors.New("corrupted ban store")
+
+	// ErrUnsupportedIP is an error returned when we attempt to parse an
+	// unsupported IP address type.
+	ErrUnsupportedIP = errors.New("unsupported IP type")
+)
+
+// Status gathers all of the details regarding an IP network's ban status.
+type Status struct {
+	// Banned determines whether the IP network is currently banned.
+	Banned bool
+
+	// Reason is the reason for which the IP network was banned.
+	Reason Reason
+
+	// Expiration is the absolute time in which the ban will expire.
+	Expiration time.Time
+}
+
+// Store is the store responsible for maintaining records of banned IP networks.
+// It uses IP networks, rather than single IP addresses, in order to coalesce
+// multiple IP addresses that are likely to be correlated.
+type Store interface {
+	// BanIPNet creates a ban record for the IP network within the store for
+	// the given duration. A reason can also be provided to note why the IP
+	// network is being banned. The record will exist until a call to Status
+	// is made after the ban expiration.
+	BanIPNet(*net.IPNet, Reason, time.Duration) error
+
+	// Status returns the ban status for a given IP network.
+	Status(*net.IPNet) (Status, error)
+}
+
+// NewStore returns a Store backed by a database.
+func NewStore(db walletdb.DB) (Store, error) {
+	return newBanStore(db)
+}
+
+// banStore is a concrete implementation of the Store interface backed by a
+// database.
+type banStore struct {
+	db walletdb.DB
+}
+
+// A compile-time constraint to ensure banStore satisfies the Store interface.
+var _ Store = (*banStore)(nil)
+
+// newBanStore creates a concrete implementation of the Store interface backed
+// by a database.
+func newBanStore(db walletdb.DB) (*banStore, error) {
+	s := &banStore{db: db}
+
+	// We'll ensure the expected buckets are created upon initialization.
+	err := walletdb.Update(db, func(tx walletdb.ReadWriteTx) error {
+		banStore, err := tx.CreateTopLevelBucket(banStoreBucket)
+		if err != nil {
+			return err
+		}
+		_, err = banStore.CreateBucketIfNotExists(banBucket)
+		if err != nil {
+			return err
+		}
+		_, err = banStore.CreateBucketIfNotExists(reasonBucket)
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
+
+// BanIPNet creates a ban record for the IP network within the store for the
+// given duration. A reason can also be provided to note why the IP network is
+// being banned. The record will exist until a call to Status is made after the
+// ban expiration.
+func (s *banStore) BanIPNet(ipNet *net.IPNet, reason Reason, duration time.Duration) error {
+	return walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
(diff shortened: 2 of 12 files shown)


banman/store.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		banIndex := banStore.NestedReadWriteBucket(banBucket)
+		if banIndex == nil {
+			return ErrCorruptedBanStore
+		}
+		reasonIndex := banStore.NestedReadWriteBucket(reasonBucket)
+		if banIndex == nil {
+			return ErrCorruptedBanStore
+		}
+		return addBannedPeer(
```
Roasbeef: Style nit: could use a new line before this return statement.

banman/store.go
```diff
@@ -0,0 +1,217 @@
...
+	Expiration time.Time
+}
+
+// BanStore is the store responsible for maintaining records of banned peers.
+type BanStore interface {
+	// BanPeer creates a ban record for the peer within the BanStore for the
+	// given duration. A reason can also be provided to note why the peer is
+	// being banned. The record will exist until a call to BanStatus is made
+	// after the ban expiration.
+	BanPeer(string, Reason, time.Duration) error
```
Roasbeef: I think we should ban based on `net.IP`, rather than a `string`. In the future, along with eclipse attack prevention measures, we may start to ban on a netblocks (IP address ranges). Example type we may grow to utilize: https://golang.org/pkg/net/#IPNet. If there's no `Mask` field set there, then it just target a single IP address.

banman/store.go
```diff
@@ -0,0 +1,217 @@
...
+		if banIndex == nil {
+			return ErrCorruptedBanStore
+		}
+
+		k := []byte(host)
+		status := fetchBanStatus(banIndex, reasonIndex, k)
+
+		// If the peer's ban duration has expired, we can remove its
+		// entry from the store.
+		if !time.Now().Before(status.Expiration) {
```
Roasbeef: Style nit: I usually find usage of `After` more readable, so: 
```go
if time.Now().After(status.Expiration)
```

blockmanager.go
```diff
@@ -710,15 +711,14 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 	// Ban any peer that responds with the wrong prev filter header.
 	for peer, msg := range headers {
 		if msg.PrevFilterHeader != *filterTip {
-			log.Infof("Banning peer=%v for invalid prev filter "+
-				"header", peer)
-
+			err := b.server.BanPeer(peer, banman.InvalidFilterHeader)
```
Roasbeef: Should either retain all the logging in-line, or move it into the `BanPeer` method.

blockmanager.go
```diff
@@ -951,13 +953,17 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
+				if err != nil {
+					log.Errorf("Unable to ban peer %v: %v",
+						peerAddr, err)
+				}
 
-				go func() {
-					b.server.BanPeer(sp)
-					sp.Disconnect()
-				}()
+				sp.Disconnect()
```
Roasbeef: IIRC that goroutines is a work around to prevent a deadlock when attempting to disconnect a peer in this particular context.

blockmanager.go
Roasbeef: Same goes for all the removed log statements in this commit.

blockmanager.go
wpaulino: I think the goroutine was actually only needed for `BanPeer` (see https://github.com/lightninglabs/neutrino/commit/49fcb98e64a6afa0621ff4163aed76a55cf11761), which is no longer an issue as it doesn't go through the server's main goroutine.

blockmanager.go
wpaulino: The log statements already exist within the `ChainService.BanPeer` wrapper method.

banman/store.go
wpaulino: That wouldn't handle the case of `time.Now() == status.Expiration`, though you could say it's not crucial.

banman/store.go
wpaulino: Doesn't seem like there's a difference between using `net.IP` and `string`? Though if we know there's a potential use case for `net.IPNet`, then I can update the PR to use that now instead.

sync_test.go
```diff
@@ -1483,13 +1484,23 @@ func checkRescanStatus() (int, int32, error) {
...
-			peer.Disconnect()
+		peerAddr := peer.Addr()
+		if peerAddr != harness.P2PAddress() {
+			continue
+		}
+
+		err := svc.BanPeer(peerAddr, banman.ExceededBanThreshold)
+		if err != nil {
+			if logLevel != btclog.LevelOff {
+				t.Logf("Unable to ban peer %v: %v", peerAddr,
```
valentinewallace: Why not `t.Fatalf`?

blockmanager.go
```diff
@@ -774,15 +774,17 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 				"headers", len(badPeers))
 
 			for _, peer := range badPeers {
-				log.Infof("Banning peer=%v for invalid filter "+
-					"headers", peer)
-
+				err := b.server.BanPeer(
+					peer, banman.InvalidFilterHeader,
```
valentinewallace: I noticed that peers that don't respond will still be banned for the `InvalidFilterHeader` reason here which could be a little misleading.

blockmanager.go
```diff
@@ -739,15 +740,14 @@ func (b *blockManager) getUncheckpointedCFHeaders(
...
-
+			err := b.server.BanPeer(peer, banman.InvalidFilterHeader)
+			if err != nil {
+				log.Errorf("Unable to ban peer %v: %v", peer, err)
+			}
 			sp := b.server.PeerByAddr(peer)
 			if sp != nil {
-				b.server.BanPeer(sp)
 				sp.Disconnect()
 			}
```
valentinewallace: Style nit: newline after this?

neutrino.go
```diff
@@ -209,26 +209,36 @@ func (sp *ServerPeer) addressKnown(na *wire.NetAddress) bool {
 // the score is above the ban threshold, the peer will be banned and
 // disconnected.
 func (sp *ServerPeer) addBanScore(persistent, transient uint32, reason string) {
-	// No warning is logged and no score is calculated if banning is disabled.
+	// No warning is logged and no score is calculated if banning is
```
valentinewallace: Reason `addBanScore` isn't used anywhere atm? Just curious..

banman/store.go
```diff
@@ -0,0 +1,219 @@
...
+	"net"
+	"time"
+
+	"github.com/btcsuite/btcwallet/walletdb"
+)
+
+var (
+	// byteOrder is the preferred byte order in which we should write things
+	// to disk.
+	byteOrder = binary.BigEndian
```
valentinewallace: Wondering if this is always needed when persisting numerical values to disk?

banman/reason.go
```diff
@@ -0,0 +1,43 @@
+package banman
+
+// Reason includes the different possible reasons which caused us to ban a peer.
+type Reason uint8
+
+// We prevent using `iota` to ensure the order does not have the value since
+// these are serialized within the database.
+const (
```
valentinewallace: Seems like some peers are `Unresponsive` which could be another `Reason`.

banman/store_test.go
```diff
@@ -0,0 +1,116 @@
...
+		cleanUp()
+		t.Fatalf("unable to create ban store: %v", err)
+	}
+
+	return banStore, cleanUp
+}
+
+// TestBanStore ensures that the BanStore's state correctly reflects the
+// BanStatus of peers.
+func TestBanStore(t *testing.T) {
```
valentinewallace: :+1:

sync_test.go
wpaulino: Done.

blockmanager.go
wpaulino: Seems like we'd have to return a ban reason from `detectBadPeers` since there are different scenarios that can cause them to get banned. I think that's fine in this case though because they previously did respond with what could've been an invalid filter header.

neutrino.go
wpaulino: Looks like Neutrino was able to respond to `getdata` messages from peers in a previous version and it used `addBanScore` there. That functionality was removed, but this remained.

banman/store.go
wpaulino: We're not just persisting numerical values in the store. Not sure I follow your question.

banman/reason.go
wpaulino: See other comment.

banman/store.go
Roasbeef: I outlined a precise use case above. Beyond that it adds stronger typing making the usage patterns of the new store more explicit. We may also want to have distinct bannig heuristics for say IPv4 vs IPv6 vs Tor. IPv4 is the most scarce so net block banning has a larger impact. Conversely for Tor hidden service addresses, making a new one is mostly costless (though rate limited a bit).

wpaulino: Pushed a new version that relies on `net.IPNet` as discussed, though it doesn't currently support Tor addresses due to the standard library only supporting IPv4 and IPv6 addresses. This can be done, but will require some more logic which could be done in a follow-up.

banman/store.go
```diff
@@ -0,0 +1,221 @@
...
+		banStore := tx.ReadWriteBucket(banStoreBucket)
+		if banStore == nil {
+			return ErrCorruptedStore
+		}
+		banIndex := banStore.NestedReadWriteBucket(banBucket)
+		if banIndex == nil {
+			return ErrCorruptedStore
+		}
+		reasonIndex := banStore.NestedReadWriteBucket(reasonBucket)
+		if banIndex == nil {
```
Roasbeef: Should be checking `reasonIndex`.

banman/store.go
```diff
@@ -0,0 +1,221 @@
...
```
Roasbeef: Should be checking `reasonIndex`.

banman/util.go
```diff
@@ -0,0 +1,51 @@
...
+	// networks from an address. This ensures that the IP network only
+	// contains the specified address.
+	defaultIPv6Mask = net.IPMask([]byte{
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+	})
+)
+
+// ParseIPNet parses the IP network that contains the given address. An optional
+// mask can be provided, to limit the scope of the IP network, otherwise the
```
Roasbeef: limit -> expand

banman/util.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"net"
+)
+
+var (
+	// defaultIPv4Mask is the default IPv4 mask used when parsing IP
+	// networks from an address. This ensures that the IP network only
+	// contains the specified address.
+	defaultIPv4Mask = net.IPv4Mask(0xff, 0xff, 0xff, 0xff)
```
Roasbeef: Not a blocker, but IMO, CIDR notation is easier to follow. So this would be a `/32` in that notation .

cfromknecht: See https://golang.org/pkg/net/#CIDRMask

wpaulino: Done using `net.CIDRMask`.

banman/util.go
wpaulino: Done.

banman/store.go
wpaulino: Done.

banman/store.go
wpaulino: Nice catch, done.

Issue #142, opened 2019-04-15
feature request: add litecoin support


losh11:
Neutrino currently does not support litecoin network, in either mainnet or testnet. It is frustrating to build litecoin software without even testnet support.

Roasbeef:
In order to do this, we'd need to abstract away all PoW checking methods and a number of other 
 components around difficulty retargeting. The alternative path is to maintain a fork.

Wowee0:
@losh11 @Roasbeef    Practically the one main change needed for Litecoin to work is btcutil/gcs/builder/builder.go needs the constant DefaultP changing from 19 to 20.  Is there a way to programmatically configure this on the fly without resorting to the entire fork of btcsuite (ltcsuite).  ?

Wowee0:
Having had time to think & discuss this some more (thanks @losh11), I realised that Builder.DefaultP does not need to be changed, as 'p' is passed in to most Builder methods.  So the handful of places Neutrino uses Builder could be changed to pass in an appropriate 'p' value dependent on whether Bitcoin or Litecoin is the active mode.  

@Roasbeef would it be acceptable to abstract&extend chainParams adding a coin identifier (Bitcoin/Litecoin) and a defaultP setting?  Also a change to checkProofOfWork for Litecoin mode.  I would like to submit a pull request to this effect.

cfromknecht:
@Wowee0 why would P value need to different for litecoin? Further, why would the P value be modified without considering the relationship with M?

Wowee0:
I did a lot of debugging in Neutrino to find out why received Litecoin cFilters would not match the expected checkpoint, eventually tracked it down to the output of builder.BuildBasicFilter.  Diffing gcs/builder.go between btcsuite/btcutil and ltcsuite/ltcutil/ indicates that the P value of 20 was used for LTCD (M was not changed).  After making the change on the Neutrino side they matched.  
That's the best I can do to answer your questions @cfromknecht.

cfromknecht:
Thanks @Wowee0, yes that does appear to be what happened and makes sense why they wouldn’t match up. 

That being said, my recommendation would be to correct the value to 19 in ltcsuite/ltcutil. Users that have cfindex enabled will need to drop and reindex, but that only needs to happen once. 

Given that there hasn’t been an official deployment on litecoin, I think would be best to keep the implementation consistent with BIP158. Note that if litecoind were to pull in changes from upstream, it would have M=19, and the two implementations would differ for the same filter type. 

[...]

PR #149, opened 2019-05-08
Implement feature request: add litecoin support (issue #142)

Wowee0:
Please could you review the following proposal for enabling litecoin support in Neutrino. It boils down to having litecoin genesis block details and using the ltcsuite CheckProofOfWork method.

Neutrino/blockmanager.go/CheckHeaderSanity :
if configured for litecoin,
copy the block header into a ltcutil.block and call ltcsuite.ltcd.blockchain.CheckProofOfWork

Neutrino/query.go/GetBlock :
if configured for litecoin,
copy the block into a ltcutil.block and call ltcsuite.ltcd.blockchain.CheckBlockSanity

ref: https://github.com/lightninglabs/neutrino/issues/142

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -19,6 +19,7 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	ltcblockchain "github.com/ltcsuite/ltcd/blockchain"
+	ltcwire "github.com/ltcsuite/ltcd/wire"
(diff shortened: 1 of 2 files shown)


kilrau:
May I ask why this was closed? @Wowee0

Wowee0:
> May I ask why this was closed?
Because we were told it would not be accepted, that we should fork Neutrino instead.  @kilrau

PR #150, opened 2019-05-14
pushtx/broadcaster: make rebroadcast tx async

halseth:
This commit makes the rebroadcast logic being run in a goroutine. This
is done to ensure we don't block the whole broadcast loop while
performing a rebroadcast.

Rebroadcasts can be slow if the number of transactions are large (each
broadcast has a timeout of 0.5s), which would cause publishing new
transactions to block. With this commit we let the rebroadcast continue
in the background, while we can keep pushing new txs.

### TODO
- [x] unit tests

### DIFF
--- a/pushtx/broadcaster.go
+++ b/pushtx/broadcaster.go
@@ -63,9 +63,6 @@ type Broadcaster struct {
 	// requests from external callers will be streamed through.
 	broadcastReqs chan *broadcastReq
 
-	// transactions is the set of transactions we have broadcast so far.
-	transactions map[chainhash.Hash]*wire.MsgTx
-
 	quit chan struct{}
 	wg   sync.WaitGroup
 }
@@ -75,7 +72,6 @@ func NewBroadcaster(cfg *Config) *Broadcaster {
 	b := &Broadcaster{
 		cfg:           *cfg,
 		broadcastReqs: make(chan *broadcastReq),
-		transactions:  make(map[chainhash.Hash]*wire.MsgTx),
 		quit:          make(chan struct{}),
 	}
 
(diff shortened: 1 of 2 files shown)


pushtx/broadcaster.go
```diff
@@ -130,31 +137,51 @@ func (b *Broadcaster) broadcastHandler(sub *blockntfns.Subscription) {
...
 		// Between blocks, we'll also try to attempt additional
 		// re-broadcasts to ensure a timely confirmation.
 		case <-reBroadcastTicker.C:
-			if len(b.transactions) == 0 {
+			select {
+			// If the rebroadcast semaphore is available, start a
+			// new goroutine to exectue a rebroadcast.
+			case <-rebroadcastSem:
+			default:
+				log.Debugf("Existing rebroadcast still in " +
```
wpaulino: Nit: this might be a little spammy, demote to trace?

pushtx/broadcaster.go
```diff
@@ -68,6 +68,8 @@ type Broadcaster struct {
 
 	quit chan struct{}
 	wg   sync.WaitGroup
+
+	sync.Mutex
```
wpaulino: Looks like we can avoid this mutex by instead making the copy of transactions before handing them over to the goroutine?

pushtx/broadcaster.go
halseth: This should only happen if a rebroadcast take more than 1min || block time. Should be fine?

pushtx/broadcaster.go
halseth: Fixed! PTAL

PR #151, opened 2019-05-20
headerfs: add ability to assert the state of the filter header chain

Roasbeef:
In this commit, we add the ability to specify an expected filter header
state when the filter header store is being created. As the filter
headers aren't committed yet, it's at times useful to nudge existing
nodes towards a filter header chain that they themselves may not have
obtained previously.

### DIFF
--- a/headerfs/store.go
+++ b/headerfs/store.go
@@ -81,7 +81,7 @@ var headerBufPool = sync.Pool{
 type headerStore struct {
 	mtx sync.RWMutex
 
-	filePath string
+	fileName string
 
 	file *os.File
 
@@ -122,7 +122,7 @@ func newHeaderStore(db walletdb.DB, filePath string,
 	}
 
 	return &headerStore{
-		filePath:    filePath,
+		fileName:    flatFileName,
 		file:        headerFile,
 		headerIndex: index,
 	}, nil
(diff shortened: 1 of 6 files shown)


headerfs/store.go
```diff
@@ -655,6 +656,26 @@ func NewFilterHeaderStore(filePath string, db walletdb.DB,
...
 	}
 
+	// If we have a state assertion then we'll check it now to see if we
+	// need to modify our filter header files before we proceed.
+	if headerStateAssertion != nil {
+		assertedHeader, err := fhs.FetchHeaderByHeight(
+			headerStateAssertion.Height,
+		)
+		if err != nil {
+			return nil, err
```
wpaulino: Not sure how we intend to deploy this flag within `lnd`, but if this is set for new clients whom have yet to fetch this filter header from the network then they won't be able to start. In that case, seems like we'd want to keep the assertion reference around in the `blockManager` as well?

Roasbeef:
New version up PTAL @wpaulino.

Roasbeef:
Confirmed that this works as is. The reason it works is that on re-start, it'll re-insert the "genesis filter hash", and then sync anew from there. Ready to review as is.

PR #157, opened 2019-06-18
query: handle closing matchSignal for peer query

halseth:
We did not check whether the channel was actually sent on or closed,
which could cause us to believe the query was answered while in reality
the peer was about to shut down.

### DIFF
--- a/query.go
+++ b/query.go
@@ -385,13 +385,16 @@ func queryChainServiceBatch(
 			mtxPeerStates.Lock()
 			peerStates[sp.Addr()] = queryMsgs[handleQuery]
 			mtxPeerStates.Unlock()
+
+			exiting := false
 			select {
 			case <-queryQuit:
-				return
+				exiting = true
 			case <-s.quit:
-				return
+				exiting = true
 			case <-quit:
-				return
+				exiting = true
+
 			case <-timeout:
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -400,7 +400,11 @@ func queryChainServiceBatch(
 					newLogClosure(func() string {
 						return spew.Sdump(queryMsgs[handleQuery])
 					}))
-			case <-matchSignal:
+			case _, ok := <-matchSignal:
+				if !ok {
+					return
```
wpaulino: Shouldn't we be resetting the state of this query to `queryWaitSubmit` so that it can be assigned to a different peer? This might also apply to some of other channel reads within this `select` block.

halseth: Good catch, that seems to be the case! Fixed it for the affected cases.

query.go
```diff
@@ -381,13 +381,16 @@ func queryChainServiceBatch(
 			mtxPeerStates.Lock()
 			peerStates[sp.Addr()] = queryMsgs[handleQuery]
 			mtxPeerStates.Unlock()
+
+			exiting := false
```
wpaulino: Non-blocking comment: as discussed offline, could make this be the query state instead so that we only update it in one place.

PR #158, opened 2019-06-18
query: increase default timeout from 3s -> 10s

halseth:


blockmanager.go
```diff
@@ -37,6 +37,10 @@ const (
 	// we're able to properly handle re-orgs in size strictly less than
 	// this value.
 	numMaxMemHeaders = 10000
+
+	// retryTeimout is the time we'll wait between failed queries to fetch
```
wpaulino: Nit: `retryTimeout`

halseth: Fixed

PR #159, opened 2019-06-18
Check filter header checkpoints against hardcoded list

halseth:
Before starting to compare filter checkpoints our peers are serving, we
do a quick check against a list of filter header checkpoints we have.

Note that this PR doesn't actually add the checkpoints, these should be cross checked from different nodes.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -21,6 +21,7 @@ import (
 	"github.com/btcsuite/btcutil/gcs"
 	"github.com/btcsuite/btcutil/gcs/builder"
 	"github.com/lightninglabs/neutrino/blockntfns"
+	"github.com/lightninglabs/neutrino/chainsync"
 	"github.com/lightninglabs/neutrino/headerfs"
 	"github.com/lightninglabs/neutrino/headerlist"
 )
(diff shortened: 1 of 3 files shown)


chainsync/filtercontrol.go
```diff
@@ -0,0 +1,57 @@
...
+
+	control, ok := filterHeaderCheckpoints[params.Net]
+	if !ok {
+		return nil
+	}
+
+	hash, ok := control[height]
+	if !ok {
+		return nil
+
```
wpaulino: Nit: extra newline.

blockmanager.go
```diff
@@ -1215,6 +1216,36 @@ func (b *blockManager) resolveConflict(
...
+				log.Warnf("Banning peer=%v since served "+
+					"checkpoints didn't match our "+
+					"checkpoint at height %d", peer, height)
+
+				sp := b.server.PeerByAddr(peer)
+				if sp != nil {
+					b.server.BanPeer(sp)
+					sp.Disconnect()
+				}
+				delete(checkpoints, peer)
```
wpaulino: Should continue to next peer here.

halseth: Good catch! Fixed: https://github.com/lightninglabs/neutrino/pull/159/commits/e47c5a97b1986d08ba872bec23f257538522a023

chainsync/filtercontrol.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/159/commits/47bc7395491bb6b7ca562a2eeaaf951e59efa8a0

halseth:
PTAL @wpaulino

PR #161, opened 2019-06-19
query: add peer cooldown

halseth:
This commit adds a peer cooldown to the query logic. This is done to
avoid giving the same query immediately to the peer that timed out,
which could cause sync to stall on a bad peer.


halseth:
Alternative to https://github.com/lightninglabs/neutrino/pull/145. Fixes the same issue.

PR #162, opened 2019-06-20
query: use per-query matchSignals.

halseth:
This fixes an issue that could happen if a response to a previously
timed out query was received during sync. This could cause the response
handler to send on the peer's matchSignal, but the peerGoroutine was no
longer listening on the channel, causing sync to stall.

Instead we tie the matchSignal channel to the specific query, such that
we know that we only send the signal if the current query is
successfully handled, and we buffer the channel to ensure no blocking
even after a timeout.

### DIFF
--- a/query.go
+++ b/query.go
@@ -297,9 +297,20 @@ func queryChainServiceBatch(
 	}
 	defer close(subQuit)
 
+	// peerState holds a query message and an answer channel that get a
+	// notice when the query got a match. If it's the peer's match, the
+	// peer can mark the query a success and move on to the next query
+	// ahead of timeout.
+	type peerState struct {
+		msg         wire.Message
+		matchSignal chan struct{}
+	}
+
 	// peerStates and its companion mutex allow the peer goroutines to
-	// tell the main goroutine what query they're currently working on.
-	peerStates := make(map[string]wire.Message)
+	// tell the main goroutine what query they're currently working on, and
+	// for the main goroutine to signal the peer when a match has been
+	// received.
+	peerStates := make(map[string]peerState)
 	var mtxPeerStates sync.RWMutex
 
 	// allDone is a helper closure we'll use to determine whether we can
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -373,11 +373,17 @@ func queryChainServiceBatch(
...
-					if sp.Connected() {
-						continue
-					} else {
-						return
-					}
+
+				// We also check the match signal, since we
+				// might receive an answer for a query that
+				// previously timed out.
+				case <-matchSignal:
```
wpaulino: This seems weird to me just from a high level overview. Why would a `matchSignal` be triggered for this peer if it's no longer handling a query? I took a look to see why this would be the case, and it seems to be because we don't modify the `peerStates` map, which tracks the current query a peer is working on. Shouldn't we be clearing a peer's entry within this map once a query times out or matches? Doing this will then require another change further down below when receiving a response through `msgChan`, which is that the current query might not be the one that prompted the response being received.

halseth: Yeah, you are right, this solution was not bulletproof. See new commit message for amended approach.

query.go
```diff
@@ -484,24 +496,35 @@ func queryChainServiceBatch(
...
 				close(quitChan)
-				close(matchSignals[peer])
 				delete(peerQuits, peer)
-				delete(matchSignals, peer)
+
+				// Delete the current state of the peer, so we
+				// won't process any lingering responses from
+				// it.
+				mtxPeerStates.Lock()
+				delete(peerStates, peer)
```
wpaulino: Would it be better to do this within `peerGoroutine` so that all modifications to this map happen there?

query.go
```diff
@@ -382,8 +392,12 @@ func queryChainServiceBatch(
 			}
 
 			// We have a query we're working on.
+			matchSignal := make(chan struct{}, 1)
```
wpaulino: :+1:

query.go
halseth: Done: 3484b86

halseth:
PTAL @wpaulino

PR #163, opened 2019-06-28
query: exit if all queries are answered after receiving a response

wpaulino:
Doing so prevents us from waiting until the query's timeout is reached if we happened to receive the last response in order to satisfy all of our in-flight queries.

### DIFF
--- a/query.go
+++ b/query.go
@@ -302,8 +302,25 @@ func queryChainServiceBatch(
 	peerStates := make(map[string]wire.Message)
 	var mtxPeerStates sync.RWMutex
 
+	// allDone is a helper closure we'll use to determine whether we can
+	// exit due to all of our queries being answered.
+	allDone := func() bool {
+		for i := 0; i < len(queryStates); i++ {
+			if atomic.LoadUint32(&queryStates[i]) !=
+				uint32(queryAnswered) {
+				return false
+			}
+		}
+		return true
+	}
+
+	// allDoneSignal is a channel we'll close to signal to the main loop
+	// that all of the queries have been answered.
+	var allDoneOnce sync.Once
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -519,20 +543,10 @@ func queryChainServiceBatch(
 				}
 			}
 		case <-time.After(qo.timeout):
-			// Check if we're done; if so, quit.
```
Roasbeef: So the timeout send it just a noop now?

query.go
```diff
@@ -442,13 +458,20 @@ func queryChainServiceBatch(
...
 				// query a success.
 				atomic.StoreUint32(&queryStates[handleQuery],
 					uint32(queryAnswered))
 
-				log.Tracef("Query #%v answered, updating state",
-					handleQuery)
+				// If we're done answering all of our queries,
+				// we can exit now.
+				if allDone() {
+					close(allDoneSignal)
```
Roasbeef: Why not check this in the main loop when we get a response from a goroutine?

wpaulino: It’s race-y because we can’t guarantee that the query state has been updated by the time we check it.

query.go
wpaulino: It already was if we weren’t done with all the queries, so it’d just wait again.  There’s no need to check if we’re done anymore since it’s handled when receiving responses.

Roasbeef: Hmm so then timeouts haven't worked at all this whole time.

query.go
Roasbeef: How is this less racy? In both instances all indexes are checked used atomics. Also it's possible that two goroutines finish at nearly the same time, and cause a panic by double closing the final since they don' tall have their own signal.

PR #170, opened 2019-07-18
query: reset the main timeout each time a message is received

### LINKED ISSUE #164: Failed to complete long rescans.
I reproduced the problem initially when syncing a node that was offline for 2 months and needed to fetch several thousands of filters.
The problem has started to show up after the change to query compact filters in batches (1000 each).

At some point I get this error:

“Peer 94.23.154.171:8333 (outbound) appears to be stalled or misbehaving, block timeout -- disconnecting”,
After disconnecting neutrino doesn't further attempt to connect and is not functioning. There is a suggestion on how to recover from this state here (https://github.com/lightninglabs/neutrino/issues/138).

I have noticed in the logs that some queries (like getdata to fetch a block) weren't answered on time.
I raised the QueryBatchTimeout from 30 seconds to 100 seconds and I was able to sync but this was just to understand the problem. I don't think there is a right number here for the timeout as we don't really know the batch size.

The godoc comment for this field is:

```
// QueryBatchTimout is the total time we'll wait for a batch fetch
// query to complete.
// TODO(halseth): instead use timeout since last received response?
```

Could be that the last line might be the solution for this problem.

Roasbeef:
In this commit, we aim to fix an issue that can cause a query to fail
all together for a large query over a possibly high latency link. Before
this commit, we if attempted to query for several filters at once, on
mobile phones, it's possible that we timeout part way through. As is,
we'll then exit, and fail to properly retry, messing up the main rescan
loop.

In this commit, we aim to address this by _resetting_ the `peerTimeout`
each time we receive a message. Rather than aim to complete the entire
query in the timeout interval, we instead modify the timeout to ensure
forward progress of the query. If the peer isn't responding, then the
`peerTimeout` will tick, and we'll switch to another peer.

As a result, it's now possible that a query takes longer than
`QueryPeerConnectTimeout` if a peer is slow, or we're using a high
latency link, or have a meager CPU. As a result, we'll now _stop_ the
now named `connectionTimeout` once we receive the first message from the
peer. If the peer stops responding all together, then the `peerTimeout`
will be ticked. Arguably, we don't even need the `connectionTimout` due
to this new logic.

Fixes #164.

### DIFF
--- a/query.go
+++ b/query.go
@@ -723,8 +723,9 @@ func queryChainServicePeers(
 	// Loop for any messages sent to us via our subscription channel and
 	// check them for whether they satisfy the query. Break the loop if
 	// it's time to quit.
-	peerTimeout := time.NewTicker(qo.timeout)
-	timeout := time.After(qo.peerConnectTimeout)
+	peerTimeout := time.NewTimer(qo.timeout)
+	connectionTimeout := time.NewTimer(qo.peerConnectTimeout)
+	connectionTicker := connectionTimeout.C
 	if queryPeer != nil {
 		peerTries[queryPeer.Addr()]++
 		queryPeer.subscribeRecvMsg(subscription)
@@ -733,7 +734,7 @@ func queryChainServicePeers(
 checkResponses:
 	for {
 		select {
-		case <-timeout:
+		case <-connectionTicker:
 			// When we time out, we're done.
 			if queryPeer != nil {
 				queryPeer.unsubscribeRecvMsgs(subscription)
@@ -762,6 +763,22 @@ checkResponses:
 			// stuck. This is a caveat for callers that should be
 			// fixed before exposing this function for public use.
 			checkResponse(sm.sp, sm.msg, queryQuit)
+
+			// Each time we receive a response from the current
+			// peer, we'll reset the main peer timeout as they're
+			// being responsive.
+			if !peerTimeout.Stop() {
+				<-peerTimeout.C
+			}
+			peerTimeout.Reset(qo.timeout)
+
+			// Also at this point, if the peerConnectTimeout is
+			// still active, then we can disable it, as we're
+			// receiving responses from the current peer.
+			if connectionTicker != nil && !connectionTimeout.Stop() {
+				<-connectionTimeout.C
+			}
+			connectionTicker = nil
 
 		// The current peer we're querying has failed to answer the
 		// query. Time to select a new peer and query it.


Roasbeef:
Note that we're planning a complete re-write of  `query.go`, so the aim of this PR is to mitigate the current issues in this area of the codebase since it can directly impact syncing/rescanning.

query.go
```diff
@@ -763,6 +763,21 @@ checkResponses:
...
+			// being responsive.
+			if !peerTimeout.Stop() {
+				<-peerTimeout.C
+			}
+			peerTimeout.Reset(qo.timeout)
+
+			// Also at this point, if the peerConnectTimeout is
+			// still active, then we can disable it, as we're
+			// receiving responses from the current peer.
+			if connectionTimeout != nil && !connectionTimeout.Stop() {
```
roeierez: Nit: Can remove `connectionTimeout != nil` from the condition.

query.go
```diff
@@ -763,6 +763,21 @@ checkResponses:
...
+				<-connectionTimeout.C
```
roeierez: This will hang on the second time it is called (in the second message).

roeierez:
I think after this change the `QueryBatchTimeout` is not needed and instead the regular `QueryTimeout` can be used.

query.go
Roasbeef: Have you seen this happen? Or is this a premonition? 

Reworked slightly to rely on the `nil` check as was originally intended.

query.go
```diff
@@ -763,6 +764,22 @@ checkResponses:
...
+			}
+			peerTimeout.Reset(qo.timeout)
+
+			// Also at this point, if the peerConnectTimeout is
+			// still active, then we can disable it, as we're
+			// receiving responses from the current peer.
+			if connectionTicker != nil && !connectionTimeout.Stop() {
+				<-connectionTimeout.C
+			}
+			connectionTicker = nil
```
wpaulino: Shouldn't we re-establish this ticker once we lose our current `queryPeer`?

query.go
wpaulino: From the godoc:
> Stop prevents the Timer from firing. It returns true if the call stops the timer, false if the timer has already expired or been stopped.

Looks like it would indeed hang because false is returned on the second `Stop` call, but this second call won't happen in the latest diff due to setting the ticker to `nil` below.

query.go
wpaulino: > As a result, we'll now _stop_ the now named `connectionTimeout` once we receive the first message from the peer. If the peer stops responding all together, then the `peerTimeout` will be ticked. Arguably, we don't even need the `connectionTimout` due to this new logic.

Why not just remove it then?

roeierez:
@Roasbeef I have tested the latest changes with an old node that was several thousands of filters behind tip. I intentionally reduced the timeout to 4 seconds with high latency link.
There were no retries during the batch fetch and the rescan finished successfully. So  it looks like the issue is fixed.

roeierez:
Update:
I tried the patch several more times and unfortunately got the stale behavior again but not every time.
More research showed that there were actually two problems, the first one originated in the client hitting timeout and is solved by this patch.
The second problem I see is that if the batch response is not finished by exactly 40 seconds (no retries on the client) I get this error and the peer disconnects. I know that btcd has his own stale control and this might cause the behavior I see.
Regardless if there should be a fix in btcd or not it seems that 1000 filters (~15M?) in a batch are too much for an average mobile client to handle in 40 seconds.
By changing the code to request 100 filters per batch the problem wasn't reproduced. The cost for this would be additional 10 roundtrips for every 1000 filters which doesn't sound like a big performance penalty considering the use case . It could can be a reasonable mitigation for the issue unless the 40 seconds sound familiar to someone and can find a more robust solution for this problem.

Roasbeef:
@roeierez thanks for the update! I think we're going ahead with this PR since it addresses the first issue. As far as the "40 second" problem, we'll need to look into that a bit more, but may just defer addressing it to the full overhaul we have planned for this section of the codebase.

query.go
Roasbeef: Wanted to make the smallest change possible in this area to resolve the issue.

Roasbeef:
> The second problem I see is that if the batch response is not finished by exactly 40 seconds (no retries on the client) I get this error and the peer disconnects. I know that btcd has its own stale control and this might cause the behavior I see.


This is curious, since the current logic more or less says: "as long as you give me a filter every 30 seconds, then i'll wait for the rest, no matter how long it takes".

roeierez:
> This is curious, since the current logic more or less says: "as long as you give me a filter every 30 seconds, then i'll wait for the rest, no matter how long it takes".

Thats why I wrote it is probably not the client responsible for this error and it looks like there is a 40 seconds restriction somewhere else.

PR #174, opened 2019-07-25
Wait for filters batch to complete before returning the requested filter.

roeierez:
**The problem**
This PR follows https://github.com/lightninglabs/neutrino/pull/170 that has fixed the case where neutrino disconnects due to timeout during a batch request.
That PR fixed one aspect of the problem and as noted here https://github.com/lightninglabs/neutrino/pull/170#issuecomment-514043210, more tests have shown there is probably an additional mechanism that disconnects the remote peer.
Since neutrino is using the same peer.go file that btcd uses (https://github.com/btcsuite/btcd/blob/master/peer/peer.go) to handle peer communication it inherits also the staleHandler (https://github.com/btcsuite/btcd/blob/c26ffa870fd817666a857af1bf6498fabba1ffe3/peer/peer.go#L1174) that tracks every sent request and disconnects peers that don't send a response in some time limit (30 seconds as basis).

The following log line is printed from neutrino in such situations just before the disconnection:

`019-07-17 15:05:21.157 [DBG] BTCN: Peer 94.23.154.171:8333 (outbound) appears to be stalled or misbehaving, notfound timeout -- disconnecting.`

The use case I encountered (in LND) that triggers such problem is as follows:

1. LND starts, the initial rescan starts, download 1000 initial filters batch.
2. During the batch response messages, one of the filters triggers block download.
3. A request for this block is sent (`getdata` message) to btcd peer.
4. btcd won't serve the getdata request until all filter messages that were already queued for response are flushed out.
5. the btcd peer is disconnected by neutrino when the stallHandler (https://github.com/btcsuite/btcd/blob/c26ffa870fd817666a857af1bf6498fabba1ffe3/peer/peer.go#L1278) detects that `getdata` didn't complete on time because sending the batch haven't been completed.

To me it looks like the fundamental issue is that btcd serves peer requests at FIFO order while neutrino doesn't enforce such order on requests sent to every peer, causing some request to be blocked until previous ones are processed, resulting in stall and disconnection.
At the previous PR (https://github.com/lightninglabs/neutrino/pull/170 ) @Roasbeef noted that there are plans to refactor the `query.go`, so maybe this information can be used in the new design. 

**The solution**
This PR mitigates the rescan issue by waiting for the batch to complete before returning the filter to the caller.
By that it ensures, to some extent, that no other requests will be sent to the remote peer until the batch response is fully received.

### DIFF
--- a/query.go
+++ b/query.go
@@ -1199,20 +1199,26 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 	}()
 
 	var ok bool
-	select {
+	var resultFilter *gcs.Filter
 
-	// We'll return immediately to the caller when the filter arrives.
-	case filter, ok = <-query.filterChan:
-		if !ok {
+	// We will wait for the query to finish before we return the requested
+	// filter to the caller.
+	for {
+		select {
+
+		case filter, ok = <-query.filterChan:
+			if !ok {
+				// Query has finished, if we have a result we'll return it.				
+				return resultFilter, nil
+			}
+
+			// We'll store the filter so we can return it later to the caller.
+			resultFilter = filter
+
+		case <-s.quit:
 			// TODO(halseth): return error?
 			return nil, nil
 		}
-
-		return filter, nil
-
-	case <-s.quit:
-		// TODO(halseth): return error?
-		return nil, nil
 	}
 
 }


query.go
```diff
@@ -1199,20 +1199,31 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
-	case filter, ok = <-query.filterChan:
-		if !ok {
+	// We will wait for the query to finish before we return the requested
+	// filter to the caller.
+	for {
+		select {
+
+		case filter, ok = <-query.filterChan:
+			if !ok {
+
```
wpaulino: Nit: extra newline.

query.go
```diff
@@ -1199,20 +1199,31 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+				// query has finished, if we have a result we'll return it.
+				if resultFilter == nil {
+					// TODO(halseth): return error?
+					return nil, nil
+				}
+				return resultFilter, nil
```
wpaulino: Nit: can just do `return resultFilter, nil`, no need for the `if` above.

query.go
roeierez: Removed.

query.go
roeierez: Fixed.

query.go
```diff
@@ -1199,20 +1199,26 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
-	case filter, ok = <-query.filterChan:
-		if !ok {
+	// We will wait for the query to finish before we return the requested
+	// filter to the caller.
+	for {
+		select {
+
+		case filter, ok = <-query.filterChan:
+			if !ok {
+				// query has finished, if we have a result we'll return it.
```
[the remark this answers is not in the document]
roeierez: Fixed.

query.go
```diff
@@ -1199,20 +1199,26 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	for {
+		select {
+
+		case filter, ok = <-query.filterChan:
+			if !ok {
+				// query has finished, if we have a result we'll return it.				
+				return resultFilter, nil
+			}
+
+			// We'll store the filter so we can return it later to the caller
```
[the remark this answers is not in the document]
roeierez: Fixed.

Issue #166, opened 2019-07-15
Failure to fetch filter during rescan causes incorrect order of notifications.


roeierez:
During a rescan there is an attempt to fetch the filter for each "connected" block (https://github.com/lightninglabs/neutrino/blob/master/rescan.go#L558). If this fetch fails the code schedules a timer for refetching the filter. (https://github.com/lightninglabs/neutrino/blob/master/rescan.go#L579). 
The issue occurs when before this fetch is completed new blocks are processed triggering out of order notifications.
I have seen this affects btcwallet, for example, where it fails to call "SetSyncedTo" with the new out of order block notification (https://github.com/btcsuite/btcwallet/blob/master/wallet/chainntfns.go#L210).
At such cases the log shows these lines:

`2019-07-14 14:05:21.136 [INF] NTFN: New block: height=585349, sha=00000000000000000010d8e8d0de3356db8d1d5f9b46f68a030c48ca23292a4e
2019-07-14 14:05:21.136 [INF] UTXN: Attempting to graduate height=585349: num_kids=0, num_babies=0
2019-07-14 14:05:21.193 [DBG] BTCN: Received cfilter from 94.23.154.171:8333 (outbound)
2019-07-14 14:05:21.193 [DBG] BTCN: Fetching filters for heights=[585351, 585351], stophash=00000000000000000007f23985e494d8312804dab1aa0269edb1610494de2c0f
2019-07-14 14:05:21.194 [DBG] BTCN: Sending getcfilters to 94.23.154.171:8333 (outbound)
2019-07-14 14:05:21.198 [ERR] LNWL: Unable to process chain backend block connected notification: failed to store sync information 00000000000000000019d62f51721192d1b03f993383a6ac04053b10ff9efdea: failed to fetch block hash for height 585349: block not found`

This case is not as rare as it might seems and happens on mobile devices from time to time, for example:
1. User node is synced and then the app is sent to the background. After a while app stops getting network resources.
2. User moves app to the foreground after several hours and the OS gives the app network access.
3. Neutrino detects new blocks start sending notifications.
4. app goes back to background again and network become not accessible again.
5. Filter fetching fails, refetch is scheduled and the races starts.

Note:
I might be mistaken here but it seems that the contract between neutrino and its subscribers is that block notifications are sent at block height order. If this is true, maybe a better approach in case of failure to fetch the filter would be buffer new upcoming connected blocks until the filter is refetches and notification is sent.

PR #175, opened 2019-08-07
rescan: stash blocks to retry after filter fetch failure

wpaulino:
In this commit, we address a lingering issue within the Rescan struct where it was possible for it to deliver notifications out of order. This was due to the fact that retrying block filter fetches replaced attempts for previous blocks, causing us to move forward without dispatching notifications for those previous blocks. We'll now continue to retry fetching a block filter until successful or stale, and new blocks that come in while doing so will be stashed and retried later on in the order that they arrived.

Along the way, to ensure the correctness of this behavior, a test harness was created to aid testing. The test harness accounts for most of the diff in this PR.

Fixes https://github.com/lightninglabs/neutrino/issues/166.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -1019,7 +1019,7 @@ func (s *ChainService) rollBackToHeight(height uint32) (*headerfs.BlockStamp, er
 	}
 
 	for uint32(bs.Height) > height {
-		header, _, err := s.BlockHeaders.FetchHeader(&bs.Hash)
+		header, headerHeight, err := s.BlockHeaders.FetchHeader(&bs.Hash)
 		if err != nil {
 			return nil, err
 		}
--- a/query.go
+++ b/query.go
@@ -1224,8 +1224,9 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 		// If there are elements left to receive, the query failed.
 		if len(query.headerIndex) > 0 {
 			numFilters := query.stopHeight - query.startHeight + 1
+			numRecv := numFilters - int64(len(query.headerIndex))
 			log.Errorf("Query failed with %d out of %d filters "+
-				"received", len(query.headerIndex), numFilters)
+				"received", numRecv, numFilters)
 			return
 		}
 	}()
@@ -1240,11 +1241,13 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 
 		case filter, ok = <-query.filterChan:
 			if !ok {
-				// Query has finished, if we have a result we'll return it.				
+				// Query has finished, if we have a result we'll
+				// return it.
 				return resultFilter, nil
 			}
 
-			// We'll store the filter so we can return it later to the caller.
+			// We'll store the filter so we can return it later to
+			// the caller.
 			resultFilter = filter
 
 		case <-s.quit:
--- a/rescan.go
+++ b/rescan.go
@@ -480,12 +480,6 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		reFetchMtx.Unlock()
 	}
 
-	// We'll need to keep track of whether we are current with the chain in
-	// order to properly recover from a re-org. We'll start by assuming that
-	// we are not current in order to catch up from the starting point to
-	// the tip of the chain.
-	current := false
-
 	// handleBlockConnected is a closure that handles a new block connected
 	// notification.
 	//
@@ -497,7 +491,6 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// and notify the callers of blocks we may have missed.
 		header := ntfn.Header()
 		if header.PrevBlock != curStamp.Hash {
-			current = false
 			return fmt.Errorf("out of order block %v: expected "+
 				"PrevBlock %v, got %v", header.BlockHash(),
 				curStamp.Hash, header.PrevBlock)
@@ -641,6 +634,12 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		return nil
 	}
 
+	// We'll need to keep track of whether we are current with the chain in
+	// order to properly recover from a re-org. We'll start by assuming that
+	// we are not current in order to catch up from the starting point to
+	// the tip of the chain.
+	current := false
+
 	// Loop through blocks, one at a time. This relies on the underlying
 	// chain source to deliver notifications in the correct order.
 rescanLoop:
@@ -706,6 +705,14 @@ rescanLoop:
 					if err != nil {
 						log.Errorf("Unable to process "+
 							"%v: %v", ntfn, err)
+
+						// Since we weren't able to
+						// successfully process the
+						// block, we'll set ourselves to
+						// not be current in order to
+						// attempt catching up with the
+						// chain ourselves.
+						current = false
 					}
 
 				case *blockntfns.Disconnected:
--- a/rescan.go
+++ b/rescan.go
@@ -495,20 +495,12 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// If we've somehow missed a header in the range, then we'll
 		// mark ourselves as not current so we can walk down the chain
 		// and notify the callers of blocks we may have missed.
-		//
-		// It's possible due to the nature of the current subscription
-		// system that we get a duplicate block. We'll catch this and
-		// continue forwards to avoid an unnecessary state transition
-		// back to the !current state.
 		header := ntfn.Header()
-		if header.PrevBlock != curStamp.Hash &&
-			header.BlockHash() != curStamp.Hash {
-
+		if header.PrevBlock != curStamp.Hash {
 			current = false
-			return fmt.Errorf("out of order block %v: "+
-				"expected PrevBlock %v, got %v",
-				header.BlockHash(), curStamp.Hash,
-				header.PrevBlock)
+			return fmt.Errorf("out of order block %v: expected "+
+				"PrevBlock %v, got %v", header.BlockHash(),
+				curStamp.Hash, header.PrevBlock)
 		}
 
 		// Do not process block until we have all filter headers. Don't
@@ -517,30 +509,25 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// notification, then we can re-process it without any issues.
 		nextBlockHeight := uint32(curStamp.Height + 1)
 		_, err := chain.GetFilterHeaderByHeight(nextBlockHeight)
-		if header.BlockHash() != curStamp.Hash && err != nil {
+		if err != nil {
 			log.Warnf("Missing filter header for height=%v, "+
 				"skipping", curStamp.Height+1)
 
 			return nil
 		}
 
-		// As this could be a re-try, we'll ensure that we don't
-		// incorrectly increment our current time stamp.
-		if curStamp.Hash != header.BlockHash() {
-			curHeader = header
-			curStamp.Hash = header.BlockHash()
-			curStamp.Height++
+		newStamp := headerfs.BlockStamp{
+			Hash:   header.BlockHash(),
+			Height: int32(nextBlockHeight),
 		}
 
-		log.Tracef("Rescan got block %d (%s)", curStamp.Height,
-			curStamp.Hash)
+		log.Tracef("Rescan got block %d (%s)", newStamp.Height,
+			newStamp.Hash)
 
 		// We're only scanning if the header is beyond the horizon of
 		// our start time.
 		if !scanning {
-			scanning = ro.startTime.Before(
-				curHeader.Timestamp,
-			)
+			scanning = ro.startTime.Before(header.Timestamp)
 		}
 
 		// If we're not scanning or our watch list is empty, then we can
@@ -548,15 +535,18 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		if !scanning || len(ro.watchList) == 0 {
 			if ro.ntfn.OnFilteredBlockConnected != nil {
 				ro.ntfn.OnFilteredBlockConnected(
-					curStamp.Height, &curHeader, nil,
+					newStamp.Height, &header, nil,
 				)
 			}
 			if ro.ntfn.OnBlockConnected != nil {
 				ro.ntfn.OnBlockConnected(
-					&curStamp.Hash, curStamp.Height,
-					curHeader.Timestamp,
+					&newStamp.Hash, newStamp.Height,
+					header.Timestamp,
 				)
 			}
+
+			curHeader = header
+			curStamp = newStamp
 
 			return nil
 		}
@@ -565,7 +555,7 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// relevant transactions and notify them.
 		queryOptions := NumRetries(0)
 		blockFilter, err := chain.GetCFilter(
-			curStamp.Hash, wire.GCSFilterRegular, queryOptions,
+			newStamp.Hash, wire.GCSFilterRegular, queryOptions,
 		)
 
 		switch {
@@ -590,11 +580,16 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		}
 
 		err = notifyBlockWithFilter(
-			chain, ro, &curHeader, &curStamp, blockFilter,
+			chain, ro, &header, &newStamp, blockFilter,
 		)
 		if err != nil {
 			return err
 		}
+
+		// With the block successfully notified, we'll advance our state
+		// to it.
+		curHeader = header
+		curStamp = newStamp
 
 		// We've successfully fetched this current block, so we'll reset
 		// the retry timer back to nil.
--- a/rescan.go
+++ b/rescan.go
@@ -31,6 +31,10 @@ var (
 	// ErrRescanExit is an error returned to the caller in case the ongoing
 	// rescan exits.
 	ErrRescanExit = errors.New("rescan exited")
+
+	// errRetryBlock is an internal error used to signal to the rescan
+	// should it should attempt to retry processing a block.
+	errRetryBlock = errors.New("block must be retried")
 )
 
 // ChainSource is an interface that's in charge of retrieving information about
@@ -198,6 +202,66 @@ func QuitChan(quit <-chan struct{}) RescanOption {
 	return func(ro *rescanOptions) {
 		ro.quit = quit
 	}
+}
+
+// blockRetryQueue is a helper struct that maintains a queue of blocks for which
+// we need to fetch filters.
+type blockRetryQueue struct {
+	blocks []*blockntfns.Connected
+}
+
+// newBlockRetryQueue constructs a new, empty retry block queue.
+func newBlockRetryQueue() *blockRetryQueue {
+	return &blockRetryQueue{}
+}
+
+// push enqueues a block at the end of the queue.
+func (q *blockRetryQueue) push(block *blockntfns.Connected) {
+	q.blocks = append(q.blocks, block)
+}
+
+// peek returns the next block to be retried but doesn't consume it.
+func (q *blockRetryQueue) peek() *blockntfns.Connected {
+	if len(q.blocks) == 0 {
+		return nil
+	}
+	return q.blocks[0]
+}
+
+// pop returns and consumes the next block to be retried.
+func (q *blockRetryQueue) pop() *blockntfns.Connected {
+	if len(q.blocks) == 0 {
+		return nil
+	}
+	block := q.blocks[0]
+	q.blocks[0] = nil
+	q.blocks = q.blocks[1:]
+	return block
+}
+
+// remove removes the block from the queue and any others that follow it. If the
+// block doesn't exit within the queue, then this acts as a NOP.
+func (q *blockRetryQueue) remove(header wire.BlockHeader) {
+	headerIdx := -1
+	targetHash := header.BlockHash()
+	for i, block := range q.blocks {
+		blockHeader := block.Header()
+		if blockHeader.BlockHash() == targetHash {
+			headerIdx = i
+			break
+		}
+	}
+	if headerIdx != -1 {
+		for i := headerIdx; i < len(q.blocks); i++ {
+			q.blocks[i] = nil
+		}
+		q.blocks = q.blocks[:headerIdx]
+	}
+}
+
+// clear clears the queue.
+func (q *blockRetryQueue) clear() {
+	q.blocks = nil
 }
 
 // updateChan specifies an update channel. This is for internal use by the
@@ -420,65 +484,15 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		}
 	}()
 
-	// blockRetryInterval is the interval in which we'll continually re-try
-	// to fetch the latest filter from our peers.
-	//
-	// TODO(roasbeef): add exponential back-off
-	blockRetryInterval := time.Millisecond * 100
-
-	// blockReFetchTimer is a stoppable timer that we'll use to reminder
-	// ourselves to refetch a block in the case that we're unable to fetch
-	// the filter for a block the first time around.
 	var (
-		blockReFetchTimer *time.Timer
-		reFetchMtx        sync.Mutex
+		blockRetrySignal <-chan time.Time
+		// blockRetryInterval is the interval in which we'll continually
+		// retry to fetch the latest filter from our peers.
+		//
+		// TODO(roasbeef): add exponential back-off
+		blockRetryInterval = time.Millisecond * 100
+		blockRetryQueue    = newBlockRetryQueue()
 	)
-
-	resetBlockReFetchTimer := func(headerTip wire.BlockHeader, height uint32) {
-		// If so, then we'll avoid notifying the block, and will
-		// instead add this to our retry queue, as we should be getting
-		// block disconnected notifications in short order.
-		if blockReFetchTimer != nil {
-			blockReFetchTimer.Stop()
-		}
-
-		log.Infof("Setting timer to attempt to re-fetch filter for "+
-			"hash=%v, height=%v", headerTip.BlockHash(), height)
-
-		blockReFetch := func() {
-			// If we're unable to process notifications at the
-			// moment (due to not being current), we'll reset our
-			// timer.
-			reFetchMtx.Lock()
-			if blockReFetchTimer != nil && blockSubscription == nil {
-				if !blockReFetchTimer.Stop() {
-					<-blockReFetchTimer.C
-				}
-				blockReFetchTimer.Reset(blockRetryInterval)
-
-				reFetchMtx.Unlock()
-				return
-			}
-			reFetchMtx.Unlock()
-
-			log.Infof("Resending rescan header for block hash=%v, "+
-				"height=%v", headerTip.BlockHash(), height)
-
-			ntfn := blockntfns.NewBlockConnected(headerTip, height)
-			select {
-			case blockSubscription.Notifications <- ntfn:
-			case <-ro.quit:
-			}
-		}
-
-		// We'll start a timer to re-send this header so we re-process
-		// if in the case that we don't get a re-org soon afterwards.
-		reFetchMtx.Lock()
-		blockReFetchTimer = time.AfterFunc(
-			blockRetryInterval, blockReFetch,
-		)
-		reFetchMtx.Unlock()
-	}
 
 	// handleBlockConnected is a closure that handles a new block connected
 	// notification.
@@ -564,10 +578,7 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// have any peers to fetch this filter from, or the peer(s) that
 		// we're trying to fetch from are in the progress of a re-org.
 		if blockFilter == nil {
-			// TODO(halseth): this is racy, as blocks can come in
-			// before we refetch.
-			resetBlockReFetchTimer(header, uint32(curStamp.Height))
-			return nil
+			return errRetryBlock
 		}
 
 		err = notifyBlockWithFilter(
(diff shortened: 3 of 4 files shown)


neutrino.go
```diff
@@ -999,7 +999,7 @@ func (s *ChainService) rollBackToHeight(height uint32) (*waddrmgr.BlockStamp, er
 	}
 
 	for uint32(bs.Height) > height {
-		header, _, err := s.BlockHeaders.FetchHeader(&bs.Hash)
+		header, headerHeight, err := s.BlockHeaders.FetchHeader(&bs.Hash)
```
Roasbeef: 🤔

Roasbeef: So block disconnects haven't been delivering the proper information this entire time? Wouldn't this also then propagate to `lnd` where we exercise these scenarios in the chain notifier?

rescan.go
```diff
@@ -482,24 +444,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 	//
 	// TODO(wilmer): refactor this and handleBlockDisconnected into their
 	// own methods.
-	handleBlockConnected := func(ntfn *blockntfns.Connected) error {
+	handleBlockConnected := func(ntfn *blockntfns.Connected, isRetry bool) error {
```
Roasbeef: Perhaps it's time to extract this closure (and the disconnect one) to a distinct method?

rescan.go
```diff
@@ -482,24 +444,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+		// If we have any blocks to retry and the notification doesn't
+		// correspond to the current block being retried, we'll queue it
+		// up to retry after.
+		newHeader := ntfn.Header()
+		if len(retryBlocks) > 0 {
+			retryHeader := retryBlocks[0].Header()
+			if retryHeader.BlockHash() != newHeader.BlockHash() {
+				log.Debug("Stashing", ntfn)
```
Roasbeef: Reminder to remove, or improve this log message before merge.

rescan.go
```diff
@@ -482,24 +444,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+				retryBlocks = append(retryBlocks, ntfn)
+				return nil
+			}
+		}
+
```
Roasbeef: Shouldn't we also return if we're getting a notification to try a block that we're already attempting to retry?

rescan.go
```diff
@@ -508,55 +474,54 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// notification, then we can re-process it without any issues.
 		nextBlockHeight := uint32(curStamp.Height + 1)
 		_, err := chain.GetFilterHeaderByHeight(nextBlockHeight)
-		if header.BlockHash() != curStamp.Hash && err != nil {
+		if !isRetry && err != nil {
```
Roasbeef: Why is it acceptable to not be able to fetch a filter header if this is  a block that we're retrying? Shouldn't we immediately fall back to fetching this filter again?

rescan.go
```diff
@@ -573,25 +538,50 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
-			blockReFetchTimer.Stop()
-			blockReFetchTimer = nil
+		// With the notifications sent, we can update our in-memory
+		// state.
+		curHeader = newHeader
+		curStamp = newStamp
+
+		// If this was a retry, we'll proceed to the next if any.
+		if isRetry {
+			retryBlocks = retryBlocks[1:]
```
Roasbeef: To prevent a GC leak, `retryBlocks[0]` should be set to `nil`.

Roasbeef: How do we handle the following case:
  * We're trying to fetch a new block, however a re-org occurs at that point 
  * As a result, if we try to fetch that blocks, we'll _never_ get a response since it's no longer in the main-chain (`btcd`'s behavior) 

In this case, it would seem that we'll need to _replace_ newer items for older items within the retry queue if we get two at the same height.

rescan.go
```diff
@@ -600,12 +590,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+		retryIdx := -1
+		for i, retryBlock := range retryBlocks {
+			retryHeader := retryBlock.Header()
+			if retryHeader.BlockHash() == blockDisconnected.BlockHash() {
+				retryIdx = i
+				break
+			}
+		}
+		if retryIdx != -1 {
+			retryBlocks = retryBlocks[:retryIdx]
```
Roasbeef: Similar comment about `nil` values here to prevent a GC pointer leak.

rescan.go
```diff
@@ -773,12 +788,12 @@ rescanLoop:
...
-					if blockSubscription != nil {
-						blockSubscription.Cancel()
-						blockSubscription = nil
-					}
-				}()
+
+				// Clear any retry blocks as we are now
+				// considered current with the chain.
+				if len(retryBlocks) > 0 {
+					retryBlocks = retryBlocks[:0]
```
Roasbeef: `nil`-ness

Roasbeef: If we get to this point: we're synced, but still have retry blocks, has an assumption in the current code been violated?

rescan_test.go
```diff
@@ -0,0 +1,621 @@
...
+				// successful.
+				block1 := ctx.chain.addNewBlock(true)
+				assertFilterQueried(ctx.t, block1.Hash)
+				ctx.chain.addNewBlock(true)
+				assertFilterQueried(ctx.t, block1.Hash)
+
+				// We'll then reorg back to a height where the
+				// new blocks we've added and retried are stale,
+				// causing us to no longer retry them.
+				ctx.chain.rollbackToHeight(bestBlock.Height, true)
```
Roasbeef: Shouldn't we verify that the next block added (for a new chain) causes us to continue fetching down that new chain?

wpaulino: Good point, done.

rescan.go
wpaulino: If we have N retry blocks and N-1 causes us to no longer be current, then we'll need to clear these once we become current again.

Other than when an explicit rescan rewind is done, the only way for us to no longer be current is for our notification source to dispatch a notification out of order. This ideally shouldn't happen unless there's a bug in that logic.

wpaulino: Done.

rescan.go
wpaulino: Done.

rescan.go
wpaulino: We could definitely do that, though the assumption is that we'll receive block disconnected notifications from the `blockManager` for those blocks, which we'll then process within `handleBlockDisconnected` and remove them. Given that, do you still think it's worth handling this case explicitly here?

wpaulino: Done.

rescan.go
wpaulino: I'm not sure what this check achieves in the first place. I noticed it was reverted as part of https://github.com/lightninglabs/neutrino/pull/104, but I don't know if that still applies cc @halseth.

The change as part of this diff is just maintaining the same check, i.e. `(header.BlockHash() != curStamp.Hash) == !isRetry`.

rescan.go
wpaulino: The assumption is that wouldn't happen given the correctness of the notification source, i.e. duplicate notifications wouldn't be dispatched for the same block. If the notification comes through with `isRetry=false` and it's already being retried, we'd mark ourselves as not current and properly catch up.

rescan.go
wpaulino: Log in its full form is actually descriptive if you look at the notification's string method.

rescan.go
wpaulino: Both closures currently depend on a lot of other internal state within the `rescan` method, so I'd rather defer doing so until we refactor the `Rescan` struct itself when moving towards a `rescan` package.

neutrino.go
wpaulino: Yeah, this could be the cause to some of the neutrino reorg tests within lnd failing as they depend on heights, though I haven't narrowed anything down.

Roasbeef: IIRC, we ended up disabling one or two of them, so we can note to enable them again once this goes in and `lnd` itself is updated.

rescan.go
Roasbeef: > We could definitely do that, though the assumption is that we'll receive block disconnected notifications from the blockManager for those blocks, which we'll then process within handleBlockDisconnected and remove them

After reading the rest of this diff, it seems this case is addressed. However, the scenario I described above can still happen if _right_ as we go to fetch that block, a re-org happens.

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+		select {
+		case hashQueried = <-filterQueried:
+			if !hashQueried.IsEqual(&hash) {
+				t.Fatalf("expected to query filter %v, got %v",
+					hash, hashQueried)
+			}
+		case <-time.After(time.Second):
+			t.Fatal("expected rescan to query for filter")
+		}
+	}
```
joostjager: Shouldn't this be part of the `mockChainSource`? Make a bool field that controls `filterNotFound` behaviour. Then also all filter queries can be pushed into a channel and asserted on, not just the not found ones.

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+			ctx.t.Fatalf("expected ErrRescanExit upon shutdown, "+
+				"got %v", err)
+		}
+
+	case <-time.After(time.Second):
+		// Print a goroutine dump before exiting.
+		pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
+		ctx.t.Fatal("rescan did not exit cleanly")
+	}
+}
```
joostjager: Check here that no unexpected events have been posted into the channels?

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+				// We'll then reorg back to the first retry
+				// block. This should cause us to stop retrying
+				// the reorged blocks.
+				ctx.chain.rollbackToHeight(block.Height, true)
+
+				// We should then receive a notification for it
+				// once we revert back to the original
+				// GetCFilter implementation.
+				ctx.chain.overrideGetCFilter(nil)
+				ctx.recvBlockConnected(block)
```
joostjager: Doesn't this need more assertions to check that the right queries are executed now? Or is that all implied by the block connected ntfn?

rescan.go
joostjager: There is already the struct `Rescan` that could hold that internal state. `Rescan` is a single use object anyway. :+1: on extracting the closures. Preferably in a commit before the actual fix is implemented.

rescan.go
```diff
@@ -482,24 +444,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
-		//
-		// It's possible due to the nature of the current subscription
-		// system that we get a duplicate block. We'll catch this and
-		// continue forwards to avoid an unnecessary state transition
-		// back to the !current state.
-		header := ntfn.Header()
-		if header.PrevBlock != curStamp.Hash &&
-			header.BlockHash() != curStamp.Hash {
-
+		if newHeader.PrevBlock != curStamp.Hash {
```
joostjager: It would have been nice to have refactoring that doesn't change behavior in a separate commit, to get a better overview on the fix itself.

rescan.go
```diff
@@ -600,12 +591,31 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+		// we've queued up to retry. If it is, we'll remove it and any
+		// others that follow as they are now considered stale.
+		retryIdx := -1
+		for i, retryBlock := range retryBlocks {
+			retryHeader := retryBlock.Header()
+			if retryHeader.BlockHash() == blockDisconnected.BlockHash() {
+				retryIdx = i
+				break
+			}
+		}
```
joostjager: Looking at the amount of code that manipulates `retryBlocks`, it may be justified to create a separate struct for it with methods to manipulate the list.

rescan.go
```diff
@@ -717,13 +722,27 @@ rescanLoop:
...
+			// Our retry signal has fired, so we'll attempt to
+			// refetch the filter for our next queued block.
+			case <-blockRetrySignal:
+				blockRetrySignal = nil
+				if len(retryBlocks) == 0 {
+					continue rescanLoop
+				}
+
+				ntfn := retryBlocks[0]
+				err := handleBlockConnected(ntfn, true)
```
joostjager: It feels a bit odd that we already know that it is a retry, but `handleBlockConnected` is still going to check `retryBlocks`. Maybe `handleBlockConnected` should be divided in two functions such that it doesn't go through this check if it is triggered by `blockRetrySignal`. Decompose it such that it is clear what the common logic is between the retry and non-retry case.

rescan.go
```diff
@@ -773,12 +792,10 @@ rescanLoop:
...
-				defer func() {
-					if blockSubscription != nil {
-						blockSubscription.Cancel()
-						blockSubscription = nil
-					}
-				}()
+
+				// Clear any retry blocks as we are now
+				// considered current with the chain.
+				retryBlocks = nil
```
joostjager: Can there still be retryBlocks at this point?

rescan.go
```diff
@@ -573,25 +538,51 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+			retryBlocks = retryBlocks[1:]
+			if len(retryBlocks) > 0 {
+				next := retryBlocks[0]
+				nextHeader := next.Header()
+
+				log.Debugf("Retrying filter fetch for height=%v, "+
+					"hash=%v after %v", next.Height(),
+					nextHeader.BlockHash(), blockRetryInterval)
+
+				blockRetrySignal = time.After(blockRetryInterval)
```
joostjager: If we are successful, is a new retry interval needed or can it continue directly with the rest of the list?

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/gcs"
+	"github.com/btcsuite/btcutil/gcs/builder"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/neutrino/blockntfns"
+)
+
+// mockChainSource is a mock implementation of the ChainSource interface that
```
halseth: This commit is quite large.  Possible to add the test/harness in a separate commit, then follow up with a commit that fixes the actual error?

rescan.go
```diff
@@ -482,24 +444,28 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 	//
 	// TODO(wilmer): refactor this and handleBlockDisconnected into their
 	// own methods.
-	handleBlockConnected := func(ntfn *blockntfns.Connected) error {
+	handleBlockConnected := func(ntfn *blockntfns.Connected, isRetry bool) error {
+		// If we have any blocks to retry and the notification doesn't
+		// correspond to the current block being retried, we'll queue it
+		// up to retry after.
+		newHeader := ntfn.Header()
```
halseth: Instead of doing it this way, what about sending every connected blocks into a new go routine, that would handle them? That way we could continue consuming `block connected` notifications while this go routine is fetching filters and handling retries.

halseth: Where would we access the internal state of the rescan? Wouldn't the caller always be looking at the filtered block?

wpaulino: Yeah and the filtered block is only updated once the filter is fetched. If I'm understanding correctly, you're suggesting a goroutine that is sent blocks to fetch filters for. If we continue consuming notifications while the goroutine is doing its thing, the rescan could end up tracking blocks ahead of the goroutine if we're unable to fetch filters. What benefit does the goroutine provide that is not already achieved?

roeierez:
@wpaulino this PR addresses https://github.com/lightninglabs/neutrino/issues/166 right?
Thanks for the great work and effort!

rescan.go
halseth: I imagine it would be much easier to reason about and unit test, as the consumer of this ordered queue of blocks can be tested in isolation. It will also pave the way for doing batch filter fetching as this queue grows.

Roasbeef: Why does batch filter fetching need to be handled there? As is, the query API handles that given hints from the caller. 

Concurrency wise, it's much easier to have all this in a single execution context as you don't need to reason about serialization beyond receiving the new block connected/disconnected notifications.

rescan.go
```diff
@@ -573,25 +538,51 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 		// If the filter is nil, then this either means that we don't
 		// have any peers to fetch this filter from, or the peer(s) that
 		// we're trying to fetch from are in the progress of a re-org.
+		// In any case, we'll attempt to retry fetching the filter until
+		// success or we determine it belongs to a stale block.
```
halseth: To simplify, we would just loop here until we have the filter.

rescan.go
halseth: We don't necessarily need to have it being handled in a different goroutine, but I mean we wouldn't proceed until we have been able to fetch the filter. (I added a comment on the line in `handleBlockConnected`)

My rationale for doing this in a new goroutine is to not stall consumption of block connected notifications, but not sure if that matters.

rescan.go
wpaulino: I don't think we can loop here cause we need a signal to stop fetching the filter if the corresponding block gets reorged.

rescan.go
wpaulino: > My rationale for doing this in a new goroutine is to not stall consumption of block connected notifications, but not sure if that matters.

No need to due to the subscription being able to buffer notifications.

rescan.go
halseth: wouldn't we get a block disconnected notification in that case?

wpaulino: Yeah but if we loop here we won't process any further notifications. Even if we do, it's possible we receive a connected notification instead, which we'd need to stash anyway.

halseth: I'm okay with keeping the current design, and we can circle back to this later if needed.

query.go
```diff
@@ -1224,8 +1224,9 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 		// If there are elements left to receive, the query failed.
 		if len(query.headerIndex) > 0 {
 			numFilters := query.stopHeight - query.startHeight + 1
+			numRecv := numFilters - int64(len(query.headerIndex))
 			log.Errorf("Query failed with %d out of %d filters "+
-				"received", len(query.headerIndex), numFilters)
+				"received", numRecv, numFilters)
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+	filterHeadersByHeight map[uint32]*chainhash.Hash
+	getFilterOverride     func(chainhash.Hash, wire.FilterType,
+		...QueryOption) (*gcs.Filter, error)
+	subscribeOverride func(uint32) (*blockntfns.Subscription, error)
+}
+
+var _ ChainSource = (*mockChainSource)(nil)
+
+// newMockChainSource creates a new mock chain backed by numBlocks.
+func newMockChainSource(numBlocks int) *mockChainSource {
```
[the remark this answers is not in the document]
wpaulino: Would like to keep everything in the rescan related file until we move towards a package.

rescan_test.go
```diff
@@ -0,0 +1,680 @@
...
+
+	testCases := []struct {
+		name string
+		run  func(ctx *rescanTestContext)
+	}{
+		{
+			// We should not attempt to process new blocks if we're
+			// retrying a previous one.
+			name: "new block after retry",
+			run: func(ctx *rescanTestContext) {
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan_test.go
wpaulino: Done.

rescan_test.go
wpaulino: Done.

rescan_test.go
wpaulino: We rely on the notification of the block being delivered if its filter is fetched. Added the assertion anyway though since we now can.

rescan.go
wpaulino: Done.

rescan.go
wpaulino: Done!

rescan.go
wpaulino: Fixed.

rescan.go
wpaulino: Yes, if we fail to process a notification in `handleBlockConnected` after queueing blocks.

rescan.go
wpaulino: We could proceed without another signal, but the reasoning behind it is to allow the possibility of a disconnected notification to be sent to the rescan so that it can stop retrying.

rescan_test.go
wpaulino: Done.

rescan_test.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	select {
+	case <-ctx.blocksConnected:
+		ctx.t.Fatalf("received unexpected block connected after shutdown")
+	case <-ctx.blocksDisconnected:
+		ctx.t.Fatalf("received unexpected block disconnected after shutdown")
+	case <-ctx.chain.filtersQueried:
+		ctx.t.Fatalf("received unexpected filter query after shutdown")
+	default:
```
joostjager: `default` can be removed

rescan.go
```diff
@@ -200,6 +204,65 @@ func QuitChan(quit <-chan struct{}) RescanOption {
...
+			q.blocks[i] = nil
+		}
+		q.blocks = q.blocks[:headerIdx]
+	}
+}
+
+// clear clears the queue.
+func (q *blockRetryQueue) clear() {
+	q.blocks = nil
+}
```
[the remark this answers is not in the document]
joostjager: I guess you answered that question already. You wanted to wait for a new package first.

rescan_test.go
```diff
@@ -504,3 +504,188 @@ func (ctx *rescanTestContext) waitUntilSynced() {
...
+	// Adding multiple new blocks to the chain should result in the first
+	// retry block being queried until successful.
+	block1 := ctx.chain.addNewBlock(true)
+	ctx.assertFilterQueried(block1.Hash)
+	block2 := ctx.chain.addNewBlock(true)
+	ctx.assertFilterQueried(block1.Hash)
+
+	// We'll then reorg back to a height where the new blocks we've added
+	// and retried are stale, causing us to no longer retry them. This won't
+	// deliver any disconnected notifications since the blocks were never
```
halseth: I think the comment should be slightly changed to point out that they were never handled from the POV of the rescan. Disconnected notifications are delivered from the chain source I reckon?

rescan_test.go
```diff
@@ -504,3 +504,188 @@ func (ctx *rescanTestContext) waitUntilSynced() {
...
+	// manually.
+	block2 := ctx.chain.addNewBlock(false)
+	block3 := ctx.chain.addNewBlock(false)
+	ctx.chain.ntfnChan <- blockntfns.NewBlockConnected(wire.BlockHeader{}, 0)
+
+	// Revert the mocked chain so that block filters can be retrieved
+	// successfully.
+	ctx.chain.setFailGetFilter(false)
+
+	// The notifications for all the blocks added above should now be
```
halseth: should we assert that the rescan loop is also fetching the blocks from the chain source here?

rescan_test.go
joostjager: True

rescan.go
```diff
@@ -722,6 +745,58 @@ rescanLoop:
...
+				default:
+					log.Errorf("Unable to process retry "+
+						"of %v: %v", retryBlock, err)
+
+					// Since we weren't able to successfully
+					// process the block, we'll set
+					// ourselves to not be current in order
+					// to attempt catching up with the chain
+					// ourselves.
+					current = false
```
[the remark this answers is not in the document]
wpaulino: Yes,

rescan_test.go
wpaulino: Doesn't seem like a useful assertion to make at the moment?

rescan_test.go
wpaulino: Done.

rescan.go
```diff
@@ -722,6 +745,58 @@ rescanLoop:
...
+					// to allow for a disconnected
+					// notification to potentially arrive
+					// for it.
+					_ = blockRetryQueue.pop()
+					nextRetryBlock := blockRetryQueue.peek()
+					if nextRetryBlock == nil {
+						continue rescanLoop
+					}
+
+					retryBlock = nextRetryBlock
```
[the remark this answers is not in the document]
wpaulino: See https://github.com/lightninglabs/neutrino/pull/175#discussion_r314287428.

rescan.go
```diff
@@ -722,6 +745,58 @@ rescanLoop:
...
+				// disconnected right before.
+				retryBlock := blockRetryQueue.peek()
+				if retryBlock == nil {
+					continue rescanLoop
+				}
+
+				err := handleBlockConnected(retryBlock)
+				switch err {
+				case nil:
+					// We successfully notified the block
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan.go
```diff
@@ -697,8 +692,29 @@ rescanLoop:
...
+						continue rescanLoop
+					}
+
 					err := handleBlockConnected(ntfn)
-					if err != nil {
+					switch err {
+					case nil:
+						break
+
+					case errRetryBlock:
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan.go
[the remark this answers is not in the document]
wpaulino: Similar to the log level comment, yes as an assumption has been violated and we should attempt to catch up with the chain ourselves.

rescan.go
```diff
@@ -722,6 +745,58 @@ rescanLoop:
...
+
+				case errRetryBlock:
+					log.Debugf("Retrying %v after %v",
+						retryBlock, blockRetryInterval)
+					blockRetrySignal = time.After(
+						blockRetryInterval,
+					)
+
+				default:
+					log.Errorf("Unable to process retry "+
```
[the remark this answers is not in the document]
wpaulino: Yes, all the errors returned are for an assumption being violated.

rescan.go
```diff
@@ -722,6 +745,58 @@ rescanLoop:
...
+				err := handleBlockConnected(retryBlock)
+				switch err {
+				case nil:
+					// We successfully notified the block
+					// this time, so we can remove it from
+					// our queue. If we have another block
+					// to retry, we'll set up another signal
+					// to allow for a disconnected
+					// notification to potentially arrive
+					// for it.
```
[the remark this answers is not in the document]
wpaulino: > Is it likely that that disconnected notification arrives at this point?

"Likely" depends on the chain. This is more about _allowing_ a disconnected notification to be received so that we properly stop retrying a stale block.

> When we are rescanning and there are no retries, there also isn't such a delay?

Rescanning where?

rescan.go
```diff
@@ -200,6 +204,65 @@ func QuitChan(quit <-chan struct{}) RescanOption {
...
+	return block
+}
+
+// remove removes the block from the queue and any others that follow it. If the
+// block doesn't exit within the queue, then this acts as a NOP.
+func (q *blockRetryQueue) remove(header wire.BlockHeader) {
+	headerIdx := -1
+	for i, block := range q.blocks {
+		blockHeader := block.Header()
+		if blockHeader.BlockHash() == header.BlockHash() {
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan.go
```diff
@@ -200,6 +204,65 @@ func QuitChan(quit <-chan struct{}) RescanOption {
...
+}
+
+// pop returns and consumes the next block to be retried.
+func (q *blockRetryQueue) pop() *blockntfns.Connected {
+	if len(q.blocks) == 0 {
+		return nil
+	}
+	block := q.blocks[0]
+	q.blocks[0] = nil
+	q.blocks = q.blocks[1:]
```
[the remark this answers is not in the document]
wpaulino: No, see https://play.golang.org/p/K00tLHvBQJG.

rescan.go
```diff
@@ -409,6 +409,12 @@ func rescan(chain ChainSource, options ...RescanOption) error {
 	scanning := ro.startTime.Before(curHeader.Timestamp)
 
 	var blockSubscription *blockntfns.Subscription
+	defer func() {
+		if blockSubscription != nil {
+			blockSubscription.Cancel()
+			blockSubscription = nil
+		}
+	}()
```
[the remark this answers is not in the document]
wpaulino: Done.

rescan.go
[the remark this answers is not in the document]
joostjager: Does that mean that `[1:]` copies the whole slice one position back in the array?

rescan.go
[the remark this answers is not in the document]
joostjager: Aren't these things that can happen when connected to a buggy neutrino node? Like an out of order block?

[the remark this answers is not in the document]
wpaulino: Out of order blocks shouldn't happen. The rescan should receive notifications properly (disconnected for reorged blocks, connected for new ones, etc).

rescan.go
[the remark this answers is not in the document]
joostjager: What I mean is that if there is a reorg and we need to fetch a series of new blocks and those blocks are fetched correctly without retries, we also don't delay in between.

rescan_test.go
```diff
@@ -0,0 +1,691 @@
...
+		}
+	case <-time.After(time.Second):
+		ctx.t.Fatalf("expected to receive block connected "+
+			"notification for %v", spew.Sdump(block))
+	}
+}
+
+// recvBlockDisconnected is a helper method used to expect a block disconnected
+// notification being delivered.
+func (ctx *rescanTestContext) recvBlockDisconnected(height int32) {
```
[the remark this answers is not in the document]
wpaulino: Removed.

rescan_test.go
halseth: We assert that filters are queried, so could also assert that blocks are, but no blocker.

PR #178, opened 2019-08-21
neutrino: misc connection management fixes

wpaulino:
In this PR, we include a series of connection management fixes that were discovered through testing chain syncs with no `connect` or `addpeers`. Each fix has been included in its own commit along with a descriptive message of the issue it's addressing.

Mainly, this PR attempts to ensure we can more aggressively connect to peers through the underlying connection manager.


blockmanager.go
```diff
@@ -1273,9 +1273,13 @@ func (b *blockManager) resolveConflict(
 					log.Errorf("Unable to ban peer %v: %v",
 						peer, err)
 				}
+				if sp := b.server.PeerByAddr(peer); sp != nil {
```
Roasbeef: Could've sworn that `BanPeer` _did_ disconnect the peer, but this isn't the case as is...

neutrino.go
```diff
@@ -1360,15 +1356,16 @@ func (s *ChainService) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn)
...
-		conn.Close()
+		s.connManager.Remove(c.ID())
 		return
 	}
 
 	sp := newServerPeer(s, c.Permanent)
 	p, err := peer.NewOutboundPeer(newPeerConfig(sp), peerAddr)
 	if err != nil {
 		log.Debugf("Cannot create outbound peer %s: %s", c.Addr, err)
-		s.connManager.Disconnect(c.ID())
```
Roasbeef: dat connection leak...

neutrino.go
```diff
@@ -1360,15 +1356,16 @@ func (s *ChainService) outboundPeerConnected(c *connmgr.ConnReq, conn net.Conn)
 	// If we're already connected to this peer, then we'll close out the new
 	// connection and keep the old.
 	if s.PeerByAddr(peerAddr) != nil {
-		conn.Close()
+		s.connManager.Remove(c.ID())
```
Roasbeef: We may want to eventually change the names of `Remove/Disconnect` in the `connmgr` to better reflect their retry semantics.

blockmanager.go
halseth: > Could modify the method to do so as well

+1

wpaulino: Done.

log.go
```diff
@@ -37,6 +38,7 @@ func UseLogger(logger btclog.Logger) {
 	addrmgr.UseLogger(logger)
 	blockntfns.UseLogger(logger)
 	pushtx.UseLogger(logger)
+	connmgr.UseLogger(logger)
```
Roasbeef: FWIW, when it's run within `lnd`, we've already hooked up the `connmgr` as a logger.

blockmanager.go
```diff
@@ -1363,10 +1350,6 @@ func (b *blockManager) resolveConflict(
 					log.Errorf("Unable to ban peer %v: %v",
 						peer, err)
 				}
-				sp := b.server.PeerByAddr(peer)
-				if sp != nil {
```
Roasbeef: 👍 

much less clutter this way by pushing the logic into `BanPeer`.

PR #179, opened 2019-08-27
[Query] Global peer work dispatcher

halseth:
This PR starts a significant rewrite of the logic used for requesting data from peers (_query logic_). The calls `queryPeers` and `queryBatch` are replaced with a general call `Query`, which takes a batch of queries, and distributes them across the connected peers. 

Later iterations will also replace the `queryAllPeers` call.

### Worker
Each peer is associated with a `worker`, which will be polling queries on its `nextJob` channel, and give them to the peer. (Currently it will poll only one job at a time, but the plan is to increase this later to speed queries). It keeps a message subscription to the peer, and signals back to the `WorkManager` when the polled query has gotten a valid response. If a response hasn't been received within the query's timeout, it will report back with a timeout error.

### WorkManager
The `WorkManager` is the main entry point for callers wanting to send queries to connected peers. It keeps a `ConnectedPeers` subscription, and will spin up and remove workers as peers connect and disconnects.

It satisfies the `QueryAccess` interface: 
```go
	// Query distributes the slice of queries to the set of connected
	// peers. It returns an error channel where the final result of the
	// batch of queries will be sent. Responses for the individual queries
	// should be handled by the response handler of each Query.
	Query(queries []*Query, options ...QueryOption) chan error
```

When a slice of queries is given to the `WorkManager` it will distribute them across the workers which is not currently working on something else. When results come back from the workers, it acts accordingly; rescheduling failed queries, and removes successful queries from its queue.

### PeerRanking
To determine which peers to prioritise for queries, a simple `PeerRanking` widget is added. It keeps track of connected peers, decreases their score when they fail a query, and increases it when they succeed. When scheduling a new query, the `WorkManager` will attempt to give it to the worker with the peer with the highest ranking, as long as it isn't working on something else.

### Dynamic query timeouts
To quickly converge on the best set of peers, an individual query's timeout is made dynamic. Earlier we were always using a timeout of 10 seconds. Now it starts out at 2 seconds, and if no response is received within this timeout, we double it. The peers that is able to answer within the timeout are rewarded, and should be more likely to pick up the next query. It also has the nice benefit that it stabilizes the `sync_tests`, as we quickly timeout on the bad peers.


notifications.go
```diff
@@ -8,15 +8,26 @@ package neutrino
...
 type getConnCountMsg struct {
 	reply chan int32
 }
 
+type subConnPeersReply struct {
+	peerChan   chan query.Peer
+	cancelChan chan struct{}
+}
+type subConnPeersMsg struct {
+	cancel <-chan struct{}
```
wpaulino: Looks like this field isn't needed?

blockmanager.go
```diff
@@ -1019,109 +1022,153 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
-				return true
-			}
+	// Keep waiting for more headers as long as we haven't received an
+	// answer for our last checkpoint, and no error is encountered.
+	for {
+		var r *wire.MsgCFHeaders
+		select {
+		case r = <-headerChan:
+		case err := <-errChan:
+			log.Errorf("Query finished with error (%v) before "+
```
wpaulino: Nit: add error at the end to make it easier to parse?

blockmanager.go
```diff
@@ -1019,109 +1022,153 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
+		}
 
-			// If this is out of order stuff that's already been
-			// written, we can ignore it.
-			if lastHeight <= curHeight {
-				log.Debugf("Received out of order reply "+
-					"end_height=%v, already written", lastHeight)
-				return true
-			}
+		checkPointIndex, ok := stopHashes[r.StopHash]
```
wpaulino: Nit: no need to check `ok` here as we already did before this.

blockmanager.go
```diff
@@ -1019,109 +1022,153 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
-			// Then, we cycle through any cached messages, adding
-			// them to the batch and deleting them from the cache.
-			for {
-				// Determine the next checkpoint index we should
-				// process.
-				checkPointIndex += maxCFCheckptsPerQuery
-				if checkPointIndex == uint32(len(checkpoints)) {
-					checkPointIndex = uint32(len(checkpoints)) - 1
-				}
+		log.Debugf("Got cfheaders from height=%v to "+
```
wpaulino: Nit: could extend some of these lines to use the new space available.

neutrino.go
```diff
@@ -669,6 +669,7 @@ type ChainService struct {
 	broadcaster          *pushtx.Broadcaster
 	banStore             banman.Store
 	workManager          *query.WorkManager
+	queryAccess          query.QueryAccess
```
wpaulino: Why is this field needed if it does the same as `WorkManager`?

query.go
```diff
@@ -828,57 +824,45 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
 
-		s.queryPeers(
-			// Send a wire.MsgGetCFilters
-			query.queryMsg(),
-
-			// Check responses and if we get one that matches, end
-			// the query early.
-			func(_ *ServerPeer, resp wire.Message, quit chan<- struct{}) {
-				s.handleCFiltersResponse(query, resp, quit)
+	msgs := []*query.Query{
```
wpaulino: Construct this as part of `prepareCFiltersQuery`?

query.go
```diff
@@ -1008,25 +996,44 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
...
+	for {
+		select {
+		case block = <-foundBlock:
+			break Loop
+		case err := <-errChan:
+			if err != nil {
+				return nil, err
+			}
+		case <-s.quit:
+			return nil, fmt.Errorf("ChainService shutting down")
```
wpaulino: Nit: use `ErrShuttingDown`.

query.go
```diff
@@ -1008,25 +996,44 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
...
 	)
-	if foundBlock == nil {
-		return nil, fmt.Errorf("Couldn't retrieve block %s from "+
-			"network", blockHash)
+
+	var block *btcutil.Block
+
+Loop:
+	// Wait for the block, or an error to occur.
+	for {
```
wpaulino: Why do we need this loop? The work manager can only deliver `nil` or `ErrWorkManagerShuttingDown`.

query/worker.go
```diff
@@ -89,10 +93,17 @@ func (w *worker) run() {
 			return
 		}
 
-		// We received a query job, send it to the peer.
-		w.peer.QueueMessageWithEncoding(
-			job.Req, nil, job.options.encoding,
-		)
+		select {
+		// There is no point in queieing the request if the job already
```
wpaulino: Nit: queueing*

query/workmanager.go
```diff
@@ -211,37 +211,61 @@ func (w *WorkManager) workDispatcher(nextJob chan<- *queryTask,
...
-				b := currentBatches[batchNum]
-				b.rem--
-
-				// If this was the last query in flight for
-				// this batch, we can notify that it finished,
-				// and delete it.
-				if b.rem == 0 {
-					b.errChan <- nil
-					delete(currentBatches, batchNum)
+				b, ok := currentBatches[batchNum]
```
wpaulino: Do we need to check `ok` here?

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+
+	Loop:
+		for {
+
+			select {
+			// A message was received from the peer, use the
+			// response handler to check whether it was answering
+			// our request.
+			case resp = <-msgChan:
+				done, progress := job.HandleResp(
```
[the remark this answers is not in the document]
wpaulino: After going through the whole diff, I noticed you refactored `HandleResp` to address my comment. Perhaps we can just document that `HandleResp` should not contain any expensive operations?

blockmanager.go
wpaulino: Also needs `if err != nil`.

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+					continue Loop
+				}
+
+				// We did get a valid response, and can break
+				// the loop.
+				break Loop
+
+			// If the timeout is reached before a valid response
+			// has been received, we exit with an error.
+			case <-timeout:
```
wpaulino: Do we want to prevent reassigning the failed query to the same peer here or we do just rely on it eventually being assigned to a different peer?

blockmanager.go
```diff
@@ -98,12 +99,39 @@ type txMsg struct {
...
 }
 
+// blockManagerCfg holds options and dependencies needed by the blockManager
+// during operation.
+type blockManagerCfg struct {
+	// ChainParams is the chain that we're running on.
+	ChainParams chaincfg.Params
+
+	BlockHeaders     headerfs.BlockHeaderStore
+	RegFilterHeaders *headerfs.FilterHeaderStore
```
Roasbeef: Why is the block header store a regular value, while this is a pointer?

blockmanager.go
```diff
@@ -98,12 +99,39 @@ type txMsg struct {
...
+	TimeSource       blockchain.MedianTimeSource
+
+	BanPeer func(addr string, reason banman.Reason) error
+
+	// GetBlock fetches a block from the p2p network.
+	GetBlock func(chainhash.Hash, ...QueryOption) (*btcutil.Block, error)
+
+	queryBatch func([]wire.Message, func(*ServerPeer, wire.Message,
```
Roasbeef: IMO either all the values should be private, or all public.

query/workqueue.go
```diff
@@ -0,0 +1,56 @@
+package query
+
+// Task is an interface that have a method for returning their index in the
```
Roasbeef: have -> has

query/workqueue.go
```diff
@@ -0,0 +1,56 @@
...
+func (w *workQueue) Push(x interface{}) {
+	w.tasks = append(w.tasks, x.(Task))
+}
+
+// Pop removes and returns element Len()-1.
+//
+// NOTE: This is part of the heap.Interface implementation.
+func (w *workQueue) Pop() interface{} {
+	n := len(w.tasks)
+	x := w.tasks[n-1]
```
Roasbeef: As `Task` is an interface, we need to `nil` the very last index, or it may not become eligible for garbage collection.

query/workqueue.go
```diff
@@ -0,0 +1,56 @@
+package query
+
+// Task is an interface that have a method for returning their index in the
+// work queue.
+type Task interface {
+	// Index returns this Task's index in the work queue.
+	Index() int
```
Roasbeef: Is the index to convey anything order than a target ordering for consumption?

query/workqueue_test.go
```diff
@@ -0,0 +1,90 @@
...
+			t.Fatalf("wrong index: %d", peek.index)
+		}
+	}
+
+	// Insert 3 elements with index 0.
+	for j := 0; j < 3; j++ {
+		heap.Push(q, tasks[0])
+	}
+
+	for i := numTasks/2 - 3; i < numTasks; i++ {
```
Roasbeef: The `3` value here can be lifted into a `numNewTasks` (or something else) constant.

query/interface.go
```diff
@@ -0,0 +1,126 @@
...
+	// message received from the peer that the request was made to. It
+	// should validate the response against the request made, and return
+	// whether the request was answered by this particular response.
+	//
+	// The first bool returned indicates whether the request was fully
+	// answered by this response. If the response did not fully answer the
+	// request, the second boolean indicates whether the response did
+	// make progress towards answering the request. This is used for the
+	// requests types where more than one response is expected.
+	HandleResp func(req, resp wire.Message, peer string) (bool, bool)
```
Roasbeef: Having two bools as a return parameter makes the function closure more difficult to reason about IMO. Instead, we can adopt a lightweight struct to handle the caller's response to the scheduler: 
```go
type QueryState {
    Finished bool
    ToBeContinued bool 
}

```

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+		w.peer.QueueMessageWithEncoding(
+			job.Req, nil, job.options.encoding,
+		)
+
+		// Wait for the correct response to be received from the peer,
+		// or an error happening.
+		var (
+			resp    wire.Message
+			jobErr  error
+			timeout = time.After(job.options.timeout)
```
Roasbeef: If we want to allow _multiple_ in flight requests (and IMO we should in order to take advantage of batching and pipelining over high latency links), then this timeout value should scale depending on the number of outstanding requests. This addresses the case where we timeout a filter query early due to head of the line blocking in the request queue of the remote peer we're requesting data from. 

Architecture wise, in order to allow multiple in-flight requests, we can keep the design _mostly_ the same, but would then have _multiple_ workers per peer. The scheduler can them inform the worker the timeout value that should be used, as it has a global view of all outstanding requests.

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+
+		switch jobErr {
+
+		// If the worker experienced a timeout trying to query this
+		// peer, we wait here for a given cooldown period before
+		// fetching the next job. This is to give other peers the
+		// chance to pick up the jobs.
+		case ErrQueryTimeout:
+			select {
+			case <-time.After(job.options.peerCooldown):
```
Roasbeef: Perhaps we should _randomize_ the cooldown to avoid a state where we're underutilizing the network bandwidth since all peers are paused.

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+	options *queryOptions
+	errChan chan error
+}
+
+// Config holds the configuration options for a new WorkManager.
+type Config struct {
+	// ConnectedPeers is a function that returns a channel where all
+	// connected peers will be sent. It is assumed that all current peers
+	// will be sent imemdiately, and new peers as they connect.
+	ConnectedPeers func() (<-chan Peer, func(), error)
```
Roasbeef: `godoc` should define what the function closure return value is.

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+	cfg *Config
+
+	// newBatches is a channel where new batches of queries will be sent to
+	// the workDispatcher.
+	newBatches chan *batch
+
+	// runWorker is function closure that should start a new worker for the
+	// given peer. We make this configurable to easily mock the worker used
+	// during tests.
+	runWorker func(peer Peer, nextJob <-chan *queryTask,
```
Roasbeef: Shouldn't the created worker be returned so we can manage its lifecycle externally?

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+// It makes sure to prioritize the queries in the order they come in, such that
+// early queries will be attemped completed first.
+//
+// NOTE: MUST be run as a goroutine.
+func (w *WorkManager) workDispatcher(nextJob chan<- *queryTask,
+	jobResults <-chan *jobResult) {
+
+	defer w.wg.Done()
+
+	// Init a work queue which will be used to sort the incoming queries in
```
Roasbeef: a -> is a

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+	work := &workQueue{}
+	heap.Init(work)
+
+	type batchProgress struct {
+		rem     int
+		errChan chan error
+	}
+
+	// We set up a batch index counter to keep track of batches that still
+	// have queries in in flight. This let us track when all queries for a
```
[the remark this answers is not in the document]
Roasbeef: let -> lets

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+	// batch has been finished, and return an (non-)error to the caller.
+	batchIndex := 0
+	currentBatches := make(map[int]*batchProgress)
+
+	// When the work dispatcher exits, we'll loop trhough the remaining
```
Roasbeef: trhough -> through

neutrino.go
```diff
@@ -162,19 +170,23 @@ type ServerPeer struct {
 	// any one time. The mutex is for subscribe/unsubscribe functionality.
 	// The sends on these channels WILL NOT block; any messages the channel
 	// can't accept will be dropped silently.
-	recvSubscribers map[spMsgSubscription]struct{}
-	mtxSubscribers  sync.RWMutex
+	// TODO(halseth): remove one of the maps when all queries go through
+	// work manager.
+	recvSubscribers  map[spMsgSubscription]struct{}
+	recvSubscribers2 map[msgSubscription]struct{}
```
Roasbeef: Why are two maps needed? Some commentary in the commit message body would be useful here.

neutrino.go
```diff
@@ -476,6 +488,23 @@ func (sp *ServerPeer) OnRead(_ *peer.Peer, bytesRead int, msg wire.Message,
...
+		// Quickly determine if this subscription has been canceled, if
+		// so delete it.
+		select {
+		case <-subscription.quitChan:
+			delete(sp.recvSubscribers2, subscription)
+			continue
+		default:
+		}
+
+		go func(subscription msgSubscription) {
```
Roasbeef: Don't we want to ensure these messages are delivered in order to the subscriber? I guess the same goes for the above.

neutrino.go
```diff
@@ -485,6 +514,43 @@ func (sp *ServerPeer) subscribeRecvMsg(subscription spMsgSubscription) {
...
+// A compile-time check to ensure that ServerPeer implements the query.Peer
+// interface.
+var _ query.Peer = (*ServerPeer)(nil)
+
+// SubscribeRecvMsg adds a OnRead subscription to the peer. All bitcoin
+// messages received from this peer will be sent on the returned channel. A
+// closure is also returned, that should be called to cancel the subscription.
+//
+// NOTE: Part of the query.Peer interface.
+func (sp *ServerPeer) SubscribeRecvMsg() (<-chan wire.Message, func()) {
```
Roasbeef: Why isn't the old `subscribeRecvMsg` sufficient?

neutrino.go
```diff
@@ -1265,6 +1274,28 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
...
+			return false
+		}
+
+		// Keep non-canceled subscribers around.
+		s.peerSubscribers[i] = sub
+		i++
+	}
+
+	// Re-align the slice to only active subscribers.
+	s.peerSubscribers = s.peerSubscribers[:i]
```
Roasbeef: Same comment here re GC leak.

blockmanager.go
```diff
@@ -873,6 +878,7 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 	//
 	// TODO(roasbeef): extract to func to test
 	currentInterval := startingInterval
+	endHeightRange := curHeight
```
Roasbeef: Why's this change necessary?

blockmanager.go
```diff
@@ -931,38 +937,36 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
+	headerChan := make(chan *wire.MsgCFHeaders, len(queryMsgs))
 
-		// Callback to process potential replies. Always called from
-		// the same goroutine as the outer function, so we don't have
-		// to worry about synchronization.
-		func(sp *ServerPeer, query wire.Message,
-			resp wire.Message) bool {
+	var msgs []*query.Query
+	for _, m := range queryMsgs {
+		handleResp := func(query, resp wire.Message,
```
Roasbeef: In order to chop this method up a bit, we can instead make this function closure an actual function, and parameterize it with the request context.

blockmanager_test.go
```diff
@@ -360,26 +382,37 @@ func TestBlockManagerInitialInterval(t *testing.T) {
...
+					msgs[index], &r, "",
+				)
+
+				if !done {
 					t.Fatalf("got response false on "+
 						"send of index %d: %v",
 						index, testDesc)
 				}
 
+				continue
```
Roasbeef: Stray continue?

blockmanager.go
```diff
@@ -1014,109 +1017,153 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
-				queryResponses[checkPointIndex] = r
+	// Hand the queries to the work manager, and consume the verified
+	// responses as they come back.
+	errChan := b.cfg.QueryAccess.Query(msgs)
 
-				return true
-			}
+	// Keep waiting for more headers as long as we haven't received an
+	// answer for our last checkpoint, and no error is encountered.
+	for {
```
Roasbeef: Similar comment re extracting the response processing into a distinct method.

query.go
```diff
@@ -234,340 +232,6 @@ const (
 // and provide some presets (including the ones below) prior to factoring out
 // the query API into its own package?
 
-// queryChainServiceBatch is a helper function that sends a batch of queries to
```
Roasbeef: What about the rest?

blockmanager.go
halseth: One is a interface, the other not.

query/workqueue.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/eb2c03ed1e3ed904f3dd3755d68262851164d445

query/workqueue.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/eb2c03ed1e3ed904f3dd3755d68262851164d445

blockmanager.go
```diff
@@ -98,12 +99,40 @@ type txMsg struct {
 	peer *ServerPeer
 }
 
+// blockManagerCfg holds options and dependencies needed by the blockManager
+// during operation.
+type blockManagerCfg struct {
+	// ChainParams is the chain that we're running on.
+	ChainParams chaincfg.Params
+
+	BlockHeaders     headerfs.BlockHeaderStore
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/7bee0f3a0bd38b9fae2524497e3c16382f7fcc68

blockmanager.go
```diff
@@ -1207,6 +1220,62 @@ func (b *blockManager) writeCFHeadersMsg(msg *wire.MsgCFHeaders,
...
+		Hash:   header.BlockHash(),
+	}
+
+	_, regHeight, err := b.cfg.RegFilterHeaders.ChainTip()
+	if err != nil {
+		return nil, err
+	}
+
+	for uint32(bs.Height) > height {
+		header, _, err := b.cfg.BlockHeaders.FetchHeader(&bs.Hash)
```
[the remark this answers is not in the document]
halseth: Cool, will keep in mind

query/interface.go
```diff
@@ -0,0 +1,126 @@
...
+// applyQueryOptions updates a queryOptions set with functional options.
+func (qo *queryOptions) applyQueryOptions(options ...QueryOption) {
+	for _, option := range options {
+		option(qo)
+	}
+}
+
+// Timeout is a query option that lets the query know how long to wait for each
+// peer we ask the query to answer it before moving on.
+func Timeout(timeout time.Duration) QueryOption {
```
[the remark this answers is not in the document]
halseth: Cooldown is no more 😎

query/interface.go
```diff
@@ -0,0 +1,126 @@
...
+	// answered by this response. If the response did not fully answer the
+	// request, the second boolean indicates whether the response did
+	// make progress towards answering the request. This is used for the
+	// requests types where more than one response is expected.
+	HandleResp func(req, resp wire.Message, peer string) (bool, bool)
+}
+
+// QueryAccess is an interface defining the API for making queries to bitcoin
+// peers.
+type QueryAccess interface {
```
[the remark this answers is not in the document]
halseth: Name `Dispatcher`: https://github.com/lightninglabs/neutrino/pull/179/commits/1fd8ebabe57867f0e89bb7b709cb07293b495840

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+	for {
+		var job *queryTask
+
+		select {
+		// Poll a new job from the nextJob channel.
+		case job = <-w.nextJob:
+
+		// Ignore any message received while not working on a task.
+		case <-msgChan:
+			continue
```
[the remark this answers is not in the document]
halseth: Added trace logs

query/worker_test.go
```diff
@@ -0,0 +1,388 @@
...
+// to its query after first receiving some non-matching messages.
+func TestWorkerIgnoreMsgs(t *testing.T) {
+	t.Parallel()
+
+	ctx, err := startWorker()
+	if err != nil {
+		t.Fatalf("unable to start worker: %v", err)
+	}
+
+	nextJob := ctx.nextJob
```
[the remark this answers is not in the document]
halseth: Leftover from eariler iteration. Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/f0576f5b0310a2bf860449770a1479ceb00473c4

query/worker_test.go
```diff
@@ -0,0 +1,388 @@
...
+	// The request should be given to the peer.
+	select {
+	case <-peer.requests:
+	case <-time.After(time.Second):
+		t.Fatalf("request not sent")
+	}
+
+	// Don't anwer the query. This should trigger a timeout, and the worker
+	// should respond with an error result.
+	var result *jobResult
```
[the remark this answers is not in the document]
halseth: It will trigger the response, so don't think it is needed.

query/worker_test.go
```diff
@@ -0,0 +1,388 @@
...
+	// the delay will be larger than the query timeout, but since we are
+	// making progress, the timeout won't trigger.
+	for i := 0; i < 5; i++ {
+		select {
+		case peer.responses <- progressResp:
+		case <-time.After(time.Second):
+			t.Fatalf("resp not received")
+		}
+
+		time.Sleep(20 * time.Millisecond)
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/179/commits/767d5639fe38f60ed2203fd0814403878e977585

query/worker.go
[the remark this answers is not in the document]
halseth: Good idea! Added: https://github.com/lightninglabs/neutrino/pull/179/commits/cba4d208656bd297c53f48df7b138140ddc89d0a

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+// early queries will be attemped completed first.
+//
+// NOTE: MUST be run as a goroutine.
+func (w *WorkManager) workDispatcher(nextJob chan<- *queryTask,
+	jobResults <-chan *jobResult) {
+
+	defer w.wg.Done()
+
+	// Init a work queue which will be used to sort the incoming queries in
+	// a first come first serverd fashion.
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/232aaf0d8a5798306264cc44a90d6e7a110940dd

query/workmanager.go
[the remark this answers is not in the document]
halseth: Fixed

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+	// have queries in in flight. This let us track when all queries for a
+	// batch has been finished, and return an (non-)error to the caller.
+	batchIndex := 0
+	currentBatches := make(map[int]*batchProgress)
+
+	// When the work dispatcher exits, we'll loop trhough the remaining
+	// batches and send on their error channel.
+	defer func() {
+		for _, b := range currentBatches {
+			b.errChan <- errors.New("WorkManager shutting down")
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/179/commits/38e1bb328ef051ef41a6b5a264cf0294cd4585a6

query/workmanager.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Add query messages to the task queue.
+	select {
+	case w.newBatches <- &batch{
+		queries: queries,
+		options: qo,
+		errChan: errChan,
+	}:
+	case <-w.quit:
+		errChan <- fmt.Errorf("workmanager shutting down")
```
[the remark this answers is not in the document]
halseth: Done

query/workmanager_test.go
```diff
@@ -0,0 +1,256 @@
...
+			results: results,
+		}
+	}
+
+	nextJob := make(chan *queryTask)
+	responses := make(chan *jobResult)
+
+	// Start the peer listener.
+	wm.wg.Add(1)
+	go wm.peerListener(nextJob, responses)
```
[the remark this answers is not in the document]
halseth: Fixed.

query/workmanager_test.go
```diff
@@ -0,0 +1,256 @@
...
+				t.Fatalf("wrong job")
+			}
+		case <-time.After(time.Second):
+			t.Fatalf("result not handled")
+		}
+	}
+}
+
+// TestWorkManagerWorkDispatcher tests that the workDispatcher goroutine
+// propely schedules the incoming queries in the order of their batch and sends
```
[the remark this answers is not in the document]
halseth: fixed

neutrino.go
```diff
@@ -662,6 +669,8 @@ type ChainService struct {
 	broadcaster          *pushtx.Broadcaster
 	banStore             banman.Store
 
+	peerSubscribers []*peerSubscription
```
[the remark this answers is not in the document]
halseth: Done. https://github.com/lightninglabs/neutrino/pull/179/commits/742a348b3bb60ba44b2e85f2df663a455c391be3

notifications.go
halseth: Yep!

blockmanager.go
halseth: Yep, fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/dab9f9f4046ea7711cbe1481ef4f52f7552f38f3

blockmanager.go
halseth: Done

blockmanager.go
halseth: done after method extraction

query/worker.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/4f4938e3b6db3b612f640c307af0ce946eefc716

query/worker.go
halseth: Idea is that the peer will be deprioritised for next attempt

blockmanager.go
halseth: These closures will be removed.

query/workqueue.go
halseth: Only priority in which request should be given to peers first

query/interface.go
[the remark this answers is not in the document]
halseth: Added `QueryProgress`: https://github.com/lightninglabs/neutrino/pull/179/commits/e4c8b3376f65550bda47ec697bd0970ab66506ab

query/interface.go
```diff
@@ -0,0 +1,126 @@
...
+// for the query messages.
+func Encoding(encoding wire.MessageEncoding) QueryOption {
+	return func(qo *queryOptions) {
+		qo.encoding = encoding
+	}
+}
+
+// Query is the main struct that defines a bitcoin network query to be sent to
+// connected peers.
+type Query struct {
```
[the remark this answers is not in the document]
halseth: Went with `Request`: https://github.com/lightninglabs/neutrino/pull/179/commits/e4c8b3376f65550bda47ec697bd0970ab66506ab

query/interface.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// QueryAccess is an interface defining the API for making queries to bitcoin
+// peers.
+type QueryAccess interface {
+	// Query distributes the slice of queries to the set of connected
+	// peers. It returns an error channel where the final result of the
+	// batch of queries will be sent. Responses for the individual queries
+	// should be handled by the response handler of each Query.
+	Query(queries []*Query, options ...QueryOption) chan error
```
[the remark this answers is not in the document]
halseth: No, but they are dmultiplexed in the sense that the response will be called for the exact `Request`s response handler

query/worker.go
```diff
@@ -0,0 +1,194 @@
...
+	*Query
+}
+
+// queryTask should satisfy the Task interface in order to be sorted by the
+// workQueue.
+var _ Task = (*queryTask)(nil)
+
+// Index returns the queryTask's index within the work queue.
+func (q *queryTask) Index() int {
+	return q.index
```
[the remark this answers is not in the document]
halseth: To prioritise rescheduled queries

query/worker.go
halseth: Cooldown is no more

query/worker.go
halseth: Yeah, will revisit when moving to multiple in-flight support

query/workmanager.go
halseth: Done: https://github.com/lightninglabs/neutrino/pull/179/commits/38e1bb328ef051ef41a6b5a264cf0294cd4585a6

query/workmanager.go
halseth: Init uses as a verb.

neutrino.go
halseth: Added to commit message.

neutrino.go
halseth: Doesn't really matter, as it is handled by response handler code

neutrino.go
halseth: See commit message

neutrino.go
halseth: Fixed

blockmanager.go
halseth: Earlier the `queryBatch` method would call the response handle every time a new response came in. Now this code is not part of the response handler anymore, and we must know when to exit the loop.

blockmanager.go
halseth: Extracted: https://github.com/lightninglabs/neutrino/pull/179/commits/04aa607a3914fda3b1c3ed9ff7dd603a410167cb

blockmanager_test.go
halseth: Yep, leftover needed to make tests pass at one point. Fixed here: https://github.com/lightninglabs/neutrino/pull/179/commits/dab9f9f4046ea7711cbe1481ef4f52f7552f38f3

query.go
halseth: removing one at a time...

halseth:
Addressed review and pushed out a new version that should fix sync tests.

We now don't use a `peerCooldown` to avoid scheduling queries on bad peers. Instead a new module `PeerRanking` keeps a counter for each peer, and ranks peer according to how many failed queries they have had. That way we will always first try to schedule queries on free peers with high rank.

We also move to dynamic timeouts. A request will start out with a 2 second timeout, which is doubled if no response is received in time. These two changes should help sync tests, as we will quickly schedule on and timeout on the bad peers, which them will have their ranking decreases.

Also moved the cfilter query change to its own PR, as this one seemed big enough: https://github.com/lightninglabs/neutrino/pull/182

halseth:
New version pushed. PTAL

Roasbeef:
Race in tests: 
```
WARNING: DATA RACE
Write at 0x00c000be3466 by goroutine 14:
  github.com/lightninglabs/neutrino.(*blockManager).getCheckpointedCFHeaders()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager.go:1161 +0x15de
  github.com/lightninglabs/neutrino.TestBlockManagerInitialInterval()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager_test.go:439 +0xa75
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:909 +0x199
Previous read at 0x00c000be3466 by goroutine 212:
  github.com/lightninglabs/neutrino.verifyCheckpoint()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager.go:1378 +0x9b
  github.com/lightninglabs/neutrino.(*checkpointedCFHeadersQuery).handleResponse()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager.go:918 +0x25c
  github.com/lightninglabs/neutrino.(*checkpointedCFHeadersQuery).handleResponse-fm()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager.go:862 +0x87
  github.com/lightninglabs/neutrino.TestBlockManagerInitialInterval.func1.1()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager_test.go:405 +0x328
Goroutine 14 (running) created at:
  testing.(*T).Run()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:960 +0x651
  testing.runTests.func1()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:1202 +0xa6
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:909 +0x199
  testing.runTests()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:1200 +0x521
  testing.(*M).Run()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:1117 +0x2ff
  main.main()
      _testmain.go:84 +0x223
Goroutine 212 (running) created at:
  github.com/lightninglabs/neutrino.TestBlockManagerInitialInterval.func1()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager_test.go:375 +0x48b
  github.com/lightninglabs/neutrino.(*mockDispatcher).Query()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager_test.go:44 +0x89
  github.com/lightninglabs/neutrino.(*blockManager).getCheckpointedCFHeaders()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager.go:1079 +0xc41
  github.com/lightninglabs/neutrino.TestBlockManagerInitialInterval()
      /home/travis/gopath/src/github.com/lightninglabs/neutrino/blockmanager_test.go:439 +0xa75
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.1.linux.amd64/src/testing/testing.go:909 +0x199
```

halseth:
Race fixed.

notifications.go
```diff
@@ -181,7 +213,30 @@ func (s *ChainService) ConnectedCount() int32 {
...
+// imemdiately, and new peers as they connect.
+func (s *ChainService) ConnectedPeers() (<-chan query.Peer, func(), error) {
+	replyChan := make(chan subConnPeersReply, 1)
 
+	select {
+	case s.query <- subConnPeersMsg{
+		reply: replyChan,
+	}:
+	case <-s.quit:
+		return nil, nil, fmt.Errorf("ChainService exiting")
```
wpaulino: Nit: use `ErrShuttingDown`.

query/workmanager.go
```diff
@@ -0,0 +1,447 @@
...
+						"%v: %v ", batchNum, batch.rem)
+
+					// If this was the last query in flight
+					// for this batch, we can notify that
+					// it finished, and delete it.
+					if batch.rem == 0 {
+						batch.errChan <- nil
+						delete(currentBatches, batchNum)
+						log.Tracef("Batch %v done",
+							batchNum)
```
wpaulino: Ran into a deadlock:
```
1 @ 0x1032d60 0x1006edb 0x1006eb1 0x1006c95 0x152677f 0x1060bd1
#	0x152677e	github.com/lightninglabs/neutrino/query.(*WorkManager).workDispatcher+0xf8e	/Users/wilmer/git/neutrino/query/workmanager.go:378

1 @ 0x1032d60 0x1043350 0x104333b 0x1042fa2 0x107e954 0x155d423 0x155d3f4 0x1ad59ac 0x1a81e3d 0x1aeb662 0x103298e 0x1060bd1
#	0x1042fa1	sync.runtime_Semacquire+0x41						/usr/local/go/src/runtime/sema.go:56
#	0x107e953	sync.(*WaitGroup).Wait+0x63						/usr/local/go/src/sync/waitgroup.go:130
#	0x155d422	github.com/lightninglabs/neutrino/query.(*WorkManager).Stop+0xb2	/Users/wilmer/git/neutrino/query/workmanager.go:137
#	0x155d3f3	github.com/lightninglabs/neutrino.(*ChainService).Stop+0x83		/Users/wilmer/git/neutrino/neutrino.go:1548
#	0x1ad59ab	github.com/lightningnetwork/lnd.initNeutrinoBackend.func3+0x3b		/Users/wilmer/git/lnd/chainregistry.go:769
#	0x1a81e3c	github.com/lightningnetwork/lnd.Main+0x269c				/Users/wilmer/git/lnd/lnd.go:589
#	0x1aeb661	main.main+0x31								/Users/wilmer/git/lnd/cmd/lnd/main.go:14
#	0x103298d	runtime.main+0x21d							/usr/local/go/src/runtime/proc.go:203
```

I believe it's due to a missing continue here. I think we'll still want to reward the peer before doing so though.

blockmanager.go
```diff
@@ -1026,71 +1026,69 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
+				curHeader, curHeight, err = b.writeCFHeadersMsg(r, store)
 				if err != nil {
 					panic(fmt.Sprintf("couldn't write "+
 						"cfheaders msg: %v", err))
 				}
+
+				// Update the next interval to write to reflect our
+				// current height.
+				currentInterval = curHeight / wire.CFCheckptInterval
+
```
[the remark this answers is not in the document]
halseth: Fixed.

query/interface.go
```diff
@@ -0,0 +1,145 @@
...
+	// used for the requests types where more than one response is
+	// expected.
+	Progressed bool
+}
+
+// Request is the main struct that defines a bitcoin network query to be sent to
+// connected peers.
+type Request struct {
+	// Req is the message request to send.
+	Req wire.Message
```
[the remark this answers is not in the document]
halseth: Makes sense to use naming `Req` and `Resp`?

query/worker.go
```diff
@@ -0,0 +1,240 @@
...
+// query's response handler, and poll more work for the peer when it has
+// successfully received a response to the request.
+type worker struct {
+	// nextJob is a channel of queries to be distributed, where the worker
+	// will poll new work from.
+	nextJob chan *queryJob
+}
+
+// NewWorker creates a new worker.
+func NewWorker() *worker {
```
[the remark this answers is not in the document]
halseth: Don't think that's a problem as long as it is adhering to the exported interface.

query/worker.go
```diff
@@ -0,0 +1,240 @@
...
+	defer cancel()
+
+	for {
+		log.Tracef("Worker %v waiting for more work", peer.Addr())
+
+		var job *queryJob
+		select {
+		// Poll a new job from the nextJob channel.
+		case job = <-w.nextJob:
+			log.Tracef("Worker %v picked up job with index %v",
```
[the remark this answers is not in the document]
halseth: Future work? 😬

query/worker.go
```diff
@@ -0,0 +1,240 @@
...
+			peer: peer,
+			err:  jobErr,
+		}:
+		case <-quit:
+			return
+		}
+
+		// If the peer disconnected, we can exit immediately.
+		if jobErr == ErrPeerDisconnected {
+			log.Debugf("Peer %v for worker disconnected, exiting",
```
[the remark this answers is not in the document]
halseth: I think it is worth keeping, as it signals that the job result was successfully sent?

query/peer_rank.go
```diff
@@ -0,0 +1,82 @@
+package query
+
+import (
+	"sort"
+)
+
+const (
+	bestScore    = 0
+	defaultScore = 4
+	worstScore   = 8
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/5753f034fbec0c442eacf0525917c39b47cde991

query/peer_rank.go
```diff
@@ -25,6 +25,10 @@ func NewPeerRanking() *peerRanking {
 	}
 }
 
+// A compile time check to ensure peerRanking satisfies the PeerRanking
+// interface.
+var _ PeerRanking = (*peerRanking)(nil)
```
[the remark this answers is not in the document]
halseth: I think either way is fine, order doesn't make much diff as they depend loosely on each other.

query/workmanager.go
```diff
@@ -0,0 +1,447 @@
...
+import (
+	"container/heap"
+	"errors"
+	"sync"
+	"time"
+)
+
+const (
+	minQueryTimeout = 2 * time.Second
+	maxQueryTimeout = 32 * time.Second
```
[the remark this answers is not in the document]
halseth: Added: https://github.com/lightninglabs/neutrino/pull/179/commits/5b6fca728da8c3db5758d11cdc16e24038d7d020

Keeping them independent for now, as they can be tuned independently .

query/workmanager.go
```diff
@@ -0,0 +1,447 @@
...
+// callers, and dispatches these to active workers.  It makes sure to
+// prioritize the queries in the order they come in, such that early queries
+// will be attemped completed first.
+//
+// NOTE: MUST be run as a goroutine.
+func (w *WorkManager) workDispatcher() {
+	defer w.wg.Done()
+
+	// Get a peer subscription.
+	peersConnected, cancel, err := w.cfg.ConnectedPeers()
```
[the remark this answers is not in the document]
halseth: Yep, good idea: https://github.com/lightninglabs/neutrino/pull/179/commits/74bc9b2320136f81b1ded4bc44836849426e998d

query/workmanager.go
halseth: Good catch! Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/54b63afc0baa3250860d10b0081297efb56b004a

notifications.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/1ecb71a939828bd34c05d21f3c45afc34dc8d922

blockmanager.go
```diff
@@ -1026,103 +1042,144 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
-				// current height.
-				currentInterval = curHeight / wire.CFCheckptInterval
+	// Keep waiting for more headers as long as we haven't received an
+	// answer for our last checkpoint, and no error is encountered.
+	for {
+		var r *wire.MsgCFHeaders
+		select {
+		case r = <-headerChan:
+		case err := <-errChan:
+			if err != nil {
```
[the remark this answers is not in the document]
halseth: Useful to log in case the work manager shuts down because of some error without the query shutting down.

blockmanager.go
```diff
@@ -838,6 +838,127 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 	return err
 }
 
+// checkpointedCFHeadersQuery holds all information necessary to perform and
+// handle a query for checkpointed filter headers.
+type checkpointedCFHeadersQuery struct {
```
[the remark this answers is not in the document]
halseth: Yeah, agreed, but might be a bit of work to do at this point....

query/workmanager.go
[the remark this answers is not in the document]
halseth: Remembered why I didn't do this in the first place, and reverted the change. `ConnectedPeers` queries the server which won't respond to the query before the `WorkManager` has started, causing a deadlock. Added a comment.

blockmanager.go
```diff
@@ -1107,16 +1104,16 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 // constructed cfheader in this range as this lets callers populate the prev
 // filter header field in the next message range before writing to disk.
 func (b *blockManager) writeCFHeadersMsg(msg *wire.MsgCFHeaders,
-	store *headerfs.FilterHeaderStore) (*chainhash.Hash, error) {
+	store *headerfs.FilterHeaderStore) (*chainhash.Hash, uint32, error) {
```
Roasbeef: Should have the `godoc` updated to reflect the new return value.

blockmanager.go
```diff
@@ -1012,10 +1016,6 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
 				log.Debugf("Got response for headers at "+
 					"height=%v, only at height=%v, stashing",
 					startHeight, curHeight)
-
-				queryResponses[checkPointIndex] = r
```
Roasbeef: So how are out of order responses handled now? Do we even needs this block any longer?

blockmanager.go
```diff
@@ -1026,71 +1026,68 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
 						"cfheaders msg: %v", err))
 				}
+
+				// Update the next interval to write to reflect our
+				// current height.
+				currentInterval = curHeight / wire.CFCheckptInterval
+			}
+
+			// If the current interval is beyond our checkpoints,
+			// we are done.
```
Roasbeef: Now sure how much of a simplification this commit actually is overall...no context in the commit message.

log.go
```diff
@@ -39,6 +40,7 @@ func UseLogger(logger btclog.Logger) {
 	blockntfns.UseLogger(logger)
 	pushtx.UseLogger(logger)
 	connmgr.UseLogger(logger)
+	query.UseLogger(logger)
```
Roasbeef: IMO it's an issue that right now these all use the same logger as it means there's no natmepsacing between the messages. In `lnd`, we can't go `BTCN=debug` without getting the flood of all the peer messages, and there's no way to turn that off. Not blocking for this PR, but just want to not that here as we add yet another hidden sub-logger.

query/interface.go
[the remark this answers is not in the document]
Roasbeef: req/resp sgtm 

In terms of naming `QueryProgress` could also just be `Progress`.

query/worker.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// jobResult is the final result of the worker's handling of the queryJob.
+type jobResult struct {
+	job  *queryJob
+	peer Peer
+	err  error
+}
+
+// worker is responsible for polling work from its work queue, and hand it to
```
Roasbeef: hand -> handing

query/worker.go
```diff
@@ -0,0 +1,239 @@
...
+// the associated peer. It validates incoming responses with the current
+// query's response handler, and poll more work for the peer when it has
```
Roasbeef: poll -> polls

query/worker.go
```diff
@@ -0,0 +1,239 @@
...
+// Run starts the worker and associates it with the given peer. peer is the
+// peer this worker will supply with queries, and handle responses from.
+// Results for any query handled by this worker will be delivered on the
+// results channel. quit can be closed to immediately make the worker exit.
+//
+// The method is blocking, and should be started in a goroutine. It will run
+// until the peer disconnects or the worker is told to quit.
+//
+// NOTE: Part of the Worker interface.
+func (w *worker) Run(peer Peer, results chan<- *jobResult,
```
Roasbeef: Non-blocking but any reason we don't bind the worker to a peer at creation time?

query/worker.go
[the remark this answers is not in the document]
Roasbeef: IIRC, it'll force callers to declare the type up front in areas where it isn't passed in as an argument with a type of the interface.

query/worker.go
```diff
@@ -0,0 +1,239 @@
...
+				if !progress.Finished {
+					// If it did make progress we reset the
+					// timeout. This ensures that the
+					// queries with multiple responses
+					// expected won't timeout before all
+					// responses have been handled.
+					// TODO(halseth): separete progress
+					// timeout value.
+					if progress.Progressed {
+						timeout = time.After(
```
Roasbeef: The old timeout value is never stopped. If we use a `time.NewTimer`, then we'll be able to stop it.

query/worker.go
```diff
@@ -0,0 +1,239 @@
...
+							job.timeout,
+						)
+					}
+					continue Loop
```
Roasbeef: What happens to the response in this case if we continue? I guess it's OK since it's been handed to the caller already via the `HandleResp` method?

query/peer_rank.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	// bestScore is the best score a peer can get after multiple rewards.
+	bestScore = 0
+
+	// defaultScore is the score given to a peer when it hasn't been
+	// rewarded or punished.
+	defaultScore = 4
+
+	// qorstScore is the worst score a peer can get after multiple
```
Roasbeef: qorstScore -> worstScore

query/peer_rank_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// We'll try to order half of the peers.
+	peers = peers[:numPeers/2]
+	ranking.Order(peers)
+
+	// Since no peer was rewarded or punished, their order
+	// should be unchanged.
+	for i := 0; i < numPeers/2; i++ {
+		p := fmt.Sprintf("peer%d", i)
+		if peers[i] != p {
+			t.Fatalf("expecred %v, got %v", p, peers[i])
```
Roasbeef: expecred -> expected

query/peer_rank_test.go
```diff
@@ -0,0 +1,59 @@
...
+		for j := 0; j <= i; j++ {
+			ranking.Punish(peers[j])
+		}
+	}
+
+	ranking.Order(peers)
+	for i := 0; i < numPeers/2; i++ {
+		p := fmt.Sprintf("peer%d", numPeers/2-i-1)
+		if peers[i] != p {
+			t.Fatalf("expecred %v, got %v", p, peers[i])
```
Roasbeef: expecred -> expected

query/workmanager.go
```diff
@@ -0,0 +1,455 @@
...
+	// worker exit.
+	//
+	// The method is blocking, and should be started in a goroutine. It
+	// will run until the peer disconnectes or the worker is told to quit.
+	Run(peer Peer, results chan<- *jobResult, quit <-chan struct{})
+
+	// NewJob returns a channel where work that is to be handled by the
+	// worker can be sent. If the worker reads a queryJob from this
+	// channel, it is guaranteed that a response will eventually be
+	// deliverd on the results channel (except when the quit channel has
```
Roasbeef: deliverd -> delivered

query/workmanager.go
```diff
@@ -0,0 +1,455 @@
...
+		cfg:        cfg,
+		newBatches: make(chan *batch),
+		jobResults: make(chan *jobResult),
+		quit:       make(chan struct{}),
+	}
+}
+
+// Start starts the WorkManager.
+func (w *WorkManager) Start() error {
+
```
Roasbeef: Style nit: unnecessary newline.

query/workmanager.go
```diff
@@ -0,0 +1,455 @@
...
+	type batchProgress struct {
+		timeout <-chan time.Time
+		rem     int
+		errChan chan error
+	}
+
+	// We set up a batch index counter to keep track of batches that still
+	// have queries in flight. This lets us track when all queries for a
+	// batch has been finished, and return an (non-)error to the caller.
```
Roasbeef: has -> have

query/workmanager.go
```diff
@@ -0,0 +1,455 @@
...
+	batchIndex := uint64(0)
+	currentBatches := make(map[uint64]*batchProgress)
+
+	// When the work dispatcher exits, we'll loop through the remaining
+	// batches and send on their error channel.
+	defer func() {
+		for _, b := range currentBatches {
+			b.errChan <- ErrWorkManagerShuttingDown
```
Roasbeef: Needs a select on quit?

query/workmanager.go
```diff
@@ -0,0 +1,455 @@
...
+	// query to the batch they are part of.
+	queryIndex := uint64(0)
+	currentQueries := make(map[uint64]uint64)
+
+	workers := make(map[string]*activeWorker)
+
+Loop:
+	for {
+		// If the work queue is non-empty, we'll take out the first
+		// element in order for distributing it to a worker.
```
Roasbeef: to distribute it to a worker

neutrino.go
```diff
@@ -1268,6 +1279,31 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
...
+	// peer.
+	i := 0
+	for _, sub := range s.peerSubscribers {
+		select {
+		case sub.peers <- sp:
+
+		// Ignore subscription if it has been cancelled.
+		case <-sub.cancel:
+			// Avoid GC leak.
+			s.peerSubscribers[i] = nil
```
Roasbeef: In this case `i` is never incremented, so it'll set the same slot to nil over and over again.

blockmanager.go
```diff
@@ -1026,102 +1042,144 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
-				lastHeight, r.PrevFilterHeader)
-
-			// If this is out of order but not yet written, we can
-			// verify that the checkpoints match, and then store
-			// them.
-			if startHeight > curHeight+1 {
-				log.Debugf("Got response for headers at "+
-					"height=%v, only at height=%v, stashing",
-					startHeight, curHeight)
+			select {
```
Roasbeef: Comments above never updated.

blockmanager.go
```diff
@@ -1026,102 +1042,144 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
+					"cfheaders msg: %v", err))
+			}
 
-		// Same quit channel we're watching.
-		b.quit,
-	)
+			// Update the next interval to write to reflect our
+			// current height.
+			currentInterval = curHeight / wire.CFCheckptInterval
+
```
Roasbeef: Style nit: extra new line.

blockmanager.go
```diff
@@ -838,6 +838,127 @@ func (b *blockManager) getUncheckpointedCFHeaders(
...
+			Req:        m,
+			HandleResp: c.handleResponse,
+		})
+	}
+	return reqs
+}
+
+// handleResponse is the internal response handler used for requests for this
+// CFHeaders query.
+func (c *checkpointedCFHeadersQuery) handleResponse(req, resp wire.Message,
```
Roasbeef: Hmm, I guess for interfaces, it doesn't matter if things are exported or not?

neutrino.go
```diff
@@ -710,7 +698,21 @@ func NewChainService(cfg Config) (*ChainService, error) {
...
+		BanPeer:          s.BanPeer,
+		DisconnectPeer:   s.DisconnectNodeByAddr,
+		GetBlock:         s.GetBlock,
+		queryBatch: func(msgs []wire.Message, f func(*ServerPeer,
+			wire.Message, wire.Message) bool, q <-chan struct{},
+			qo ...QueryOption) {
+			queryChainServiceBatch(&s, msgs, f, q, qo...)
+		},
+		queryAllPeers: s.queryAllPeers,
+	}, s.firstPeerConnect)
```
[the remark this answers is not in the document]
halseth: Done: https://github.com/lightninglabs/neutrino/pull/179/commits/889eeeda7e527cf0232c02e67d3e3ea7535f60ac

query/interface.go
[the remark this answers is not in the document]
halseth: Renamed to `Progress`: https://github.com/lightninglabs/neutrino/pull/179/commits/30748c1a04c846fabf96674cf07c7d00be7b7615

query/worker.go
halseth: Not really, probably leftover from earlier iterations of the tests. Agreed that is is better: https://github.com/lightninglabs/neutrino/pull/179/commits/3edcd829728fd318bf48ea0f4dcb80aab3ee894e

query/worker.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/05d21e979c729c095b1178911b008aad25b390b0

query/worker.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/05d21e979c729c095b1178911b008aad25b390b0

query/worker.go
halseth: Done. https://github.com/lightninglabs/neutrino/pull/179/commits/05d21e979c729c095b1178911b008aad25b390b0

query/worker.go
halseth: Yeah, it will have been handled by the response handler, and we only need to reset the timeout to ensure progressing queries don't timeout.

query/peer_rank.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/0f1443c7d3388be7076716b703524ac7cc58d66e

query/peer_rank_test.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/0f1443c7d3388be7076716b703524ac7cc58d66e

query/peer_rank_test.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/0f1443c7d3388be7076716b703524ac7cc58d66e

query/workmanager.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/01d1604c4b0ec42a9d30500cda1526ee3fd80fa1

query/workmanager.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/01d1604c4b0ec42a9d30500cda1526ee3fd80fa1

query/workmanager.go
halseth: `ErrChans` are buffered with 1, and always safe to send on once.

query/workmanager.go
halseth: `ErrChans` are buffered with 1, and always safe to send on once.

query/workmanager.go
```diff
@@ -0,0 +1,447 @@
...
+			go func() {
+				defer w.wg.Done()
+				defer close(onExit)
+
+				r.Run(peer, w.jobResults, w.quit)
+			}()
+
+		// A new result came back.
+		case result := <-w.jobResults:
+			log.Tracef("Result for job %v received. err=%v",
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/01d1604c4b0ec42a9d30500cda1526ee3fd80fa1

query/workmanager.go
```diff
@@ -0,0 +1,447 @@
...
+						log.Debugf("Canceled batch %v",
+							batchNum)
+					}
+
+				// If the query ended with any other error, put
+				// it back into the work queue.
+				default:
+					log.Warnf("Query(%d) failed with "+
+						"error: %v. Rescheduling",
+						result.job.index, result.err)
```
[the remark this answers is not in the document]
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/01d1604c4b0ec42a9d30500cda1526ee3fd80fa1

query/worker.go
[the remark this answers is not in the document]
halseth: I changed it to use the exported type, since it makes it nicer to match the method signature in `blockManagerCfg` in later commits: https://github.com/lightninglabs/neutrino/pull/179/commits/814cf6b9479fdba3bb384eaf76e9b4f198d75525

neutrino.go
halseth: Uh, you're right. Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/c7174a06f5a2cacf6d0ea8e356fd87e3ed60ba08

blockmanager.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/3b345191d69a765eae9fcfc962536b59a3309cf4

blockmanager.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/a0ea730f977296fb9fb166b5a33edd40371757c6

blockmanager.go
halseth: It is handled below by always writing the response to the map, then iterating over the map. If statement kept to keep the logging.

blockmanager.go
halseth: Expanded  commit message.

blockmanager.go
halseth: ?

query/interface.go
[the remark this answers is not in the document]
wpaulino: > Makes sense to use naming Req and Resp?

Just seems weird to have `Request.Req`, but fine as is.

neutrino.go
```diff
@@ -1250,6 +1261,31 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
 		s.addrManager.Good(sp.NA())
 	}
 
+	// We'll go through each peer subscriber and notify it about the added
+	// peer.
+	n := 0
+	for i, sub := range s.peerSubscribers {
+		select {
+		case sub.peers <- sp:
```
wpaulino: Should we be concerned about this blocking the main handler of the `ChainService` if a subscriber isn't ready to read from the channel?

query/interface.go
```diff
@@ -0,0 +1,145 @@
...
+
+// applyQueryOptions updates a queryOptions set with functional options.
+func (qo *queryOptions) applyQueryOptions(options ...QueryOption) {
+	for _, option := range options {
+		option(qo)
+	}
+}
+
+// Timeout is a query option that specifies the total time a query is allowed
+// to be retried before it is failed.
```
wpaulino: Nit: s/retried/tried/?

query/worker.go
```diff
@@ -0,0 +1,247 @@
...
+
+// NewWorker creates a new worker assosiated with the given peer.
+func NewWorker(peer Peer) *worker {
+	return &worker{
+		peer:    peer,
+		nextJob: make(chan *queryJob),
+	}
+}
+
+// Run starts the worker. The worker will supply its with queries, and handle
```
wpaulino: Nit: it's peer*

query/worker.go
```diff
@@ -0,0 +1,247 @@
...
+		case <-quit:
+			return
+		}
+
+		select {
+		// There is no point in queueing the request if the job already
+		// is canceled, so we check this quickly.
+		case <-job.cancelChan:
+			log.Tracef("Worker %v found job with index %v "+
+				"already canceled", peer.Addr(), job.Index())
```
wpaulino: Needs a `continue` after this?

query/worker.go
wpaulino: Looks like we could limit the scope of `resp` to only be within the `select` statement then? Its current position makes it seem as if it will be used later on.

query/worker.go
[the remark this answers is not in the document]
wpaulino: Maybe consolidate them into one? Don't see why it should be logged twice.

query/workmanager.go
```diff
@@ -0,0 +1,464 @@
...
+type batch struct {
+	requests []*Request
+	options  *queryOptions
+	errChan  chan error
+}
+
+// Worker is the interface that must be satisfied by workers managed by the
+// WorkManager.
+type Worker interface {
+	// Run starts the worker and associates it with the given peer. peer is
```
wpaulino: This comment is now outdated.

query/workmanager.go
```diff
@@ -0,0 +1,464 @@
...
+		// A new result came back.
+		case result := <-w.jobResults:
+			log.Tracef("Result for job %v received from peer %v "+
+				"(err=%v)", result.job.index,
+				result.peer.Addr(), result.err)
+
+			// Delete the job from the worker's active job, such
+			// that the slot gets opened for more work.
+			r, ok := workers[result.peer.Addr()]
+			if ok {
```
wpaulino: Why is this check needed?

query/workmanager.go
```diff
@@ -0,0 +1,464 @@
...
+			// Get the index of this query's batch, and delete it
+			// from the map of current queries, since we don't have
+			// to track it anymore.
+			batchNum := currentQueries[result.job.index]
+			delete(currentQueries, result.job.index)
+			batch := currentBatches[batchNum]
+
+			switch {
+			case result.err != nil:
+				switch {
```
wpaulino: Nit: it looks like we can merge this `switch` statement with the one above and save on indentation.

blockmanager.go
```diff
@@ -1012,114 +1025,155 @@ func (b *blockManager) getCheckpointedCFHeaders(checkpoints []*chainhash.Hash,
...
 			}
 
-			// If this is out of order stuff that's already been
-			// written, we can ignore it.
-			if lastHeight <= curHeight {
-				log.Debugf("Received out of order reply "+
-					"end_height=%v, already written", lastHeight)
-				return true
+			return query.Progress{
+				Finished:   true,
```
wpaulino: Don't we want this to be true once we've received all the headers?

wpaulino: Or does this only apply to a single query, rather than a batch?

wpaulino: Not sure how useful these `Finished`/`Progress` bools are given that the workers act upon a single message request and not a batch.

blockmanager.go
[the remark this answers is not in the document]
wpaulino: Just worried this will end up polluting the logs on shutdowns (it does).

log.go
halseth: Created https://github.com/lightningnetwork/lnd/issues/3662

query/interface.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/7526f709c18f132d28fa4e3b38b4b3cc7d390ed6

query/worker.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/c0f835f89d40a0250ad4094dec399364be94a201

query/worker.go
halseth: It should continue to send the response, added comment to make this clear: https://github.com/lightninglabs/neutrino/pull/179/commits/44a529ea3c830c15b5f5619709167102e9c31ea9

query/worker.go
halseth: Yes, very good catch! Probably a leftover from when it was used as part of the response. Fixed! https://github.com/lightninglabs/neutrino/pull/179/commits/20edcf05830ee6d4665f41bd98374df79eb8fb7a

query/worker.go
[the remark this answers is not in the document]
halseth: Removed the log: https://github.com/lightninglabs/neutrino/pull/179/commits/05cab790c90463db2e7de30b72579902d4ac9af9

query/workmanager.go
halseth: Awesome, yeah that's much nicer: https://github.com/lightninglabs/neutrino/pull/179/commits/43dd217aa8438064fe29febc707432253a8a7fcc

query/workmanager.go
halseth: I don't think it is! Removed: https://github.com/lightninglabs/neutrino/pull/179/commits/43dd217aa8438064fe29febc707432253a8a7fcc

query/workmanager.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/179/commits/36c12ad48585b21ba8e1c6fa73b9c6c29edf0aaa

neutrino.go
halseth: Idea was that it wasn't a problem blocking the peer handler, but one client stopping reading from its subscription will block the whole thing, which is obviously not good. Fixed by sending in a goroutine: https://github.com/lightninglabs/neutrino/pull/179/commits/1cdadade3d121aadf0ee999dbd515854f2eafd46

blockmanager.go
halseth: They are useful in the single message/multiple request case, such as `getCFFilters`, `getBlocks` etc.

blockmanager.go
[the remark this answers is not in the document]
halseth: Removed log on shutdown: https://github.com/lightninglabs/neutrino/pull/179/commits/1b582273b3327e22667279554d445076e1034aa9

blockmanager.go
Roasbeef: As in this seems to adhere to the interface, but using a private method to do so?

query/workmanager.go
```diff
@@ -0,0 +1,452 @@
...
+	queryIndex := uint64(0)
+	currentQueries := make(map[uint64]uint64)
+
+	workers := make(map[string]*activeWorker)
+
+Loop:
+	for {
+		// If the work queue is non-empty, we'll take out the first
+		// element in order to distribute it to a worker.
+		if work.Len() > 0 {
```
Roasbeef: One thing about this loop, is that when the job queue is non-empty, and all peers are handling a job, then it'll spin rather rapidly. I don't think this is blocking, but it can be resolved by waiting on a sort of semaphore channel `make(chan struct{}, len(peersConnected))`. This channel would be populated with a `struct{}{}` for each peer to start with, each time we hand a job over we remove an element, and each time we get a response we re-add an element. 

The end result is that we can `select` at the top using this channel, avoiding a tight spin loop when all the peers are saturated. So something like: 
```go
if work.Len() > 0 {
     // Wait for an open slot to become available. 
    select {
        // Free lost, can hand out work. 
        case <-workSema:
        
        case <-quit:
    }
}
```

We re-add to this channel each time we receive a response. Each time the peer set changes, we re-size the channel (re-create) to match the total number of peers.

halseth: this shouldn't be the case, as `freeWorkers` will be empty, and we'll immediately move down to the below select waiting for a result or a new peer.

Agreed that this could be made more clear, but it will change a bit in the follow up adding more slots to each worker, so will revisit then.

blockmanager.go
wpaulino: A single `getcfilter` message will only ever return one response though. In the case of a batch filter fetch (multiple `getcfilter` messages) I would understand the purpose of these fields if a single worker was handling them, but that doesn't seem to be the case (the messages can be distributed among distinct workers).

halseth: a `getCFilters` message specifies a start and stop height (https://github.com/btcsuite/btcd/blob/a41498d578a99c1619037746abd916176ea61052/wire/msggetcfilters.go#L20), and each height is returned as a single `MsgCFilter` (https://github.com/btcsuite/btcd/blob/a41498d578a99c1619037746abd916176ea61052/server.go#L819).

wpaulino: Ah I see, thanks for clearing that up. In any case, `Progressed` and `Finished` still seem mutually exclusive.

PR #187, opened 2019-10-11
Return error on failed CFilter queries

halseth:
Previously we would return a nil filter with nil error, which would mask
a failed filter fetch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
   - mkdir -p $GOPATH/src/github.com/btcsuite/
   - "[ -d $GOPATH/src/github.com/btcsuite/btcd ] || git clone https://github.com/btcsuite/btcd.git $GOPATH/src/github.com/btcsuite/btcd"
   - pushd $GOPATH/src/github.com/btcsuite/btcd
+  - git checkout .
   - $GOPATH/src/github.com/lightninglabs/neutrino/btcd_checkout.sh
   - go install . ./cmd/...
   - popd
--- a/go.sum
+++ b/go.sum
@@ -40,6 +40,7 @@ github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/jessevdk/go-flags v0.0.0-20141203071132-1679536dcc89 h1:12K8AlpT0/6QUXSfV0yi4Q0jkbq8NDtIKFtF61AoqV0=
 github.com/jessevdk/go-flags v0.0.0-20141203071132-1679536dcc89/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
 github.com/jrick/logrotate v1.0.0 h1:lQ1bL/n9mBNeIXoTUoYRlK4dHuNJVofX9oWqBtPnSzI=
 github.com/jrick/logrotate v1.0.0/go.mod h1:LNinyqDIJnpAur+b8yyulnQw/wDuN1+BYKlTRt3OuAQ=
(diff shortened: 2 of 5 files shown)


go.sum
```diff
@@ -40,6 +40,7 @@ github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/jessevdk/go-flags v0.0.0-20141203071132-1679536dcc89 h1:12K8AlpT0/6QUXSfV0yi4Q0jkbq8NDtIKFtF61AoqV0=
```
wpaulino: I feel like we've addressed this many times before and it seems to keep coming back?

query.go
```diff
@@ -1243,6 +1243,11 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 			if !ok {
 				// Query has finished, if we have a result we'll
 				// return it.
+				if resultFilter == nil {
+					return nil, fmt.Errorf("unable to " +
```
wpaulino: Would it be useful to extract the error into a variable?

query.go
```diff
@@ -1251,11 +1256,9 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 			resultFilter = filter
 
 		case <-s.quit:
-			// TODO(halseth): return error?
-			return nil, nil
+			return nil, fmt.Errorf("shutting down")
```
wpaulino: Nit: can use `ErrShuttingDown`.

rescan.go
```diff
@@ -562,22 +562,13 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
-		// If the filter is nil, then this either means that we don't
-		// have any peers to fetch this filter from, or the peer(s) that
-		// we're trying to fetch from are in the progress of a re-org.
-		if blockFilter == nil {
+		if err != nil {
+			// If the query failed, then this either means that we
+			// don't have any peers to fetch this filter from, or
+			// the peer(s) that we're trying to fetch from are in
+			// the progress of a re-org.
+			log.Errorf("unable to get filter for hash=%v, "+
```
wpaulino: Perhaps just log the error if it fails with something other than `ErrFilterNotFound`? Can imagine it getting spammy in some cases.

.travis.yml
```diff
@@ -14,6 +14,7 @@ install:
   - mkdir -p $GOPATH/src/github.com/btcsuite/
   - "[ -d $GOPATH/src/github.com/btcsuite/btcd ] || git clone https://github.com/btcsuite/btcd.git $GOPATH/src/github.com/btcsuite/btcd"
   - pushd $GOPATH/src/github.com/btcsuite/btcd
+  - git checkout .
```
wpaulino: This should be addressed if we pin to the latest btcd version.

rescan.go
halseth: You mean in cases where we fail to fetch the filter because we have no peers/timeout etc? Shouldn't be a very expected error though?

.travis.yml
halseth: Partly addressed in #187. Should be fully addressed by adding the tag to the `btcd` go.mod

query.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/187/commits/8c1f6cf2a794a7259b097602469cf0f01d2d49bb

query.go
halseth: Fixed: https://github.com/lightninglabs/neutrino/pull/187/commits/8c1f6cf2a794a7259b097602469cf0f01d2d49bb

go.sum
halseth: yeah, not sure why, but in this case it looks correct since there is only one entry.

rescan.go
wpaulino: Just thinking it's more likely in a mobile scenario where users are switching networks, losing reception, etc.

halseth: Mhm, i think it is worth logging though

.travis.yml
halseth: Should be done in its own PR when we update the dependency

wpaulino: Why can't we just update the dependency now and get rid of this commit?

rescan.go
wpaulino: Nit: logged error should start capitalized.

.travis.yml
Roasbeef: Will update the dep after btcd's `v0.20.0` tag, then we can both fetch that and pin against that in our `go.mod`.

PR #188, opened 2019-10-11
[travis] fix btcd checkout script

halseth:


wpaulino:
Is this still needed?

halseth:
Yes, the current script attempts to find the btcd version in the go.sum file inside the btcd directory. If you inspect the logs you can see that the version is not actually being checked out.

btcd_checkout.sh
```diff
@@ -1,3 +1,12 @@
...
-git pull && git checkout $(cat go.sum | grep -A1 github.com/btcsuite/btcd | tail -n1 | awk -F " " '{ print $2 }' | awk -F "-" '{ print $3 }' | awk -F "/" '{ print $1 }')
+BTCD_COMMIT=$(cat go.mod | \
+        grep github.com/btcsuite/btcd | \
+        tail -n1 | \
+        awk -F " " '{ print $2 }' | \
+        awk -F "/" '{ print $1 }')
+echo "Fetching btcd at $BTCD_COMMIT"
+
+pushd /tmp
+GO111MODULE=on go get -v github.com/btcsuite/btcd@$BTCD_COMMIT
```
wpaulino: I wonder if this is actually installing `btcd` at this commit? Given that this is being executed in a directory that supports modules, something tells me that this is being picked up by modules  due to `GO111MODULE=on` (I've seen issues with people trying to install `lnd` this way). I'd rather just use good ol' `git clone`.

halseth: it is  being run from `/tmp` dir to avoid this.

wpaulino: Haha missed that!

PR #190, opened 2019-10-29
neutrino: don't connect to peer which deny us relay

Roasbeef:
In this commit, we fix an existing bug that would allow us to connect to
the peer that explicitly signals they don't accept transaction relay.
When we send our version message we explicitly set the `DisableTxRelay`
bool to `false`, but then fail to check this value for any remote peers.
With this change, we'll disconnect a peer in the `OnVersion` call back
if they signal no transaction relay.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -255,6 +255,17 @@ func (sp *ServerPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
 	// the local clock to keep the network time in sync.
 	sp.server.timeSource.AddTimeSample(sp.Addr(), msg.Timestamp)
 
+	// If the peer doesn't allow us to relay any transactions to them, then
+	// we won't add them as a peer, as they aren't of much use to us.
+	if msg.DisableRelayTx {
+		log.Debugf("%v does not allow transaction relay, disconecting",
+			sp)
+
+		sp.Disconnect()
+
+		return nil
+	}
+
 	// Check to see if the peer supports the latest protocol version and
 	// service bits required to service us. If not, then we'll disconnect
 	// so we can find compatible peers.


neutrino.go
```diff
@@ -255,6 +255,17 @@ func (sp *ServerPeer) OnVersion(_ *peer.Peer, msg *wire.MsgVersion) *wire.MsgRej
 	// the local clock to keep the network time in sync.
 	sp.server.timeSource.AddTimeSample(sp.Addr(), msg.Timestamp)
 
+	// If the peer doesn't allow us to relay any transactions to them, then
+	// we won't add them as a peer, as they aren't of much use to us.
+	if msg.DisableRelayTx {
+		log.Debugf("%v does not allow transaction relay, disconecting",
+			sp)
+
+		sp.Disconnect()
```
wpaulino: Should we also ban them to prevent connecting to them again?

Sjors:
A node without transaction relay could still be providing filters. That's how I found this :-)

wpaulino:
We could opt to allow the connection and track which peers we can relay transactions to, though we'd want to make sure that we have enough peers so that transactions are actually relayed when the time comes.

neutrino.go
Roasbeef: IMO no, as this can change for them with just a restart. In the end we aren't fundamentally incompatible. I think we also saw this issue in the past at times when it seemed like things weren't being broadcast, it may have been the case that we were connected to a node running in blocks only mode.

Roasbeef:
> We could opt to allow the connection and track which peers we can relay transactions to

Possibly, but IMO in order to ensure out transactions are well propagated, we should only be connected to peers that can relay our transaction. The current transaction broadcast package also has this assumption baked in.

neutrino.go
wpaulino: Can't the same be said about filter support?

Any reason to not return a `Reject` message with the reason?

halseth: By just disconnecting without banning, don't we risk going into a connection loop? How likely are we to attempt connecting to this peer again?

wpaulino: Seems likely to happen if the address manager continues to return it as part of `newAddressFunc`, which is very aggressive at the moment.

Roasbeef:
!lightninglabs-deploy mute

PR #191, opened 2019-10-31
neutrino: don't request new connections once max peers is reached

wpaulino:
Doing so would get us in a loop where we continue to request connections but we end up disconnecting them due to already having our maximum number of peers.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -1244,6 +1244,8 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
 // handleDonePeerMsg deals with peers that have signalled they are done.  It is
 // invoked from the peerHandler goroutine.
 func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
+	// If the peer is being tracked internally, i.e., we received their
+	// VerAck, we'll need to remove them.
 	var list map[int32]*ServerPeer
 	if sp.persistent {
 		list = state.persistentPeers
@@ -1253,17 +1255,24 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
 	if _, ok := list[sp.ID()]; ok {
 		state.outboundGroups[addrmgr.GroupKey(sp.NA())]--
 		delete(list, sp.ID())
+
+		log.Debugf("Removed peer %s", sp)
+
+		// Only request a new connection if the peer being disconnected
+		// is not persistent. There's no need to do so if the peer is
+		// persistent since the connection manager will attempt to
+		// reconnect.
 		if sp.persistent {
 			s.connManager.Disconnect(sp.connReq.ID())
 		} else {
 			s.connManager.Remove(sp.connReq.ID())
 			go s.connManager.NewConnReq()
 		}
-		log.Debugf("Removed peer %s", sp)
 		return
 	}
 
-	// We'll always remove peers that are not persistent.
+	// We'll always request a new connection for peers that have not
+	// completed the version handshake for whatever reason.
 	s.connManager.Remove(sp.connReq.ID())
 	go s.connManager.NewConnReq()
 }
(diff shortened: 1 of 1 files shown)


neutrino.go
```diff
@@ -1251,16 +1251,14 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
 		list = state.outboundPeers
 	}
 	if _, ok := list[sp.ID()]; ok {
-		if !sp.Inbound() && sp.VersionKnown() {
```
halseth: where do we ensure no peer can make inbound connections to us?

neutrino.go
```diff
@@ -1251,16 +1251,14 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
 		list = state.outboundPeers
 	}
 	if _, ok := list[sp.ID()]; ok {
-		if sp.VersionKnown() {
-			state.outboundGroups[addrmgr.GroupKey(sp.NA())]--
-		}
+		state.outboundGroups[addrmgr.GroupKey(sp.NA())]--
+		delete(list, sp.ID())
```
halseth: this method really needs some comments

neutrino.go
```diff
@@ -1257,15 +1257,17 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
...
+		// We'll always remove peers that are not persistent.
+		s.connManager.Remove(sp.connReq.ID())
 	}
 
-	// We'll always remove peers that are not persistent.
-	s.connManager.Remove(sp.connReq.ID())
-	go s.connManager.NewConnReq()
+	// Request a new connection if we still need more.
+	if state.Count() < MaxPeers {
+		go s.connManager.NewConnReq()
```
halseth: this changes the logic slightly, by also requesting new connections after disconnecting a permanent peer. Which I guess is correct..? Why wasn't this done before?

neutrino.go
Roasbeef: I don't think we ever listen using the `connmgr`?

neutrino.go
Roasbeef: Any fundamental reason to move the `delete` call here? Just grouping of related map operations I guess?

neutrino.go
```diff
@@ -1257,15 +1257,17 @@ func (s *ChainService) handleDonePeerMsg(state *peerState, sp *ServerPeer) {
 			s.connManager.Disconnect(sp.connReq.ID())
 		} else {
 			s.connManager.Remove(sp.connReq.ID())
-			go s.connManager.NewConnReq()
 		}
 		log.Debugf("Removed peer %s", sp)
-		return
+	} else {
+		// We'll always remove peers that are not persistent.
```
Roasbeef: Wasn't this always already hit above (outbound non-persistent). This is for peers that never completed the handshake, so were never added to our internal peer tracking state.

neutrino.go
Roasbeef: Much of the original code was lifted from `btcd`, which didn't care as much about aggressively maintaining the capped number of outbound peers. We should really start to consolidate some of this code though, as of recent anytime we make a change to this area of the codebase, then typically make a mirrored version for `btcd`.

neutrino.go
wpaulino: > I don't think we ever listen using the connmgr?

Correct. There's no logic to even handle inbound connections at the moment.

neutrino.go
wpaulino: > Just grouping of related map operations I guess?

Yeah.

> this method really needs some comments

Done.

neutrino.go
wpaulino: > Which I guess is correct..?

It's not correct, this is actually a regression. Fixed in the latest version. We don't request for permanent peers because those will be reconnected automatically by the connection manager.

neutrino.go
wpaulino: Correct. I've unified the handling of peers being tracked and not being tracked to hopefully make this easier to follow.

PR #192, opened 2019-11-16
utxoscanner: reuse options where possible

Crypt-iQ:
On startup, noticed that if there are a lot of `filterEntries` leads to high mem usage, they all get copied here:
```
options = &rescanOptions{
     watchList: reporter.filterEntries,
}
```
and they continually get copied until a match is found. With pprof, noticed about 38MB, sometimes more depending on the `channel.db` I was using:
```
   38.01MB    38.01MB    321:			options := rescanOptions{
         .          .    322:				watchList: reporter.filterEntries,
         .          .    323:			}
```

With this change, pprof doesn't even log the memory since it's so low. :) 

EDIT: I know that with batching filter requests, this probably isn't necessary, but for now it's a nice memory improvement.

### DIFF
--- a/utxoscanner.go
+++ b/utxoscanner.go
@@ -297,6 +297,7 @@ func (s *UtxoScanner) scanFromHeight(initHeight uint32) error {
 	)
 
 	reporter := newBatchSpendReporter()
+	options := defaultRescanOptions()
 
 scanToEnd:
 	// Scan forward through the blockchain and look for any transactions that
@@ -324,11 +325,9 @@ scanToEnd:
 		// outpoints.
 		fetch := len(newReqs) > 0
 		if !fetch {
-			options := rescanOptions{
-				watchList: reporter.filterEntries,
-			}
+			options.watchList = reporter.filterEntries
 
-			match, err := s.cfg.BlockFilterMatches(&options, hash)
+			match, err := s.cfg.BlockFilterMatches(options, hash)
 			if err != nil {
 				return reporter.FailRemaining(err)
 			}


utxoscanner.go
```diff
@@ -295,6 +295,7 @@ func (s *UtxoScanner) scanFromHeight(initHeight uint32) error {
 scanToEnd:
 	// Scan forward through the blockchain and look for any transactions that
 	// might spend the given UTXOs.
+	var options *rescanOptions
```
halseth: nice catch! Could we even initiate the options here directly, avoiding the nil check in the loop?

Crypt-iQ: I don't think so bc the options can change during the loop if there are any new requests according to `newReqs := s.dequeueAtHeight(height)`

halseth: Ah. I missed the nilling of the `options` below.

I'm curious though, why does the current version hog up memory? When the previous iteration's `options` go out of scope the GC should be able to free it.

Crypt-iQ: The GC should be able to free it when it runs, but I think this PR could reduce heap pressure. WDYT? Memory hogging is just from repeatedly copying the options over and over

halseth: I see, it is the copying on each iteration that hogs memory faster than the gc can free it probably.

Could an alternative solution be to just pass a reference to the watch list instead of copying it?

halseth:
Ping @Crypt-iQ

Crypt-iQ:
@halseth Have not looked at this codebase in quite some time so would need to refresh my memory.  I think tackling https://github.com/lightninglabs/neutrino/issues/196 could be a better use of time as this PR may only cut down ~50MB whereas the other is on the order of GB. Though if we use checkpoints / update them then the memory usage should drop

Roasbeef:
@creslinux they're additive though right? And this PR is just a few lines, so seems like an easy win to me...

halseth:
> @creslinux they're additive though right? And this PR is just a few lines, so seems like an easy win to me...

@Crypt-iQ ?

Crypt-iQ:
Ok will fix up

utxoscanner.go
Crypt-iQ: I don't know if refactoring `rescanOptions.watchList` to `*[][]byte` is worth it for this optimization as there would be a lot of dereferencing in `rescan.go`

Crypt-iQ: Remembered that slices are pointers and don't copy the underlying data, the allocations should be from creating the `rescanOptions` struct so many times.

Roasbeef:
What needs to be done here? The change looks pretty minimal.

halseth:
I think we need to understand what is actually hogging memory here.

Crypt-iQ:
PTAL @halseth @carlaKC 

turns out the memory-hogging was due to `rescanOptions` being allocated every iteration.  If we initialize once and then copy the `filterEntries` slice, all is well and no memory is logged with pprof.  We have to set/copy `watchList` to `filterEntries` because `filterEntries` is appended to, which creates a new slice.

halseth (approved):
Nice, this looks a good 👍

carlaKC (approved):
LGTM!

PR #194, opened 2019-11-24
Add PersistToDisk property to ChainService

### LINKED ISSUE #193: PersistToDisk inconsistency with FilterCache in GetCFilter
Currently, PersistToDisk is an defined and used as an option to GetCFilter. So each caller to GetCFilter can decide if she want the result to be stored persistently in the disk.
The problem is that when the in memory FilterCache is used (the default) and the filter is already in the cache because of a first call without PersistToDisk option, the call to GetCFilter doesn't write persistently to the disk even when the option is set.

One way to solve the problem, is to add a flag to the cached filter (the CacheableFilter struct) in order to know if the filter was written to disk. Then if GetCFilter is called with PersistToDisk option and finds a specific filter in the memory cache, it will check this new flag to decide if it needs to write the filter to the disk.

Another way (which I tend to prefer) is to remove the PersistToDisk from the query options and pass the flag to NewChainService struct, the same way as the FilterCacheSize is passed to it.
This means that the PersisToDisk property will be handled by the ChainService instance and not by each GetCFilter call.

yaslama:
Fix #193 
Currently PersisToDisk is an option to GetCFilter.
The problem is that the option is not working well with the in-memory
FilterCache. The biggest problem is that if there are two calls of
GetCFilter, the first one without the PersistToDisk option, and the second
one with the PersistToDisk option, the second one returns the filter from
the cache and doesn't write it persistently to the disk.
Consequently, if a filter was already queried, there is no way to write it
to disk. So for instance if a neutrino user (mainly lnd) is closed before
it is fully synchronized, it need to re-download all the filters when it
runs again.
When handling the PersistToDisk in the ChainService itself, the caller of
NewChainService decides in _one place_ if the filters need to be stored
in the disk, and that will not depend on the order of calls to
GetCFilter.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -570,6 +570,12 @@ type Config struct {
 	// cache will hold in memory at most.
 	BlockCacheSize uint64
 
+	// persistToDisk indicates whether the filter should also be written
+	// to disk in addition to the memory cache. For "normal" wallets, they'll
+	// almost never need to re-match a filter once it's been fetched unless
+	// they're doing something like a key import.
+	PersistToDisk bool
+
 	// AssertFilterHeader is an optional field that allows the creator of
 	// the ChainService to ensure that if any chain data exists, it's
 	// compliant with the expected filter header state. If neutrino starts
@@ -597,6 +603,7 @@ type ChainService struct {
 	FilterDB         filterdb.FilterDatabase
 	BlockHeaders     headerfs.BlockHeaderStore
 	RegFilterHeaders *headerfs.FilterHeaderStore
+	persistToDisk    bool
 
 	FilterCache *lru.Cache
 	BlockCache  *lru.Cache
@@ -689,6 +696,7 @@ func NewChainService(cfg Config) (*ChainService, error) {
 		userAgentVersion:  UserAgentVersion,
 		nameResolver:      nameResolver,
 		dialer:            dialer,
+		persistToDisk:     cfg.PersistToDisk,
 	}
 	s.workManager = query.New(&query.Config{
 		ConnectedPeers: s.ConnectedPeers,
--- a/query.go
+++ b/query.go
@@ -78,12 +78,6 @@ type queryOptions struct {
 	// it's run in a goroutine.
 	doneChan chan<- struct{}
 
-	// persistToDisk indicates whether the filter should also be written
-	// to disk in addition to the memory cache. For "normal" wallets, they'll
-	// almost never need to re-match a filter once it's been fetched unless
-	// they're doing something like a key import.
-	persistToDisk bool
-
 	// optimisticBatch indicates whether we expect more calls to follow,
 	// and that we should attempt to batch more items with the query such
 	// that they can be cached, avoiding the extra round trip.
@@ -170,14 +164,6 @@ func Encoding(encoding wire.MessageEncoding) QueryOption {
 func DoneChan(doneChan chan<- struct{}) QueryOption {
 	return func(qo *queryOptions) {
 		qo.doneChan = doneChan
-	}
-}
-
-// PersistToDisk allows the caller to tell that the filter should be kept
-// on disk once it's found.
-func PersistToDisk() QueryOption {
-	return func(qo *queryOptions) {
-		qo.persistToDisk = true
 	}
 }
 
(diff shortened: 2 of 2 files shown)


halseth:
What you outline above sounds like a bug? If it is found in the cache, but is not yet on disk, maybe it should be written? Or would that lead to performance issues?

yaslama:
Thanks @halseth for your comments. When found in the cache, there are two ways to check if it's on disk: 1) to make a read to disk, which means that when persisttodisk is used, a disk access is done for every call. 2) to add a "writtentodisk" flag in the cache for every filter.
In addition both possibilities are not aligned to the way the memory cache works because the memory cache is handled globally (for the ChainService)

halseth:
OK, makes sense.

A way to fix this without significant added delay would be to spin up a goroutine that could write filters to disk lazily. 

I also like the 2) you outline, by adding a flag in the cache that indicates whether the filter was written. That would be a pretty simple change?

yaslama:
I agree that, technically, the two solutions you propose are good way to fix the issue. The problem is the semantic of using the persistToDisk for some fetch calls and not using it for other calls. What is the use case of having only some filters to disk (with holes)?

halseth:
The use case is that you are given more control of what filters to persist. For instance if you are checking a historical spend that happened a long time ago, it might not make sense to persist those old filters, as you are likely to never read them again. 

If we were to flip a switch to persist all filters, we would also need to have some eviction strategy I think, as a rescan might cause the whole chain to be scanned and a big storage cost.

yaslama:
In fact the main scenario I see for persisting the filter is when doing a long rescan and the app stops in the middle (for instance when running the app in a mobile). I agree that if it's a switch, I need to add a way to evict filters after rescans.

halseth:
I see. Maybe we can check whether we could make persisted rescan "checkpoints", such that we could only store the filters between those checkpoints?

piecegift:
@yaslama Thank you for the fix!

I wrote a package to simplify watching for addresses. I am interested only in deposits, but it can be used for withdrawals as well

https://github.com/piecegift/watch

I used your PR, because PersistToDisk from master didn't work for me: when I run the first time, I found transactions, but the second time I was getting `unable to fetch cfilter`.

But even with your PR I can not work with blocks that are before the block I used as starting block in the first run with this directory. Also if I don't run any rescan at all in the first run of the tool then in subsequent runs it produces the same error.

Let me show:

```
# Build
$ git clone github.com/piecegift/watch
$ cd ./watch
$ GO111MODULE=on go get ./watchtool

$ mkdir /tmp/tmp-watchtool
$ watchtool -dir /tmp/tmp-watchtool -start-block 613000 -tor-socks '' -address 176SsYVUoBcp3kzaBob7b6ikT5MKC16Rz1
# Works fine.

# Try again - still works.
$ watchtool -dir /tmp/tmp-watchtool -start-block 613000 -tor-socks '' -address 176SsYVUoBcp3kzaBob7b6ikT5MKC16Rz1

# Decrease start-block.
$ watchtool -dir /tmp/tmp-watchtool -start-block 612000 -tor-socks '' -address 176SsYVUoBcp3kzaBob7b6ikT5MKC16Rz1
Rescan error: unable to fetch cfilter.

# Also if you first run without rescans, and then with rescan, it fails.
$ rm -r /tmp/tmp-watchtool
$ watchtool -dir /tmp/tmp-watchtool -start-block 613000 -tor-socks ''
$ watchtool -dir /tmp/tmp-watchtool -start-block 613000 -tor-socks '' -address 176SsYVUoBcp3kzaBob7b6ikT5MKC16Rz1
Rescan error: unable to fetch cfilter.
```

The tool is already usable in my use case, but there is a bug somewhere: in my code, in neutrino or in your PR.

It looks like neutrino remembers somewhere first cfilter that was downloaded (or blockchain tip if there was no rescan) and then refuses to load anything older than it from network. I spent few days because of this behavior. I think it should be more stateless and straightforward: if the user specifies the range of blocks, filters should be loaded regardless of the directory state. If they are not stored locally, then just load from network again.

Roasbeef:
Any updates on this? Has the bug reported by `piecegift` above been resolved?

yaslama:
> Any updates on this? Has the bug reported by `piecegift` above been resolved?

I don't think that the bug reported by @piecegift is related to this PR.

piecegift:
Any idea what is the root cause of the bug I reported?

Roasbeef:
> Maybe we can check whether we could make persisted rescan "checkpoints", such that we could only store the filters between those checkpoints?

This seems to be implemented already, but not yet used in `lnd`, so perhaps we just need to have the feature find its way into lnd?

Roasbeef:
@piecegift can you make a _new_ issue for the trouble you're running into, thanks!

Roasbeef (approved):
LGTM 🎥

halseth (approved):

halseth:
@yaslama Is a PR to LND needed to make use of this?

PR #198, opened 2020-01-16
neutrino: support utxo scanner progress.

roeierez:
This PR suggests to add progress notifications to the utxo scanning process.
Currently the caller only gets feedback at the end of the process where it can be sometimes very long.
Besides the feedback there is a need to be able to start the scanning from the last point it has stopped. This is achieved in LND, for example, by using a "SpendHint" cache but it only updates the cache at the end of the process which creates a problem for mobile users that don't open their app for several weeks. The utxo scanning at this case is likely to be longer than the time their app will be online resulting in scanning all over again on every restart.
This addition will enable LND to update the hint cache during the scanning progress resulting in better incremental rescan between restarts.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -76,11 +76,16 @@ type ChainSource interface {
 	Subscribe(bestHeight uint32) (*blockntfns.Subscription, error)
 }
 
+// ScanProgressHandler is used in rescanOptions to update the caller with the
+// rescan progress.
+type ScanProgressHandler func(lastProcessedBlock uint32)
+
 // rescanOptions holds the set of functional parameters for Rescan.
 type rescanOptions struct {
 	queryOptions []QueryOption
 
-	ntfn rpcclient.NotificationHandlers
+	ntfn            rpcclient.NotificationHandlers
+	progressHandler ScanProgressHandler
 
 	startTime  time.Time
 	startBlock *headerfs.BlockStamp
@@ -117,6 +122,18 @@ func QueryOptions(options ...QueryOption) RescanOption {
 func NotificationHandlers(ntfn rpcclient.NotificationHandlers) RescanOption {
 	return func(ro *rescanOptions) {
 		ro.ntfn = ntfn
+	}
+}
+
+// ProgressHandler specifies a handler to be used when the utxo
+// scanner reports its progress.
+// The passed handler should be non-blocking for the rescan to continue
+// normally.
+func ProgressHandler(
+	handler ScanProgressHandler) RescanOption {
+
+	return func(ro *rescanOptions) {
+		ro.progressHandler = handler
 	}
 }
 
@@ -1430,7 +1447,7 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
 	}
 
 	req, err := s.utxoScanner.Enqueue(
-		&ro.watchInputs[0], uint32(ro.startBlock.Height),
+		&ro.watchInputs[0], uint32(ro.startBlock.Height), ro.progressHandler,
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


Roasbeef:
Is there meant to be a PR to `lnd` that exposes this information? Otherwise as is, it's dead code.

Roasbeef:
> This addition will enable LND to update the hint cache during the scanning progress resulting in better incremental rescan between restarts.

Will this be made as a follow up PR?

roeierez:
> Will this be made as a follow up PR?

Yes, I actually have it almost ready.

rescan.go
```diff
@@ -76,11 +76,14 @@ type ChainSource interface {
...
 
+type scanProgressHandler func(lastProcessedBlock uint32)
+
 // rescanOptions holds the set of functional parameters for Rescan.
 type rescanOptions struct {
 	queryOptions []QueryOption
 
-	ntfn rpcclient.NotificationHandlers
+	ntfn                       rpcclient.NotificationHandlers
+	utxoScannerProgressHandler scanProgressHandler
```
halseth: naming suggesting: just `progressHandler`?

rescan.go
```diff
@@ -76,11 +76,14 @@ type ChainSource interface {
 	Subscribe(bestHeight uint32) (*blockntfns.Subscription, error)
 }
 
+type scanProgressHandler func(lastProcessedBlock uint32)
```
halseth: add a comment to this type

rescan.go
```diff
@@ -120,6 +123,16 @@ func NotificationHandlers(ntfn rpcclient.NotificationHandlers) RescanOption {
 	}
 }
 
+// UtxoScannerProgressHandler specified a handler to be used when the utxo
+// scanner reports its progress.
```
halseth: should add a note that this should be non-blocking, otherwise it might stall progress.

rescan.go
```diff
@@ -1430,7 +1443,7 @@ func (s *ChainService) GetUtxo(options ...RescanOption) (*SpendReport, error) {
 	}
 
 	req, err := s.utxoScanner.Enqueue(
-		&ro.watchInputs[0], uint32(ro.startBlock.Height),
+		&ro.watchInputs[0], uint32(ro.startBlock.Height), ro.utxoScannerProgressHandler,
```
halseth: long line.

utxoscanner.go
```diff
@@ -35,6 +35,10 @@ type GetUtxoRequest struct {
 	// request.
 	result *getUtxoResult
 
+	// onProgress is the method to be used by the scanner to report
+	// its progress.
+	onProgress scanProgressHandler
```
halseth: add note that can be nil if not specified by the caller.

rescan.go
roeierez: Certainly!

roeierez:
> Looks good! Last missing piece is to add/modify existing unit test to check that the progess callback works as expected 👍

Tests updated.

rescan.go
```diff
@@ -120,6 +125,18 @@ func NotificationHandlers(ntfn rpcclient.NotificationHandlers) RescanOption {
 	}
 }
 
+// ProgressHandler specified a handler to be used when the utxo
+// scanner reports its progress.
+// The passed handler should be non-blocking for the rescan to continue
+// normally.
+func ProgressHandler(
+	handler scanProgressHandler) RescanOption {
```
wpaulino, line 133: Since `ProgressHandler` is exported, we should also export `scanProgressHandler`.

rescan.go
```diff
@@ -120,6 +125,18 @@ func NotificationHandlers(ntfn rpcclient.NotificationHandlers) RescanOption {
 	}
 }
 
+// ProgressHandler specified a handler to be used when the utxo
```
wpaulino, line 128:
```suggestion
// ProgressHandler specifies a handler to be used when the utxo
```

rescan.go
roeierez: Nice catch! fixed.

PR #199, opened 2020-01-23
filterdb: Implement PurgeFilters.

roeierez:
This PR adds the ability to purge the filters from the db. Neutrino has the `PersistToDisk` query option which is useful for background sync in mobile applications. These apps can set a background job to fetch the filters periodically, preparing the data for the app when it starts by the user, which achieves a fast sync time and immediate readiness for payment.
This change adds a way to purge these filters and enables the app to implement its own eviction strategy preventing the db from growing boundlessly.

### DIFF
--- a/filterdb/db.go
+++ b/filterdb/db.go
@@ -51,6 +51,9 @@ type FilterDatabase interface {
 	// target block hash cannot be found, then ErrFilterNotFound is to be
 	// returned.
 	FetchFilter(*chainhash.Hash, FilterType) (*gcs.Filter, error)
+
+	// PurgeFilters purge all filters with a given type from persistent storage.
+	PurgeFilters(FilterType) error
 }
 
 // FilterStore is an implementation of the FilterDatabase interface which is
@@ -105,6 +108,29 @@ func New(db walletdb.DB, params chaincfg.Params) (*FilterStore, error) {
 	return &FilterStore{
 		db: db,
 	}, nil
+}
+
+// PurgeFilters purge all filters with a given type from persistent storage.
+//
+// NOTE: This method is a part of the FilterDatabase interface.
+func (f *FilterStore) PurgeFilters(fType FilterType) error {
+	return walletdb.Update(f.db, func(tx walletdb.ReadWriteTx) error {
+		filters := tx.ReadWriteBucket(filterBucket)
+
+		switch fType {
+		case RegularFilter:
+			if err := filters.DeleteNestedBucket(regBucket); err != nil {
+				return err
+			}
+			if _, err := filters.CreateBucket(regBucket); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown filter type: %v", fType)
+		}
+
+		return nil
+	})
 }
 
 // putFilter stores a filter in the database according to the corresponding


Crypt-iQ:
I like this change, but do you notice a change in the size of the db when running this? `bbolt` doesn't shrink when deleting things (keeps the freed pages on disk still). May be useful if/when neutrino migrates to an abstracted k/v store that supports things besides `bbolt`.

roeierez:
The bolt db file size indeed doesn't change but with this change the app layer at least can make sure it's eviction strategy prevents the db from growing too much.
As far as I know the only way to shrink the db is to run compact and this is also an option (that should be done with care) for applications that already have a big db file.

Crypt-iQ:
> The bolt db file size indeed doesn't change but with this change the app layer at least can make sure it's eviction strategy prevents the db from growing too much.
> As far as I know the only way to shrink the db is to run compact and this is also an option (that should be done with care) for applications that already have a big db file.

Ah I see, I'll take a look at the code :)

filterdb/db.go
```diff
@@ -51,6 +51,9 @@ type FilterDatabase interface {
 	// target block hash cannot be found, then ErrFilterNotFound is to be
 	// returned.
 	FetchFilter(*chainhash.Hash, FilterType) (*gcs.Filter, error)
+
+	// PurgeFilters purge all filters with a given type from persistent storage
+	PurgeFilters(FilterType) error
```
halseth: dead code?

roeierez: Not used within neutrino but intended to be used by apps that use neutrino as a chain service.
For example, we use a background job that run neutrino  to persist the filters and this gives us a way to purge the filters when not needed.

filterdb/db.go
```diff
@@ -51,6 +51,9 @@ type FilterDatabase interface {
 	// target block hash cannot be found, then ErrFilterNotFound is to be
 	// returned.
 	FetchFilter(*chainhash.Hash, FilterType) (*gcs.Filter, error)
+
+	// PurgeFilters purge all filters with a given type from persistent storage
```
halseth: nit: missing period

filterdb/db.go
```diff
@@ -107,6 +110,28 @@ func New(db walletdb.DB, params chaincfg.Params) (*FilterStore, error) {
 	}, nil
 }
 
+// PurgeFilters purge all filters with a given type from persistent storage
+// NOTE: This method is a part of the FilterDatabase interface.
```
halseth: style: add empty comment line above `NOTE`

filterdb/db.go
```diff
@@ -107,6 +110,29 @@ func New(db walletdb.DB, params chaincfg.Params) (*FilterStore, error) {
...
+				return err
+			}
+			if _, err := filters.CreateBucket(regBucket); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown filter type: %v", fType)
+		}
+
+		return nil
```
Crypt-iQ: This statement can't be reached, any way to coalesce this with the above default case?

halseth: Looks okay, as it will fall through after succesfully deleting the bucket for the filter type above

Crypt-iQ: Ah you're right

PR #203, opened 2020-03-09
Add MaxBatchSize query option.

roeierez:
This PR adds a query option `MaxBatchSize` that allows the caller to limit the batch size requested when using either `OptimisticBatch` or `OptimisticReverseBatch`.
This option is useful in cases where the caller needs a limited amount of items (such as scanning backwards to a certain point) and wants to allow neutrino save bandwidth by "hinting" about the max items it needs.
The term "maxBatchSize" is used rather than "batchSize" since the caller is not exposed to the max number of items neutrino used as a default and is only allowed to limit this number.

This change is the first step towards fixing https://github.com/lightningnetwork/lnd/issues/4060

### DIFF
--- a/query.go
+++ b/query.go
@@ -88,6 +88,12 @@ type queryOptions struct {
 	// and that we should attempt to batch more items with the query such
 	// that they can be cached, avoiding the extra round trip.
 	optimisticBatch optimisticBatchType
+
+	// maxBatchSize is the maximum items that the query should return in the
+	// case the optimisticBatch option is used. It saves bandwidth in the case
+	// the caller has a limited amount of items to fetch but still wants to use
+	// batching.
+	maxBatchSize int64
 }
 
 // optimisticBatchType is a type indicating the kind of batching we want to
@@ -194,6 +200,14 @@ func OptimisticBatch() QueryOption {
 func OptimisticReverseBatch() QueryOption {
 	return func(qo *queryOptions) {
 		qo.optimisticBatch = reverseBatch
+	}
+}
+
+// MaxBatchSize allows the caller to limit the number of items fetched
+// in a batch.
+func MaxBatchSize(maxSize int64) QueryOption {
+	return func(qo *queryOptions) {
+		qo.maxBatchSize = maxSize
 	}
 }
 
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -197,6 +203,14 @@ func OptimisticReverseBatch() QueryOption {
 	}
 }
 
+// MaxBatchSize allows the caller to limit the number of items fetched
+// in a batch.
+func MaxBatchSize(maxSize uint) QueryOption {
```
wpaulino: We should note that this is bound by `wire.MaxGetCFiltersReqRange`.

query.go
```diff
@@ -88,6 +88,12 @@ type queryOptions struct {
 	// and that we should attempt to batch more items with the query such
 	// that they can be cached, avoiding the extra round trip.
 	optimisticBatch optimisticBatchType
+
+	// maxBatchSize is the maximum items that the query should return in
+	// the case the optimisticBatch option is used.
+	// It saves bandwidth in the case the caller has a limited amount of items
```
wpaulino: Nit: fold line into previous.

query.go
roeierez: > We should note that this is bound by `wire.MaxGetCFiltersReqRange`.

I actually avoided document that as it seems to me that the query options are designed to be used in different types of queries and not only in filters query.
That being said, I am not against adding a note in the spirit of "If this option is used in filters query it is bound to `wire.MaxGetCFiltersReqRange`".
LMK what you think.

query.go
roeierez: Done.

query.go
wpaulino: Ah I see, good point. Maybe just say that each batch is bound to its request's max size, but that may already be obvious? Not a blocker in any case.

query.go
```diff
@@ -586,6 +600,10 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
 	// the maximum number of filters in anticipation of calls for the
 	// following or preceding filters.
 	var startHeight, stopHeight int64
+	batchSize := int64(wire.MaxGetCFiltersReqRange)
+	if qo.maxBatchSize > 0 && qo.maxBatchSize < wire.MaxGetCFiltersReqRange {
```
halseth: add a comment and update above comment

query.go
```diff
@@ -88,6 +88,12 @@ type queryOptions struct {
 	// and that we should attempt to batch more items with the query such
 	// that they can be cached, avoiding the extra round trip.
 	optimisticBatch optimisticBatchType
+
+	// maxBatchSize is the maximum items that the query should return in the
+	// case the optimisticBatch option is used. It saves bandwidth in the case
+	// the caller has a limited amount of items to fetch but still wants to use
+	// batching.
+	maxBatchSize uint
```
halseth: can make it `int64` to avoid below type conversion?

PR #208, opened 2020-09-24
chainservice: retry lookup IP on startup

halseth:
If network access was not yet established (e.g. on mobile device in
flight mode) the ChainService would fail to start since the IP address
of permanent peers could not be looked up.

We retry if this lookup fail, to allow the chain service to start.

Fixes https://github.com/lightningnetwork/lnd/issues/4016


wpaulino (approved):

Roasbeef (approved):
LGTM 🎮

PR #214, opened 2021-03-05
add block info to SpendReport for unspent outputs

buck54321:
That modifies the `SpendReport` struct to add fields for `BlockHash`, `BlockHeight`, and `BlockIndex` for unspent outputs. The data was already available, just not being recorded.

### DIFF
--- a/sync_test.go
+++ b/sync_test.go
@@ -316,61 +316,53 @@ func testRescan(harness *neutrinoHarness, t *testing.T) {
 	// this to test rescans and notifications.
 	modParams := harness.svc.ChainParams()
 	secSrc = newSecSource(&modParams)
-	privKey1, err := btcec.NewPrivateKey(btcec.S256())
-	if err != nil {
-		t.Fatalf("Couldn't generate private key: %s", err)
+
+	newPkScript := func() (btcutil.Address, []byte, *wire.TxOut) {
+		privKey, err := btcec.NewPrivateKey(btcec.S256())
+		if err != nil {
+			t.Fatalf("Couldn't generate private key: %s", err)
+		}
+		addr, err := secSrc.add(privKey)
+		if err != nil {
+			t.Fatalf("Couldn't create address from key: %s", err)
+		}
+		script, err := secSrc.GetScript(addr)
+		if err != nil {
+			t.Fatalf("Couldn't create script from address: %s", err)
+		}
+		return addr, script, &wire.TxOut{
+			PkScript: script,
+			Value:    1000000000,
+		}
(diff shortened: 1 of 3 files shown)


sync_test.go
```diff
@@ -379,21 +372,9 @@ func testRescan(harness *neutrinoHarness, t *testing.T) {
...
-			ourIndex = i
-		}
-	}
-	if ourIndex != 1<<30 {
-		ourOutPoint = wire.OutPoint{
-			Hash:  tx1.TxHash(),
-			Index: uint32(ourIndex),
-		}
-	} else {
-		t.Fatalf("Couldn't find the index of our output in transaction"+
```
buck54321, line 395: From what I could tell, there was no index search required, since we are constructing the transaction ourselves just a few lines up, and there is not output shuffling in the `CreateTransaction` path. Which is all for the better, because I wanted an output at a non-zero index to check that the field was being populated correctly.

wpaulino (approved):

sync_test.go
```diff
@@ -316,61 +316,54 @@ func testRescan(harness *neutrinoHarness, t *testing.T) {
...
-	}
-	// Fee rate is satoshis per byte
-	tx2, err = harness.h1.CreateTransaction(
-		[]*wire.TxOut{&out2}, 1000, true,
-	)
-	if err != nil {
-		t.Fatalf("Couldn't create transaction from script: %s", err)
+
+	newPkScript := func() (btcutil.Address, []byte, *wire.TxOut) {
+		privKey, err := btcec.NewPrivateKey(btcec.S256())
```
guggero, line 321: nit: add `t.Helper()` here too?

sync_test.go
```diff
@@ -316,61 +316,54 @@ func testRescan(harness *neutrinoHarness, t *testing.T) {
...
+			Value:    1000000000,
+		}
 	}
-	_, err = harness.h1.Node.SendRawTransaction(tx2, true)
-	if err != nil {
-		t.Fatalf("Unable to send raw transaction to node: %s", err)
+
+	createTx := func(txOuts ...*wire.TxOut) *wire.MsgTx {
+		t.Helper()
+		// Fee rate is satoshis per byte
```
guggero, line 341: nit: newline before comment.

guggero (commented):
LGTM barring two small nits. Also can you please squash the two commits? Thanks!

guggero (approved):
LGTM, thanks!

Commits should still be squashed, but we can also do that on merge. @wpaulino do you have merge permissions?

Issue #207, opened 2020-07-23
filters: add additional verification of filters for p2wsh inputs


Roasbeef:
As is, we can improve the verification of the filters we receive (given they aren't yet committed) by also verifying proper inclusion fo the expected pkScript for a given _segwit_ input. Using the code from btcutil's new `pkscript` package, [we can derive the expected `pkScript` for a given witness and/or sigScript instance](https://github.com/btcsuite/btcd/blob/master/txscript/pkscript.go#L159). This allows us to assert that the filter has the information we (typically) care about, as in LN-land, we always use native segwit outputs.

PR #215, opened 2021-03-11
Validate received filters after block download, update filter checkpoints

guggero:
Fixes https://github.com/lightninglabs/neutrino/issues/207.

Whenever we download a block, we want to make sure that the filter we
had for it was correct. It's trivial to check all output scripts of the
block's transactions were included in the filter. The trickier part is
to verify the scripts of the outputs being spent are also included in
the filter.
By looking at the signature script or witness we can derive the script
that's being spent and from that can make sure they are all included in
the filter.

### DIFF
--- a/verification.go
+++ b/verification.go
@@ -1,0 +1,154 @@
+package neutrino
+
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/gcs"
+	"github.com/btcsuite/btcutil/gcs/builder"
+)
+
+// VerifyBasicBlockFilter asserts that a given block filter was constructed
+// correctly and according to the rules of BIP-0158 to contain both the output's
+// pk scripts as well as the pk scripts the inputs are spending.
+func VerifyBasicBlockFilter(filter *gcs.Filter, block *btcutil.Block) (int,
+	error) {
+
+	var (
+		opReturnMatches int
+		key             = builder.DeriveKey(block.Hash())
+	)
+	for idx, tx := range block.Transactions() {
+		// Skip coinbase transaction.
+		if idx == 0 {
+			continue
+		}
+
+		// Check outputs first.
+		for outIdx, txOut := range tx.MsgTx().TxOut {
+			switch {
+			// If the script itself is blank, then we'll skip this
+			// as it doesn't contain any useful information.
+			case len(txOut.PkScript) == 0:
+				continue
+
+			// We'll also skip any OP_RETURN scripts as well since
+			// we don't index these in order to avoid a circular
+			// dependency.
+			case txOut.PkScript[0] == txscript.OP_RETURN:
+				// Previous versions of the filters did include
+				// OP_RETURNs. To be able disconnect bad peers
+				// still serving these old filters we attempt to
+				// check if there's an unexpected match. Since
+				// there might be false positives, an OP_RETURN
+				// can still match filters not including them.
+				// Therefore, we count the number of such
+				// unexpected matches for each peer, such that
+				// we can ban peers matching more than the rest.
+				match, err := filter.Match(key, txOut.PkScript)
+				if err != nil {
+					// Mark peer bad if we cannot match on
+					// its filter.
+					return 0, fmt.Errorf("error "+
+						"validating block %v outpoint "+
+						"%v:%d script %x: %v",
+						block.Hash(), tx.Hash(), outIdx,
+						txOut.PkScript, err)
+				}
+
+				// If it matches on the OP_RETURN output, we
+				// increase the op return counter.
+				if match {
+					opReturnMatches++
+				}
+
+				continue
+			}
+
+			// This is a "normal" script where we definitely expect
+			// a match.
+			match, err := filter.Match(key, txOut.PkScript)
+			if err != nil {
+				return 0, fmt.Errorf("error validating block "+
+					"%v outpoint %v:%d script %x: %v",
+					block.Hash(), tx.Hash(), outIdx,
+					txOut.PkScript, err)
+			}
+
+			if !match {
+				return 0, fmt.Errorf("filter for block %v is "+
+					"invalid, outpoint %v:%d script %x "+
+					"wasn't matched by filter",
+					block.Hash(), tx.Hash(), outIdx,
+					txOut.PkScript)
+			}
+		}
+
+		// Now we can go through all inputs and check that the filter
+		// also included any pk scripts of the outputs being _spent_.
+		// We can do this for witness items since the witness always
+		// contains the full script as the last element on the stack.
+		for inIdx, in := range tx.MsgTx().TxIn {
+			// There are too many edge cases to cover for non-
+			// witness scripts. And in LN land we're interested in
+			// witness spends only anyway. Therefore let's skip any
+			// input that has no witness.
+			//
+			// TODO(guggero): Add all those edge cases to
+			// ComputePkScript?
+			if len(in.Witness) == 0 {
+				continue
+			}
+
+			// The only input type that has both set is a nested
+			// P2PKH (P2SH-P2WKH). We can verify that one because
+			// the script hash has to be HASH160(OP_PUSH32 <PKH>).
+			script, err := txscript.ComputePkScript(
+				in.SignatureScript, in.Witness,
+			)
+
+			// Just skip any inputs that we can't derive the pk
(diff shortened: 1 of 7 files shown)


verification.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/gcs"
+	"github.com/btcsuite/btcutil/gcs/builder"
+)
+
+// VerifyBasicBlockFilter asserts that a given block filter was constructed
```
Roasbeef, line 12: We can likely consolidate some of this logic with what we have in `resolveFilterMismatchFromBlock`. This is called when our peers advertise a differing set of filters and we want to verify as much as we can. As is, it only verifies that all the pkScripts produced in the block are present in the filter. This works as there's no such thing as a false negative (query says it isn't there and it's actually there) with this filter type still.

verification.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for idx, tx := range block.Transactions() {
+		// Skip coinbase transaction.
+		if idx == 0 {
+			continue
+		}
+
+		// Check outputs first.
+		for outIdx, out := range tx.MsgTx().TxOut {
+			match, err := filter.MatchAny(
```
Roasbeef, line 26: I think `Match` can be used safely here instead.

verification.go
```diff
@@ -0,0 +1,101 @@
...
+					out.PkScript)
+			}
+		}
+
+		// Now we can go through all inputs and check that the filter
+		// also included any pk scripts of the outputs being _spent_.
+		// We can do this for witness items since the witness always
+		// contains the full script as the last element on the stack.
+		for inIdx, in := range tx.MsgTx().TxIn {
+			script, err := txscript.ComputePkScript(
```
Roasbeef, line 50: Note that as is, this doesn't cover nested P2SH, as the logic in `ComputePkScript` branches based on the length of the witness and sigScript. We can bridge this by checking if _both_ are active, then we'd compute the witness program based on the witness script, then compute the P2SH script of that and assert everything matches up. cc @wpaulino

Roasbeef: Actually, I think it's fine as is, since we'll return the P2SH script here which is what should be inserted into the filter...

verification.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				log.Debug("Skipping filter validation for "+
+					"input %d of tx %v in block %v "+
+					"because computing the script failed: "+
+					"%v", inIdx, block.Hash(), err)
+
+				continue
+			}
+
+			match, err := filter.MatchAny(
```
Roasbeef, line 80: Similar comment here re just using `Match`.

verification_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+var (
+	chainParams = &chaincfg.RegressionNetParams
+
+	dummySignature = []byte{34, 56, 78, 99}
+)
+
+// TestVerifyBlockFilter tests that a filter is correctly inspected for validity
+// against a downloaded block.
+func TestVerifyBlockFilter(t *testing.T) {
```
Roasbeef, line 25: Solid test! 

In the past we never implemented this tyupe of verification as we just had a hunch that all the basses were covered. We then created the `ComputePkScript` when we moved to support pkScript based spend+confirmation notifications, knowing that if we were correct, we could eventually use that here once we were confident in our assumptions.

Roasbeef (commented):
It lives!!! 

I think the only thing at this point we need to do is go back through `ComputePkScript` and ensure there aren't any standard script edge cases we missed. In addition we may want to go over things again to check that there aren't any stronger checks we can use to assert.

verification.go
guggero: Ah, I wasn't aware of this function. But it definitely makes sense to consolidate, which I've done now.

verification_test.go
guggero: It turns out that bare P2PK isn't handled correctly by `ComputePkScript`. Maybe there are more of the "standard" transaction types that aren't fully supported yet. I added a TODO for now and skipped any non-witness inputs. That way we can at least get the witness verification part in.

guggero:
I've cleaned up the PR and addressed all the comments so far.

Two questions remain:
 - It turned out that bare P2PK aren't supported by `ComputePkScript` (they're not detected and P2SH is assumed). Do we want to block this PR on the fix or do we want to continue with the witness only verification for now?
 - In the rescan logic (where we actually download the block and verify it), we don't know which peer we got the filter from originally. So there's no way for us to disconnect/ban that peer in a graceful manner. Is it okay to just error out in that case? Continuing with a bad filter would probably be quite bad. Or what else could we do in that case?

Roasbeef:
> It turned out that bare P2PK aren't supported by ComputePkScript (they're not detected and P2SH is assumed). Do we want to block this PR on the fix or do we want to continue with the witness only verification for now?

Ah I think we opted to ignore those as we needed this for stuff in the chain notifier, and nothing in lnd uses bare P2PK. I'm thinking we should either make it possible to ignore those, or update the upstream library to support it. Thoughts @wpaulino? 

> In the rescan logic (where we actually download the block and verify it), we don't know which peer we got the filter from originally

Yeah we definitely shouldn't continue with a bad filter there. An assumption here is that if we're able to do a full header chain sync w/o finding any divergence, then we do indeed have the correct set of filters. However, this may not be the case in practice assuming we had all bad peers to begin with. Only until we either download a filter, to go verify it and find out it's invalid or we get a proper peer will we discover that we had an invalid chain. However as is, we don't unwind our filter header chain (and wipe all filters) in order to start from scratch...

Roasbeef (commented):
Latest version reads well, will start testing this in the wild now. Discussion below has led to the re-acknowledgement that we do indeed need to unwind (read delete) our filter header chain if we discovered an empirical conflict after we've already synced a filter header chain. However, I think that's best left as a follow up issue.

verification.go
```diff
@@ -0,0 +1,154 @@
...
+		// contains the full script as the last element on the stack.
+		for inIdx, in := range tx.MsgTx().TxIn {
+			// There are too many edge cases to cover for non-
+			// witness scripts. And in LN land we're interested in
+			// witness spends only anyway. Therefore let's skip any
+			// input that has no witness.
+			//
+			// TODO(guggero): Add all those edge cases to
+			// ComputePkScript?
+			if len(in.Witness) == 0 {
```
wpaulino, line 100: `ComputePkScript` supports `P2PKH` as well, so we could opt to verify those as well for users still using those addresses.

verification.go
```diff
@@ -0,0 +1,154 @@
...
+			// P2PKH (P2SH-P2WKH). We can verify that one because
+			// the script hash has to be HASH160(OP_PUSH32 <PKH>).
+			script, err := txscript.ComputePkScript(
+				in.SignatureScript, in.Witness,
+			)
+
+			// Just skip any inputs that we can't derive the pk
+			// script from.
+			if err == txscript.ErrUnsupportedScriptType {
+				log.Debugf("Skipping filter validation for "+
```
wpaulino, line 114: Demote to trace? This will be very spammy otherwise.

wpaulino (commented):
>> It turned out that bare P2PK aren't supported by ComputePkScript (they're not detected and P2SH is assumed). Do we want to block this PR on the fix or do we want to continue with the witness only verification for now?

> Ah I think we opted to ignore those as we needed this for stuff in the chain notifier, and nothing in lnd uses bare P2PK. I'm thinking we should either make it possible to ignore those, or update the upstream library to support it. Thoughts @wpaulino?

It doesn't seem worth adding P2PK support for our use case (who's still using it in _2021_!?), and any additional verification wouldn't provide much for us since we're already verifying what we expect is relevant to us.

verification.go
guggero: You mean run `ComputePkScript` even if there is no witness and see if the returned script is P2PKH and only then verify it? I'm a big afraid that we'd run into another edge case here. If I look at the `computeNonWitnessPkScript` function, it looks like we could be very unlucky and if a P2PK script has a 0x02 or 0x03 in the wrong place of the signature, this would actually match and we'd fail the whole block for it.
I'd have a better feeling if we added more cases to the non-witness script computation first before we fail the verification here.

wpaulino: Ah I see, good points. Probably not worth the hassle anyway.

wpaulino (approved):

Roasbeef (approved):
LGTM 🦚

PR #216, opened 2021-03-19
neutrino: allow custom block cache

ellemouton:
This PR allows a custom block cache to be passed to neutrino so that there can be a shared block cache between neutrino and whichever client may be using it. 

This diff is a step towards having an LND block cache (WIP is [here](https://github.com/ellemouton/lnd/pull/8))

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -566,8 +566,13 @@ type Config struct {
 	// hold in memory at most.
 	FilterCacheSize uint64
 
+	// BlockCache is an LRU block cache. If none is provided then the a new
+	// one will be instantiated.
+	BlockCache *lru.Cache
+
 	// BlockCacheSize indicates the size (in bytes) of blocks the block
-	// cache will hold in memory at most.
+	// cache will hold in memory at most. If a BlockCache is provided then
+	// BlockCacheSize is ignored.
 	BlockCacheSize uint64
 
 	// AssertFilterHeader is an optional field that allows the creator of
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -965,7 +965,7 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
 	inv := wire.NewInvVect(invType, &blockHash)
 
 	// If the block is already in the cache, we can return it immediately.
-	blockValue, err := s.BlockCache.Get(*inv)
+	blockValue, err := s.BlockCache.Get(blockHash)
```
guggero, line 968: The main difference to before is that we now don't use the `invType` in the key anymore now. I'm not sure if this can have any side effects, @wpaulino? I assume not since we're only interested in witness blocks anyway. Just wanted to highlight this.

guggero (approved):
Great work on this, thanks!
The code looks good to me and as far as I can judge this shouldn't change the overall behavior :+1:

query.go
wpaulino: I had suggested this change to maintain the cache keys used here and within lnd consistent. I didn't realize this would break the API however, as now the block won't include the witnesses if it was previously cached without them.

@ellemouton It seems like we'll need to introduce the inv type to lnd's cache key as well, and just always set it to `WitnessEncoding` as that's the default for `GetBlock` on other backends.

ellemouton: cool, updated 👍

wpaulino (approved):

PR #217, opened 2021-03-23
Add Makefile, use golangci-lint, move to GitHub actions

guggero:
As a new contributor to the Neutrino repository it was a bit of a challenge to figure out how to use the clunky `gotest.sh` file. That's why I think it would be nice to switch over to a `Makefile`.

At the same time I felt like we should fix a number of smaller issues that an updated linter should find. That's why I opted to add the `golangci-lint` linter as well.
And I think I might even have discovered a possible explanation for some of the flakes we see in the `lnd` itests with Neutrino (see 0a824919, could also be mistaken about the significance of this).


blockntfns/manager.go
```diff
@@ -139,6 +139,7 @@ func (m *SubscriptionManager) Stop() {
 	var wg sync.WaitGroup
 	wg.Add(len(m.subscribers))
 	for _, subscriber := range m.subscribers {
+		subscriber := subscriber
```
wpaulino, line 142: Nit: could also provide it as an argument to the goroutine func.

batch_spend_reporter.go
```diff
@@ -61,14 +61,16 @@ func (b *batchSpendReporter) NotifyUnspentAndUnfound() {
...
+
 		// A nil SpendReport indicates the output was not found.
 		tx, ok := b.initialTxns[outpoint]
 		if !ok {
 			log.Warnf("Unknown initial txn for getuxo request %v",
 				outpoint)
 		}
 
-		b.notifyRequests(&outpoint, requests, tx, nil)
+		b.notifyRequests(&op, requests, tx, nil)
```
wpaulino, line 73: :+1:

blockmanager.go
```diff
@@ -803,7 +803,7 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 	// set of peers.
 	pristineHeaders, ok := headers[key]
 	if !ok {
-		return fmt.Errorf("All peers served bogus headers! Retrying " +
+		return fmt.Errorf("all peers served bogus headers! Retrying " +
```
wpaulino, line 806: Nit: replace `!` with a comma?

neutrino.go
```diff
@@ -1489,9 +1489,13 @@ func (s *ChainService) Start() error {
 	s.addrManager.Start()
 	s.blockManager.Start()
 	s.blockSubscriptionMgr.Start()
-	s.workManager.Start()
+	if err := s.workManager.Start(); err != nil {
```
wpaulino, line 1492: :+1:

blockmanager_test.go
```diff
@@ -440,7 +440,7 @@ func TestBlockManagerInitialInterval(t *testing.T) {
 		if test.partialInterval {
 			startHeight = wire.CFCheckptInterval / 3
 		}
-		go func() {
+		go func() { // nolint:staticcheck
```
wpaulino, line 443: Can we remove the lint exception for all the `t.Fatal` in goroutines and follow https://golang.org/doc/go1.16#vet?

headerfs/store.go
```diff
@@ -85,14 +85,14 @@ var headerBufPool = sync.Pool{
...
-// to a databse which indexes that flat file. Together, these two abstractions
+// to a database which indexes that flat file. Together, these two abstractions
 // can be used in order to build an indexed header store for any type of
 // "header" as it deals only with raw bytes, and leaves it to a higher layer to
 // interpret those raw bytes accordingly.
 //
 // TODO(roasbeef): quickcheck coverage
 type headerStore struct {
-	mtx sync.RWMutex
+	mtx sync.RWMutex // nolint:structcheck, false positive
```
wpaulino, line 95: Interesting. I removed the lint exception to see what the warning was and got `mtx is unused`. It seems to not like the fact that `headerStore` is embedded in other structs.

neutrino.go
```diff
@@ -158,7 +158,7 @@ type ServerPeer struct {
 	server         *ChainService
 	persistent     bool
 	knownAddresses map[string]struct{}
-	banScore       connmgr.DynamicBanScore
+	banScore       connmgr.DynamicBanScore // nolint:unused, false positive
```
wpaulino, line 161: The ban scoring logic was removed in a previous commit, this should probably be removed there as well.

pushtx/broadcaster.go
```diff
@@ -81,19 +81,19 @@ func NewBroadcaster(cfg *Config) *Broadcaster {
 // Start starts all of the necessary steps for the Broadcaster to begin properly
 // carrying out its duties.
 func (b *Broadcaster) Start() error {
-	var err error
+	var returnErr error
```
wpaulino, line 84: :+1:

query/workmanager_test.go
```diff
@@ -202,13 +202,13 @@ func TestWorkManagerWorkDispatcherFailures(t *testing.T) {
...
-	// Fot each wotker, spin up a goroutine that will forward the job it
-	// got to our slice of sheduled jobs, such that we can handle them in
+	// Fot each worker, spin up a goroutine that will forward the job it
+	// got to our slice of scheduled jobs, such that we can handle them in
 	// order.
 	for i := 0; i < len(workers); i++ {
 		wk := workers[i]
 		go func() {
-			for {
+			for { // nolint:gosimple
```
wpaulino, line 211: Looks like we can remove the `select` since we're only waiting on one channel?

wpaulino (commented):
Neutrino was clearly in need of a better linter! `btcwallet` would be another good candidate.

It looks like the GitHub actions aren't running yet, is there anything else that needs to be done?

neutrino.go
guggero: Right, removed.

blockmanager_test.go
guggero: Ah, I didn't know it was that easy to fix (expected a bigger refactor). Thanks for the hint, done!

guggero:
> It looks like the GitHub actions aren't running yet, is there anything else that needs to be done?

Changes in GH actions in a PR aren't applied until the PR is merged. I think it's a security measure to make sure repository secrets cannot be extracted easily just by opening a PR.

headerfs/store.go
guggero: Yeah, this is a bit annoying. I tried the newest version of `golanglint-ci` but that also still flags it. Added a comment to make clear why it's a false positive.

wpaulino (approved):

guggero:
@cfromknecht PTAL

.golangci.yml
```diff
@@ -0,0 +1,51 @@
...
+  disable:
+    # Global variables are used in many places throughout the code base.
+    - gochecknoglobals
+
+    # Some lines are over 80 characters on purpose and we don't want to make them
+    # even longer by marking them as 'nolint'.
+    - lll
+
+    # We don't care (enough) about misaligned structs to lint that.
+    - maligned
```
halseth, line 24: maybe some of these could be enabled?

batch_spend_reporter.go
```diff
@@ -61,14 +61,16 @@ func (b *batchSpendReporter) NotifyUnspentAndUnfound() {
 	log.Debugf("Finished batch, %d unspent outpoints", len(b.requests))
 
 	for outpoint, requests := range b.requests {
+		op := outpoint
```
halseth, line 64: yeah, good to catch these! 😀

halseth (approved):
Excellent! Awesome to finally bring neutrino into the new decade 🚀

halseth:
@guggero Feel free to merge when you've checked out whether the comments are actionable

.golangci.yml
guggero: Yeah, why not. I enabled `malign` globally but only fixed one struct that is used often. For the others the readability/order of the struct would suffer too much for very little benefit.

Also skipping the `prealloc` check only for test code.

guggero:
@halseth I rebased and added two more commits. I don't have merge permissions in this repo so feel free to push the button if you think this is ready.

.golangci.yml
```diff
@@ -48,3 +44,7 @@ issues:
         - gosec
         - errcheck
         - dupl
+
+        # Instances of table driven tests that don't pre-allocate shouldn't
+        # trigger the linter.
+        - prealloc
```
halseth, line 50: 👍

halseth (approved):

PR #220, opened 2021-04-28
neutrino+pushtx: make broadcast timeout configurable with fallback

wpaulino:
This is useful for enabling shorter timeouts than we would use in a production environment in testing environments, resulting in shorter test execution times. If a timeout isn't provided, we'll use the default of 5 seconds.


guggero (approved):
LGTM :tada:

Roasbeef (approved):
LGTM 🍀

PR #226, opened 2021-09-21
pushtx+rescan: stop rebroadcast on tx confirm

### LINKED ISSUE #206: Neutrino seems to be expecting a MsgReject when broadcasting a tx, but btcd does not reply to the INV
In order to decide when to stop rebroadcasting, [Broadcaster is expecting to see a MsgReject](https://github.com/lightninglabs/neutrino/blob/38bf97e348a3d89c8f473209abf811c64074e5b5/pushtx/broadcaster.go#L236). But I see that broadcaster calls cfg.Broadcast which ends in [ChainService.sendTransaction()](https://github.com/lightninglabs/neutrino/blob/39da434881572f823bee55994995a9e46fc985e9/query.go#L1118) and this is sending an INV message. On the btcd side, INV messages which are already known *in any way* are quietly ignored, as you can see in [SyncManager.handleInvMessage()](https://github.com/btcsuite/btcd/blob/96897255fd17525dd12426345d279533780bc4e1/netsync/manager.go#L1118) so unless I am mistaken, this idea of rebroadcasting until someone rejects as duplicate is not just non-ideal, it actually doesn't work in the normal case (unless bitcoind works differently, in which a case could be made that this is btcd's bug).

ellemouton:
This PR allows rescan to notify the `ChainServices` `broadcaster` that a relevant transaction is confirmed.

Fixes #206  using the approach mentioned [here](https://github.com/lightninglabs/neutrino/issues/206#issuecomment-887919095)

**Testing the PR**
- _To see the bug_: An easy way to test this just by running an lnd itest that creates some channels and then observing the logs. So currently, if you run `$ make itest backend=neutrino icase=....` (choose a test that creates a lot of channels) and then look at the logs: `[DBG] BTCN: Re-broadcasting %d transactions`  will show that the number of txs being rebroadcasted just keeps increasing. 
- _To see the fix:_ now in lnd go.mod, add: `replace github.com/lightninglabs/neutrino => github.com/ellemouton/neutrino v0.11.1-0.20210921111409-eb86c412013d` and re-run the same itest. You will see that the number of txs being rebroadcast no longer increases when a tx is confirmed.

Fixes https://github.com/lightninglabs/neutrino/issues/206

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -1007,6 +1007,13 @@ func extractBlockMatches(chain ChainSource, ro *rescanOptions,
 
 		if relevant {
 			relevantTxs = append(relevantTxs, tx)
+
+			chainSource, ok := chain.(*RescanChainSource)
+			if ok {
+				chainSource.broadcaster.MarkAsConfirmed(
+					*tx.Hash(),
+				)
+			}
 		}
 	}
 
--- a/pushtx/broadcaster.go
+++ b/pushtx/broadcaster.go
@@ -67,6 +67,10 @@ type Broadcaster struct {
 	// requests from external callers will be streamed through.
 	broadcastReqs chan *broadcastReq
 
+	// confChan is a channel used to notify the broadcast handler about
+	// confirmed transactions.
+	confChan chan chainhash.Hash
+
 	quit chan struct{}
 	wg   sync.WaitGroup
 }
@@ -76,6 +80,7 @@ func NewBroadcaster(cfg *Config) *Broadcaster {
 	b := &Broadcaster{
 		cfg:           *cfg,
 		broadcastReqs: make(chan *broadcastReq),
+		confChan:      make(chan chainhash.Hash),
 		quit:          make(chan struct{}),
 	}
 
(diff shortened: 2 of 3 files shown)


rescan.go
```diff
@@ -1007,6 +1007,11 @@ func extractBlockMatches(chain ChainSource, ro *rescanOptions,
 
 		if relevant {
 			relevantTxs = append(relevantTxs, tx)
+
+			cs, ok := chain.(*RescanChainSource)
```
arshbot, line 1011: nit: `cs` for `chainsource`? Although it doesn't seem enforced in this codebase, longer names could improve readability :)

arshbot (changes requested):
Love the minimal diff! Great commit structure. Left a tiny nit but this is fairly straightforward.

rescan.go
ellemouton: sure, updated 👍

Roasbeef (approved):
Very straightforward fix! 

LGTM 🧩

Will need a PR now to the `lnd` repo to update the go.mod there

Roasbeef:
Also re which "option" this PR took: with the added method those that use the library outside of the lnd+btcwallet interaction can use that method to clean up the main state.

ellemouton:
> added method those that use the library outside of the lnd+btcwallet interaction can use that method to clean up the main state.

Hmmm but the method is not exposed to them if they just have ChainServices. Should I add a method to ChainServices to expose this method on Broadcaster?

ellemouton:
> Will need a PR now to the lnd repo to update the go.mod there

Is it fine to point to the latest commit or shall we push a new tag to neutrino?

Kenme1234 (commented):
Fb

PR #230, opened 2021-10-18
chain service: use safe AddrManager address accessor

chappjc:
This PR requires https://github.com/btcsuite/btcd/pull/1760.  The `replace` would be removed and the btcd `require` updated if that is merged.  The data races between neutrino and btcd are described on that PR.

The `*KnownAddress` returned by `(*AddrManager).GetAddress` is not safe for use while other goroutines are using the `AddrManager`'s methods for updating peer states e.g. `Connect`/`Attempt`/`AddAddresses`/`SetServices`.

The `newAddressFunc` closure created by `NewChainService` as the `ConnManager`'s `GetNewAddress` function was using `GetNetAddress` concurrent with other peer management goroutines. e.g. a `Peer`'s `inHandler` that has a callbacks to `AddAddresses` and `SetServices`.

This resolves the issue by using the new `GetNetAddress` method instead of `GetAddress`. This is safe since the returned `wire.NetAddress` is immutable, unlike the `KnownAddress` from `GetAddress`.


Roasbeef:
Thanks for this PR! 

Re the underlying bug, is there any concrete unintended interaction that results from it? So like us re-trying the same address over and over again, when we think it's actually a new address?

chappjc:
An update that I've proposed an alternate btcd fix that requires no code changes from neutrino, just a btc require bump: https://github.com/btcsuite/btcd/pull/1763. dcrd had fixed it that way (add a mutex instead of making a copies for external consumers).  Either way seems fine to me, although https://github.com/btcsuite/btcd/pull/1763 requires no code changes to neutrino, and it does not touch the API.

PR #231, opened 2021-10-18
server: Actively prevent regnet network discovery

chappjc:
This treats regnet like simnet with respect to node discovery and outbound connections.

Without this change, using regnet without `--connect`/`ConnectPeers` results in peer connection errors being logged ad nauseum.  e.g. tons of logs like `Failed to connect to reqid 78: no valid connect address`, eventually giving up.,
The workaround is to always use `ConnectPeers` on regnet, but peer discovery is really best disabled for regnet as well as simnet.

A similar change was made in dcrd in https://github.com/decred/dcrd/commit/f5fdd95eb578a485b414331836f9731fcd0f2756

Would like to test a bit more to be sure nothing unexpected happens on our regnet harness, so in draft for the moment.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -84,6 +84,13 @@ var (
 	DefaultBlockCacheSize uint64 = 4096 * 10 * 1000 // 40 MB
 )
 
+// isDevNetwork indicates if the chain is a private development network, namely
+// simnet or regtest/regnet.
+func isDevNetwork(net wire.BitcoinNet) bool {
+	return net == chaincfg.SimNetParams.Net ||
+		net == chaincfg.RegressionNetParams.Net
+}
+
 // updatePeerHeightsMsg is a message sent from the blockmanager to the server
 // after a new block has been accepted. The purpose of the message is to update
 // the heights of peers that were known to announce the block before we
@@ -316,11 +323,11 @@ func (sp *ServerPeer) OnReject(_ *peer.Peer, msg *wire.MsgReject) {
 // OnAddr is invoked when a peer receives an addr bitcoin message and is
 // used to notify the server about advertised addresses.
 func (sp *ServerPeer) OnAddr(_ *peer.Peer, msg *wire.MsgAddr) {
-	// Ignore addresses when running on the simulation test network.  This
+	// Ignore addresses when running on a private development network.  This
 	// helps prevent the network from becoming another public test network
 	// since it will not be able to learn about other peers that have not
 	// specifically been provided.
-	if sp.server.chainParams.Net == chaincfg.SimNetParams.Net {
+	if isDevNetwork(sp.server.chainParams.Net) {
 		return
 	}
 
@@ -741,12 +748,12 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	s.blockSubscriptionMgr = blockntfns.NewSubscriptionManager(s.blockManager)
 
 	// Only setup a function to return new addresses to connect to when not
-	// running in connect-only mode.  The simulation network is always in
-	// connect-only mode since it is only intended to connect to specified
-	// peers and actively avoid advertising and connecting to discovered
-	// peers in order to prevent it from becoming a public test network.
+	// running in connect-only mode.  Private development networks are always in
+	// connect-only mode since it is only intended to connect to specified peers
+	// and actively avoid advertising and connecting to discovered peers in
+	// order to prevent it from becoming a public test network.
 	var newAddressFunc func() (net.Addr, error)
-	if s.chainParams.Net != chaincfg.SimNetParams.Net {
+	if !isDevNetwork(s.chainParams.Net) {
 		newAddressFunc = func() (net.Addr, error) {
 
 			// Gather our set of currently connected peers to avoid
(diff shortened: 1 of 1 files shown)


neutrino.go
```diff
@@ -320,7 +320,8 @@ func (sp *ServerPeer) OnAddr(_ *peer.Peer, msg *wire.MsgAddr) {
 	// helps prevent the network from becoming another public test network
 	// since it will not be able to learn about other peers that have not
 	// specifically been provided.
-	if sp.server.chainParams.Net == chaincfg.SimNetParams.Net {
+	if sp.server.chainParams.Net == chaincfg.SimNetParams.Net ||
```
Roasbeef, line 323: How about we spin these repeated conditionals out into something like `isDevNetwork`? Applies also in the diff for the other file.

Roasbeef (changes requested):

neutrino.go
chappjc: Made a `isDevNetwork` helper taking a `wire.BitcoinNet`.  Seemed cleanest.

blockmanager.go
```diff
@@ -2680,9 +2680,7 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 
 	// If not current, request the next batch of headers starting from the
 	// latest known header and ending with the next checkpoint.
-	if b.cfg.ChainParams.Net == chaincfg.SimNetParams.Net ||
-		b.cfg.ChainParams.Net == chaincfg.RegressionNetParams.Net ||
-		!b.BlockHeadersSynced() {
+	if isDevNetwork(b.cfg.ChainParams.Net) || !b.BlockHeadersSynced() {
```
ellemouton, line 2683: nit: I dont think this one is necessary. Looks like this is only done for unit testing purposes.

neutrino.go
```diff
@@ -747,8 +753,7 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	// peers and actively avoid advertising and connecting to discovered
 	// peers in order to prevent it from becoming a public test network.
 	var newAddressFunc func() (net.Addr, error)
-	if s.chainParams.Net != chaincfg.SimNetParams.Net &&
-		s.chainParams.Net != chaincfg.RegressionNetParams.Net {
+	if !isDevNetwork(s.chainParams.Net) {
```
ellemouton, line 756: nit: can you update the comment above to indicate that we are no longer just switching on simnet here.

neutrino.go
```diff
@@ -1265,8 +1270,7 @@ func (s *ChainService) handleAddPeerMsg(state *peerState, sp *ServerPeer) bool {
 	// the simulation test network since it is only intended to connect to
 	// specified peers and actively avoids advertising and connecting to
 	// discovered peers.
-	if s.chainParams.Net != chaincfg.SimNetParams.Net &&
-		s.chainParams.Net != chaincfg.RegressionNetParams.Net {
+	if !isDevNetwork(s.chainParams.Net) {
```
ellemouton, line 1273: same as above re comment

neutrino.go
```diff
@@ -320,8 +327,7 @@ func (sp *ServerPeer) OnAddr(_ *peer.Peer, msg *wire.MsgAddr) {
 	// helps prevent the network from becoming another public test network
 	// since it will not be able to learn about other peers that have not
 	// specifically been provided.
-	if sp.server.chainParams.Net == chaincfg.SimNetParams.Net ||
-		sp.server.chainParams.Net == chaincfg.RegressionNetParams.Net {
+	if isDevNetwork(sp.server.chainParams.Net) {
```
ellemouton, line 330: same as below re comment

neutrino.go
```diff
@@ -84,6 +84,13 @@ var (
 	DefaultBlockCacheSize uint64 = 4096 * 10 * 1000 // 40 MB
 )
 
+// isDevNetwork indicates if the chain is a development network, namely simnet
+// or regtest/regnet.
+func isDevNetwork(net wire.BitcoinNet) bool {
+	return net == chaincfg.SimNetParams.Net ||
+		net == chaincfg.RegressionNetParams.Net
+}
+
```
ellemouton, line 93: This commit can be squashed into the first commit

ellemouton (commented):
tACK 🚀  just pending some nits re comments. Also think this should all be squashed into 1 commit. I also think the change in `blockmanager.go` is not needed for this.

blockmanager.go
chappjc: I reverted this one back to just simnet, so no changes on blockmanager.go anymore.

neutrino.go
chappjc: Done.  Sorry about re-wrapping the comment.  It was starting to look wonky after making the change.

ellemouton (approved):
LGTM 🚀

chappjc:
Thanks, @lightninglabs-deploy, it's already approved. :stuck_out_tongue_winking_eye:

ellemouton:
@chappjc , it needs 2 approvals :) so just need to re-request review from the other reviewer

Roasbeef (approved):
LGTM 🦧

PR #235, opened 2021-11-09
set headerfs.BlockStamp.Timestamp

chappjc:
The `headerfs.BlockStamp` returned from `(*ChainService).BestBlock` had a zero `Timestamp`.
This change sets the `Timestamp` field from the block header if available.

Internally it does not seem to matter that `Timestamp` is the zero `time.Time`, but for `ChainService` consumers it is helpful to have the block's timestamp.  For example, one may compare the timestamp from the current best block from `ChainService` to the wallet `Manager`'s birthday time.

### DIFF
--- a/headerfs/store.go
+++ b/headerfs/store.go
@@ -330,14 +330,15 @@ func (h *blockHeaderStore) RollbackLastBlock() (*BlockStamp, error) {
 		return nil, err
 	}
 
-	// With this height obtained, we'll use it to read the latest header
+	// With this height obtained, we'll use it to read the previous header
 	// from disk, so we can populate our return value which requires the
 	// prev header hash.
-	bestHeader, err := h.readHeader(chainTipHeight)
+	prevHeader, err := h.readHeader(chainTipHeight - 1)
 	if err != nil {
 		return nil, err
 	}
-	prevHeaderHash := bestHeader.PrevBlock
+
+	prevHeaderHash := prevHeader.BlockHash()
 
 	// Now that we have the information we need to return from this
 	// function, we can now truncate the header file, and then use the hash
@@ -350,8 +351,9 @@ func (h *blockHeaderStore) RollbackLastBlock() (*BlockStamp, error) {
 	}
 
 	return &BlockStamp{
-		Height: int32(chainTipHeight) - 1,
-		Hash:   prevHeaderHash,
+		Height:    int32(chainTipHeight) - 1,
+		Hash:      prevHeaderHash,
+		Timestamp: prevHeader.Timestamp,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


headerfs/store.go
```diff
@@ -349,9 +349,15 @@ func (h *blockHeaderStore) RollbackLastBlock() (*BlockStamp, error) {
...
 
+	// prevHeader, err := h.readHeader(chainTipHeight - 1)
+	// if err != nil {
+	// 	return nil, err
+	// }
+
 	return &BlockStamp{
 		Height: int32(chainTipHeight) - 1,
 		Hash:   prevHeaderHash,
+		// Timestamp: prevHeader.Timestamp,
```
chappjc, line 360: I am less certain about this one.

neutrino.go
```diff
@@ -942,8 +942,9 @@ func (s *ChainService) BestBlock() (*headerfs.BlockStamp, error) {
 	}
 
 	return &headerfs.BlockStamp{
-		Height: int32(bestHeight),
-		Hash:   bestHeader.BlockHash(),
+		Height:    int32(bestHeight),
+		Hash:      bestHeader.BlockHash(),
+		Timestamp: bestHeader.Timestamp,
```
chappjc, line 947: This is the only one I'm really after, but I did most others when the data was available for consistency.

headerfs/store.go
ellemouton: Probably can remove reading the current header completely:

```
	// With this height obtained, we'll use it to read the previous header
	// from disk, so we can populate our return value which requires the
	// prev header hash.
	prevHeader, err := h.readHeader(chainTipHeight - 1)
	if err != nil {
		return nil, err
	}

	prevHeaderHash := prevHeader.BlockHash()

	// Now that we have the information we need to return from this
	// function, we can now truncate the header file, and then use the hash
	// of the prevHeader to set the proper index chain tip.
	if err := h.singleTruncate(); err != nil {
		return nil, err
	}
	if err := h.truncateIndex(&prevHeaderHash, true); err != nil {
		return nil, err
	}

	return &BlockStamp{
		Height:    int32(chainTipHeight) - 1,
		Hash:      prevHeaderHash,
		Timestamp: prevHeader.Timestamp,
	}, nil
```

chappjc: Oh yeah, `bestHeader` was only retrieved for `PrevBlock` hash.  I'll make this change too.

ellemouton (approved):
tACK 🚀

Roasbeef (approved):
LGTM 🪓

chappjc:
:pray:  Thanks!  This will be a helpful change on our end.

PR #236, opened 2021-11-11
neutrino+query+rescan: improve rescan speed

ellemouton:
This PR should probably be split into 3 PR's. Doing it in 1 for now just to get some initial feedback and to better show the progression. The three sections are as follows:

1. The first 2 commits alter the `GetBlock` and `GetCFilter` functions to use the work dispatcher for their queries instead of using the old `queryPeers` function. The function is now removed bringing us one step closer to removing all query logic from the main package. 
2. The 3rd commit isolates the bottleneck of the `GetCFilter` function which is persisting filters to the DB. In this commit, this operation is spun off into a goroutine thus allowing the `GetCFilter` function to return as soon as all the filters are written to the cache. 
3. Finally, the 4th commit ensures that `rescan` can make use of batch filter fetching by waiting until the header chain is either current or until it is ahead of the specified end height. Before this commit, if rescan is started before the chain is current, then the filters are fetched one by one instead which is what makes things super slow. For example, on local regtest: before this commit it would take 7 seconds to sync 3000 blocks but with this commit, it takes 900ms. (full testnet rescan from local testnet bitcoind took like 5 mins)

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -597,6 +597,7 @@ type ChainService struct { // nolint:maligned
 	utxoScanner          *UtxoScanner
 	broadcaster          *pushtx.Broadcaster
 	banStore             banman.Store
+	queryDispatcher      query.Dispatcher
 	workManager          *query.WorkManager
 
 	// peerSubscribers is a slice of active peer subscriptions, that we
@@ -678,6 +679,8 @@ func NewChainService(cfg Config) (*ChainService, error) {
 		NewWorker:      query.NewWorker,
 		Ranking:        query.NewPeerRanking(),
 	})
+
+	s.queryDispatcher = s.workManager
 
 	// We set the queryPeers method to point to queryChainServicePeers,
 	// passing a reference to the newly created ChainService.
--- a/neutrino.go
+++ b/neutrino.go
@@ -575,11 +575,6 @@ type ChainService struct { // nolint:maligned
 	FilterCache *lru.Cache
 	BlockCache  *lru.Cache
 
-	// queryPeers will be called to send messages to one or more peers,
-	// expecting a response.
-	queryPeers func(wire.Message, func(*ServerPeer, wire.Message,
-		chan<- struct{}), ...QueryOption)
-
 	chainParams          chaincfg.Params
 	addrManager          *addrmgr.AddrManager
 	connManager          *connmgr.ConnManager
@@ -681,13 +676,6 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	})
 
 	s.queryDispatcher = s.workManager
-
-	// We set the queryPeers method to point to queryChainServicePeers,
-	// passing a reference to the newly created ChainService.
-	s.queryPeers = func(msg wire.Message, f func(*ServerPeer,
-		wire.Message, chan<- struct{}), qo ...QueryOption) {
-		queryChainServicePeers(&s, msg, f, qo...)
-	}
 
 	var err error
 
--- a/query.go
+++ b/query.go
@@ -335,168 +335,6 @@ checkResponses:
 	}
 }
 
-// queryChainServicePeers is a helper function that sends a query to one or
-// more peers of the given ChainService, and waits for an answer. The timeout
-// for queries is set by the QueryTimeout package-level variable or the Timeout
-// functional option.
-func queryChainServicePeers(
-	// s is the ChainService to use.
-	s *ChainService,
-
-	// queryMsg is the message to send to each peer selected by selectPeer.
-	queryMsg wire.Message,
-
-	// checkResponse is called for every message within the timeout period.
-	// The quit channel lets the query know to terminate because the
-	// required response has been found. This is done by closing the
-	// channel.
-	checkResponse func(sp *ServerPeer, resp wire.Message,
-		quit chan<- struct{}),
-
-	// options takes functional options for executing the query.
-	options ...QueryOption) {
-
-	// Starting with the set of default options, we'll apply any specified
-	// functional options to the query.
-	qo := defaultQueryOptions()
-	qo.applyQueryOptions(options...)
-
-	// We get an initial view of our peers, to be updated each time a peer
-	// query times out.
-	queryPeer := s.blockManager.SyncPeer()
-	peerTries := make(map[string]uint8)
-
-	// This will be state used by the peer query goroutine.
-	queryQuit := make(chan struct{})
-	subQuit := make(chan struct{})
-
-	// Increase this number to be able to handle more queries at once as
-	// each channel gets results for all queries, otherwise messages can
-	// get mixed and there's a vicious cycle of retries causing a bigger
-	// message flood, more of which get missed.
-	msgChan := make(chan spMsg)
-	subscription := spMsgSubscription{
-		msgChan:  msgChan,
-		quitChan: subQuit,
-	}
-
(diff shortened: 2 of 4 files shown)


query.go
```diff
@@ -782,7 +784,33 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+				select {
+				case resp, ok = <-persistChan:
+					if !ok {
+						return
+					}
+
+				case <-s.quit:
+					return
+				}
+
```
ellemouton, line 813: One worry with this is that we spin up like 700 goroutines (on mainnet). So another option to rate limit this a bit is to have a single outer goroutine that listens on a channel with a large buffer. Then we rate limit ourselves with that buffered channel

ellemouton, line 813: One worry with this is that we (in worst case sync from genesis) spin up like 700 goroutines (on mainnet). So another option to rate limit this a bit is to have a single outer goroutine that listens on a channel with a large buffer. Then we rate limit ourselves with that buffered channel

ellemouton: also gonna look into doing batch writes to the db instead of one by one

rescan.go
```diff
@@ -476,6 +484,51 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
+
+				return true
+			}
+
+			// If a block hash was specified, check if the
+			// notification is for that block.
+			return hash == ro.endBlock.Hash
+		}); err != nil {
+			return err
+		}
```
ellemouton, line 531: could maybe change this a bit so that if we are lagging by at least 1000, we start preemptively fetching filters (similarly to what is done for filter header syncing)

positiveblue (commented):
First time diving into the neutrino repository so I may be missing something. Better have an extra couple of eyes to check this but everything made sense to me. A couple of nits:

- Maybe some of the functions defined inside functions can be put outside or as a struct method so our routines do not get that long.

- Maybe we can only create the `persistChan` and write into it if `s.persistToDisk`.

> For example, on local regtest: before this commit it would take 7 seconds to sync 3000 blocks but with this commit, it takes 900ms

Thats a big improvement :fire: :fire: :fire: great work!

query.go
```diff
@@ -947,75 +949,96 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
...
+				// by the time we get here, it's been
+				// checked during header
+				// synchronization
+				s.chainParams.PowLimit,
+				s.timeSource,
+			); err != nil {
+				log.Warnf("Invalid block for %s "+
+					"received from %s -- ",
+					blockHash, peer)
+				fmt.Println(err)
```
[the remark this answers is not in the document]
ellemouton: oh haha yeah that was left over from debugging 🙈  will fix. thanks!

ellemouton:
Thanks for taking a look @positiveblue!!! Will address your comments soon 👍

ellemouton:
!lightninglabs-deploy mute

chappjc:
Gave it a test on mainnet and there were no more single cfilters requests, just big chunks of blocks, which is great.  Not always 1000, but never just 1.

ellemouton:
awesome! Thanks for testing @chappjc!

losh11:
hey guys, is anyone at lightninglabs looking into this?
would really appreciate it! resyncing takes forever, and makes user experience really, really bad.

query.go
Roasbeef, line 952: Prior logic would disconnect here, but now we'll continue... 

Seems worthy of a future spin off to propagate a "ban worthy" error back up to the main scheduler.

neutrino.go
```diff
@@ -575,11 +575,6 @@ type ChainService struct { // nolint:maligned
 	FilterCache *lru.Cache
 	BlockCache  *lru.Cache
 
-	// queryPeers will be called to send messages to one or more peers,
-	// expecting a response.
-	queryPeers func(wire.Message, func(*ServerPeer, wire.Message,
```
Roasbeef, line 580: 🙌

query.go
```diff
@@ -529,29 +367,153 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
...
+	// We're only interested in "cfilter" messages.
+	response, ok := resp.(*wire.MsgCFilter)
+	if !ok {
+		return query.Progress{
+			Finished:   false,
+			Progressed: false,
+		}
+	}
+
+	// If the request filter type doesn't match the type we were expecting,
```
Roasbeef, line 414: Few lines here and below look to run over.

query.go
```diff
@@ -529,29 +367,153 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
...
+	if filterHeader != curHeader {
+		return query.Progress{
+			Finished:   false,
+			Progressed: false,
+		}
+	}
+
+	// At this point the filter matches what we know about it, and we
+	// declare it sane. We send it into a channel to be processed elsewhere.
+	q.filterChan <- &filterResponse{
```
Roasbeef, line 475: Should select on quit both here and below.

query.go
```diff
@@ -851,31 +712,89 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
-		// If there are elements left to receive, the query failed.
-		if len(query.headerIndex) > 0 {
-			numFilters := query.stopHeight - query.startHeight + 1
-			numRecv := numFilters - int64(len(query.headerIndex))
-			log.Errorf("Query failed with %d out of %d filters "+
-				"received", numRecv, numFilters)
-			return
+		for {
+			select {
+			case resp, ok = <-q.filterChan:
```
Roasbeef, line 736: Hmm, I think better to not have the select assignment override the variable above? Also given that `ok` isn't used in the scope below. So can use an intermediate variable.

query.go
Roasbeef: Batch writes would be great, and shouldn't be too difficult to add in. 

Re the amount of goroutines: we can instead make a basic worker pool here. Or we only allow so many of them to be active at one time via a semaphore.

query.go
```diff
@@ -782,7 +784,33 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+		// GetCFilter calls from the caller sooner.
+		go func() {
+			var (
+				resp *filterResponse
+				ok   bool
+			)
+
+			for {
+				select {
+				case resp, ok = <-persistChan:
```
Roasbeef, line 805: I think would be cleaner to send the filter to persist, instead of relying on a chan close above as a signal to check a local closure variable.

rescan.go
```diff
@@ -409,19 +409,24 @@ func rescan(chain ChainSource, options ...RescanOption) error {
...
 
-	// We'll need to ensure that the backing chain has actually caught up to
-	// the rescan's starting height.
-	bestBlock, err := chain.BestBlock()
-	if err != nil {
-		return err
-	}
+	// waitFor is a helper closure that can be used to wait on block
+	// notifications until the given predicate returns true.
+	waitFor := func(predicate func(hash chainhash.Hash,
```
Roasbeef, line 414: Perhaps move into new function to cut down on the vertical distance a bit here?

Roasbeef (changes requested):

ellemouton:
Closing - gonna open 3 follow-up PRs 🤓

ellemouton:
Closing - gonna open 3 follow-up PRs 🤓

PR #240, opened 2021-12-31
query: end completed queries in sendTransaction before timeout

chappjc:
`(*ChainService).sendTransaction`  always times out even if each peer does respond.  This is because after receiving a `getdata` message in response to the outgoing `inv`, the `checkResponse` closure provided to `queryAllPeers` neglects to close the `peerQuit` channel.  As a result, `sendTransaction` always blocks for the full configured `BroadcastTimeout`.

This change adds the missing `close(peerQuit)` when a matching response (either a getdata or reject) is received.  I have tested this fix in our app, and sends no longer always time out when all peers respond with a getdata.

This change also makes a note in the `sendTransaction` function that the `if numReplied == len(rejections) {` test at the end is currently impossible to hit since `numReplied` would also need to be incremented when _rejections_ are recorded.  I did not fix this however since it would potentially start returning non-nil errors for harmless rejections, such as "already-in-mempool", "already have transaction", and "transaction already exists" (`pushtx.Mempool` and `pushtx.Confirmed`).  There are existing TODO comments from @wpaulino (wilmer) to this effect.  I did add a debug log though to help characterize what rejection messages are actually received in the wild.

Finally, this adds a comment in `(*blockManager).fetchFilterFromAllPeers`, which also appears to be lacking a `close(peerQuit)` in it's `queryAllPeers` check closure.  I added a debug log instead.  I'm pretty certain this fix is needed here too, but this code is only called from `(*blockManager).detectBadPeers`, and I haven't hit that in my tests yet today.

### DIFF
--- a/query.go
+++ b/query.go
@@ -1078,6 +1078,8 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 						)
 
 						numReplied++
+
+						close(peerQuit)
 					}
 				}
 
@@ -1096,6 +1098,13 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 					response, sp.Addr(),
 				)
 				rejections[*broadcastErr]++
+
+				log.Debugf("Transaction %v rejected by peer "+
+					"%v: code = %v, reason = %q", txHash,
+					sp.Addr(), broadcastErr.Code,
+					broadcastErr.Reason)
+
+				close(peerQuit)
 			}
 		},
 		append(
@@ -1109,7 +1118,7 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// transaction upon every block connected/disconnected.
 	if numReplied == 0 {
 		log.Debugf("No peers replied to inv message for transaction %v",
-			tx.TxHash())
+			txHash)
 		return nil
 	}
 
@@ -1118,7 +1127,7 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// it so we'll return the most rejected error between all of our peers.
 	//
 	// TODO(wilmer): This might be too naive, some rejections are more
-	// critical than others.
+	// critical than others. e.g. pushtx.Mempool and pushtx.Confirmed are OK.
 	//
 	// TODO(wilmer): This does not cover the case where a peer also rejected
 	// our transaction but didn't send the response within our given timeout
@@ -1126,7 +1135,7 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// threshold of rejections instead.
 	if numReplied == len(rejections) {
 		log.Warnf("All peers rejected transaction %v checking errors",
-			tx.TxHash())
+			txHash)
 
 		mostRejectedCount := 0
 		var mostRejectedErr pushtx.BroadcastError
--- a/sync_test.go
+++ b/sync_test.go
@@ -474,6 +474,29 @@ func testStartRescan(harness *neutrinoHarness, t *testing.T) {
 		}
 	}
 
+	// waitTx will poll for a transaction to appear on given node for up to
+	// 5 seconds.
+	waitTx := func(node *rpcclient.Client, hash chainhash.Hash) {
+		t.Helper()
+		exitTimer := time.NewTimer(5 * time.Second)
+		defer exitTimer.Stop()
+		for {
+			<-time.After(200 * time.Millisecond)
+
+			select {
+			case <-exitTimer.C:
+				t.Fatalf("Timeout waiting to see transaction.")
+			default:
+			}
+
+			if _, err := node.GetRawTransaction(&hash); err != nil {
+				continue
+			}
+
+			return
+		}
+	}
+
 	// Create another address to send to so we don't trip the rescan with
 	// the old address and we can test monitoring both OutPoint usage and
 	// receipt by addresses.
@@ -520,6 +543,9 @@ func testStartRescan(harness *neutrinoHarness, t *testing.T) {
 	if err != nil && !strings.Contains(err.Error(), "already have") {
 		t.Fatalf("Unable to send transaction to network: %s", err)
 	}
+	// SendTransaction does not know when the MsgTx was actually sent, only
+	// that a getdata request was received and a MsgTx queued to send.
+	waitTx(harness.h1.Client, authTx1.Tx.TxHash())
 	_, err = harness.h1.Client.Generate(1)
 	if err != nil {
 		t.Fatalf("Couldn't generate/submit block: %s", err)
@@ -565,6 +591,7 @@ func testStartRescan(harness *neutrinoHarness, t *testing.T) {
 	if err != nil && !strings.Contains(err.Error(), "already have") {
 		t.Fatalf("Unable to send transaction to network: %s", err)
 	}
+	waitTx(harness.h1.Client, authTx2.Tx.TxHash())
 	_, err = harness.h1.Client.Generate(1)
 	if err != nil {
 		t.Fatalf("Couldn't generate/submit block: %s", err)


arshbot (changes requested):
Thanks for the bugfix @chappjc!

Just did a quick review - is this PR in a reviewable state? There is commented out code and notes jotted about.

blockmanager.go
```diff
@@ -1879,6 +1879,9 @@ func (b *blockManager) fetchFilterFromAllPeers(
 				// and wait for the next response.
 				filterResponses[sp.Addr()] = gcsFilter
 
+				// close(peerQuit) // this peer responded. don't ask again
```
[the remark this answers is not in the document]
chappjc: This is what I described in the PR description.  I need input because I never hit this `fetchFilterFromAllPeers` function but from what I can tell it is missing this `close(peerQuit)`

blockmanager.go
```diff
@@ -1879,6 +1879,9 @@ func (b *blockManager) fetchFilterFromAllPeers(
...
+				log.Debugf("got gcs filter for %v from %v", blockHash, sp.Addr())
```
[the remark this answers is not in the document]
chappjc: I would remove this debug log entirely if it looks like the `close` was indeed required.  The debug log is added to help understand when this function is called and if it indeed makes repeated request to the same peer incorrectly after getting an initial response.

query.go
```diff
@@ -1096,6 +1098,18 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 					response, sp.Addr(),
 				)
 				rejections[*broadcastErr]++
+
+				// To ever get (numReplied == len(rejections)), it is necessary
+				// to also increment numReplied. However, this would be a
+				// breaking change since common innocuous errors such as Mempool
+				// and Confirmed do not presently cause a non-nil error return.
+				//
+				// numReplied++
```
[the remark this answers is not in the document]
chappjc: It is.  This is the bit in the PR description regarding @wpaulino (wilmer)'s comments.  The conditional below is impossible to hit afaict, and this commented change would seem to be the intention.  However, it's a breaking change since it would start returning errors in harmless situations such as the one described here and below w.r.t the `pushtx` types.

blockmanager.go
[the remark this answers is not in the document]
chappjc: Maybe I remove this blockManager change entirely since I am uncertain.  I was hoping that you or @ellemouton would have context on the `detectBadPeer` > `fetchFilterFromAllPeers` call path that could shed some light on this.  But at the end of the day, it's really just the `peer` fix that I was after.

query.go
```diff
@@ -1096,6 +1098,18 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
+
+				log.Debugf("Transaction %v rejected by peer %v: code = %v, reason = %q",
+					txHash, sp.Addr(), broadcastErr.Code, broadcastErr.Reason)
+
+				close(peerQuit)
```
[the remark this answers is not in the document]
arshbot: Ah I knew I was missing something! Thanks for the clarification :). And yes that works

query.go
```diff
@@ -1118,15 +1131,14 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// it so we'll return the most rejected error between all of our peers.
 	//
 	// TODO(wilmer): This might be too naive, some rejections are more
-	// critical than others.
+	// critical than others. e.g. pushtx.Mempool and pushtx.Confirmed are OK.
```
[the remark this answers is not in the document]
chappjc: To clarify, this was like saying `both pushtx.{Mempool,Confirmed} would be ok`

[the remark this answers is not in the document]
arshbot: gotcha! thank you for the clarification

arshbot (approved):
LGTM!

query.go
```diff
@@ -1096,6 +1098,18 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 					response, sp.Addr(),
 				)
 				rejections[*broadcastErr]++
+
+				// To ever get (numReplied == len(rejections)), it is necessary
+				// to also increment numReplied. However, this would be a
+				// breaking change since common innocuous errors such as Mempool
+				// and Confirmed do not presently cause a non-nil error return.
```
ellemouton: nit: wrap at 80 chars

query.go
```diff
@@ -1108,8 +1122,7 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// error as the reliable broadcaster will take care of broadcasting this
 	// transaction upon every block connected/disconnected.
 	if numReplied == 0 {
-		log.Debugf("No peers replied to inv message for transaction %v",
-			tx.TxHash())
+		log.Debugf("No peers replied to inv message for transaction %v", txHash)
```
ellemouton, line 1118: nit: leave as is to keep it wrapped at 80 chars

query.go
ellemouton: I think we should account for those cases then. Would you mind adding that? something like this:

```

func (c BroadcastErrorCode) IsCritical() bool {
	return c != Mempool && c != Confirmed	
}
```

Then, I think we can add the `numReplied++` here and then below we tally the rejections that are not critical and only return then we can keep the `if numReplied == numCriticalErrors` type logic

query.go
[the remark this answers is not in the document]
ellemouton: see my previous comment. I think we can just account for those non-critical error cases

query.go
```diff
@@ -1118,15 +1131,14 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
+	// critical than others. e.g. pushtx.Mempool and pushtx.Confirmed are OK.
 	//
 	// TODO(wilmer): This does not cover the case where a peer also rejected
 	// our transaction but didn't send the response within our given timeout
 	// and certain other cases. Due to this, we should probably decide on a
 	// threshold of rejections instead.
 	if numReplied == len(rejections) {
-		log.Warnf("All peers rejected transaction %v checking errors",
-			tx.TxHash())
+		log.Warnf("All peers rejected transaction %v checking errors", txHash)
```
ellemouton, line 1134: nit: can keep this wrapped

query.go
ellemouton: ok i see now that the overall effect of the PR is just to close the quit channel and to not include these comments. 

So then rather: squash all the commits into 1 that just does that. I will open an issue to fix the other things you have pointed out. Nice catch!

ellemouton (changes requested):
Nice catch @chappjc! 🚀 
And thanks for also pointing out the other issues in `sendTransaction`.

I think, instead of adding comments to point those out and then removing them, either we should fix them all in this PR or we should just squash this PR down to fixing just the channel close issue and then open up a new issue to fix the rest of the bugs. Seems from the overall code change here that you would be happy with option 2?

query.go
```diff
@@ -1118,15 +1124,14 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
 	//
 	// TODO(wilmer): This might be too naive, some rejections are more
-	// critical than others.
+	// critical than others. e.g. pushtx.Mempool and pushtx.Confirmed are OK.
 	//
 	// TODO(wilmer): This does not cover the case where a peer also rejected
 	// our transaction but didn't send the response within our given timeout
 	// and certain other cases. Due to this, we should probably decide on a
 	// threshold of rejections instead.
 	if numReplied == len(rejections) {
```
ellemouton, line 1133: this check actually also doesnt work cause `rejections` is a map... so this check only would work if each rejections error received was a unique one since this len will return the number of keys in the map

query.go
chappjc: Right sorry for the confusion there.  I kept the commits "unsquashed" so you could see the issues illustrated.  Will squash it down to one commit.

query.go
chappjc: Good point.  They'd all have to be unique errors, not so helpful.  Oh well.
Hopefully we see some rejected transactions logged to get an idea of when and if it's safe to throw and error back and remove the tx from the txn manager.

query.go
chappjc: re-wrapped, but both of these logs were under 80.  The cursor on this one says "81" at the end of this line, which is the limit.  No matter though.

query.go
chappjc: This commit is removed, so it's a moot point, but I think we must have our editors configured differently or something.  My comment wrapping extension puts it like this, and I am seeing it come right to the limit but not over.  My cursor at the beginning of the line says 1, and at the end of "Mempool" says 81.  Possibly we have different conventions?  :shrug:  Only pressing the nit so I can adhere to your conventions in the future...

chappjc:
Uh oh, random CI error after squashing.  https://github.com/lightninglabs/neutrino/runs/4853905111?check_suite_focus=true#step:4:160

```
 --- FAIL: TestNeutrinoSync (80.86s)
    --- FAIL: TestNeutrinoSync/start_long-running_rescan (39.21s)
        sync_test.go:535: Wrong number of relevant transactions. Want: 2, got: 1
```

chappjc:
Well darn I'm not sure what the test failure is about, but it must be related.  I have been running the unit tests prior to pushing so it's unfortunately a transient failure:

```
--- PASS: TestNeutrinoSync (115.56s)
    --- PASS: TestNeutrinoSync/initial_sync (3.10s)
    --- PASS: TestNeutrinoSync/one-shot_rescan (3.10s)
    --- PASS: TestNeutrinoSync/start_long-running_rescan (48.75s)
    --- PASS: TestNeutrinoSync/test_blocks_and_filters_in_random_order (0.15s)
    --- PASS: TestNeutrinoSync/check_long-running_rescan_results (58.73s)
PASS
coverage: 63.1% of statements
```

Looking closer now, but but I'm not familiar with this unit test and nothing is jumping out yet.

chappjc:
Looks like the tests are assuming the `SendTransaction` will return after the `MsgTx` was actually sent.  `sp.QueueMessageWithEncoding` is an async send, and if `SendTransaction` does not always time out, it's more likely that the subsequent `harness.h1.Node.Generate(1)` in the test will not include the transaction having not yet received it.

https://github.com/lightninglabs/neutrino/blob/53b628ce175698d112ddff3b60856902556c2fc2/sync_test.go#L516-L520

I don't think timeout of `SendTransaction` that we're fixing here was by-design, but it did introduce an artificial Sleep between queuing the `MsgTx` send and generating the next block that increased the likelihood of the mining node having received and accepted it for work prior to generating the next block.  AFAICT, there was still no confirmation of the `MsgTx` actually being sent, only queued, but there was an artificial delay before returning that made this more likely.

To keep the test working reliably on slower single core machines like CI probably uses, a `time.Sleep(5*time.Second)` would need to be inserted between `SendTransaction` and `Generate`, to reproduce what `SendTransaction` previously did.  Alternatively, a getrawtransaction loop on `harness.h1.Node` could be rigged.  I hate to suggest a `time.Sleep` in a test, but that would emulate the delay that is now gone.

chappjc:
That seems to have been the issue with the test.  The assumption was indeed that when `SendTransaction` returned without error that the miner node be ready to include that in a block, which worked when it always timed out because it gave the peer's out handler time to send it and the node time to accept it.  I hope this update to the test is acceptable.

sync_test.go
```diff
@@ -399,6 +399,23 @@ func testRescan(harness *neutrinoHarness, t *testing.T) {
 	}
 }
 
+func waitTx(t *testing.T, node *rpcclient.Client, hash chainhash.Hash) {
```
ellemouton, line 402: cool, this lgtm 👍 

could also move this into an embedded function in the test itself since it is only used there. And we could even copy the pattern of `wait.Predicate(...)` from the lnd repo and write it like this (up to you but its nice that it waits a bit before calling the func the first time. But i like that your way uses `time.NewTimer` instead of `time.After` 👍 ):
```
waitTx := func(node *rpcclient.Client, hash chainhash.Hash) error {
		exitTimer := time.After(5 * time.Second)
		for {
			<-time.After(200 * time.Millisecond)

			select {
			case <-exitTimer:
				return fmt.Errorf("tx not found after time out")
			default:
			}

			if _, err := node.GetRawTransaction(&hash); err != nil {
				continue
			}

			return nil
		}
	}
```

query.go
```diff
@@ -1096,6 +1098,11 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 					response, sp.Addr(),
 				)
 				rejections[*broadcastErr]++
+
+				log.Debugf("Transaction %v rejected by peer %v: code = %v, reason = %q",
+					txHash, sp.Addr(), broadcastErr.Code, broadcastErr.Reason)
+
+				close(peerQuit)
```
ellemouton, line 1105: nit: pls wrap at 80 chars

ellemouton (commented):
Nice catch on why that test was failing! This LGTM. Left a nit & an optional suggestion. I will rebase & open my pr with the other small fixes once this is in 👍

query.go
ellemouton: are you sure your IDE sets tab size to 8 chars?

chappjc: Oh, tabs. Noi use 4.  That explains that

sync_test.go
chappjc: Made it a closure, and incorporated the initial delay as you indicated was nice, but left it as a test helper.

ellemouton (commented):
LGTM pending 1 last line-wrap nit 🔥

query.go
```diff
@@ -1096,6 +1098,12 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 					response, sp.Addr(),
 				)
 				rejections[*broadcastErr]++
+
+				log.Debugf("Transaction %v rejected by peer %v: code = %v, "+
+					"reason = %q", txHash, sp.Addr(), broadcastErr.Code,
```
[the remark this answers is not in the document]
chappjc: oof these big tabs are killing me  :wink:

ellemouton (approved):

Roasbeef:
Needs a rebase!

Roasbeef (approved):
LGTM 🎋

query.go
```diff
@@ -1078,6 +1078,8 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 						)
 
 						numReplied++
+
+						close(peerQuit)
```
guggero, line 1082: I think this is wrong. Why would we close the connection to the peer if we just got an INV message? That means they want to know about the TX, so we send it to them. Wouldn't we also want to wait for their response before quitting?

ellemouton: Closing the `peerQuit` channel is just used to indicate that the peer has responded so that we free up the resource so we can use the peer connection for other calls. Closing this channel doesnt close the connection to the peer

guggero: Ah okay. But are we guaranteed to get further responses from that peer? E.g. if a reject message from that peer comes in later, do we get it in any case or depending on the timing (e.g. there's a chance we might not get it because we quit the peer here already)?

guggero: I just tried this out in combination with the `lnd` test `TestLightningWallet/btcwallet/neutrino:publish_transaction`. If I remove the `close(peerQuit)` in the INV message case, I correctly get the reject for a double spent transaction. If I leave it in there, the reject is never received.
There's also a bug in the result interpretation, so I'm going to open a PR anyway. Can discuss this further in that new PR.

PR #244, opened 2022-02-08
multi: small query fixes

ellemouton:
depends on #240 

Distinguish critical Tx broadcast errors from non-critical ones & fix a few other small bugs.

### DIFF
--- a/pushtx/error.go
+++ b/pushtx/error.go
@@ -47,6 +47,13 @@ func (c BroadcastErrorCode) String() string {
 	}
 }
 
+// IsCritical returns true if the error returned from the peer on broadcast
+// indicates that there was a problem that cased the transaction not to be
+// accepted and broadcast by them.
+func (c BroadcastErrorCode) IsCritical() bool {
+	return c != Mempool && c != Confirmed
+}
+
 // BroadcastError is an error type that encompasses the different possible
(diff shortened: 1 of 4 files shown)


query.go
```diff
@@ -1078,6 +1078,8 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 						)
 
 						numReplied++
+
+						close(peerQuit)
```
arshbot, line 1083: nit: I know this isn't your commit, but does it make sense to just `close(peerQuit)` at the end of this function, instead of at the end of each case statement?

pushtx/error.go
```diff
@@ -47,6 +47,13 @@ func (c BroadcastErrorCode) String() string {
 	}
 }
 
+// IsCritical returns true if the error returned from the peer on broadcast
+// indicates that there was a problem that cased the transaction not to be
```
arshbot, line 51: nit af

```suggestion
// indicates that there was a problem that caused the transaction not to be
```

arshbot (changes requested):
Love the bug fixes, overall looks good, just a few nits

ellemouton:
closing for now. not sure why the bot is not picking up that this is in draft mode

PR #245, opened 2022-02-08
multi: update to btcsuite/btcd/btcutil and btcec/v2

Roasbeef:
In this commit, we update the entire repo to the latest `btcec` and
`btcutil` modules. After this PR, our deps point to the new home of
btcutil, and also are able to utilize the latest `btcec` (which is only
really used in tests atm).

Once this PR lands, we can tag a new version, and then update the
`btcwallet` repo in a similar manner.


positiveblue (approved):
LGTM!

PR #247, opened 2022-02-23
Fix TX reject messages not being counted anymore

guggero:
This is a follow-up PR to fix an issue that https://github.com/lightninglabs/neutrino/pull/240 seems to have introduced.
When we close the `peerQuit` channel after receiving an INV message, we don't wait for any reject messages from that peer.

Instead we want to wait a bit longer after an INV message to make sure they don't respond with a reject message. Otherwise we would never detect a double spend TX with Neutrino (this was found through a unit test in lnd: https://github.com/lightningnetwork/lnd/pull/6285#issuecomment-1047332223)

Not sure if this proposed solution is any better, I'm definitely open for suggestions as I'm not too familiar with this code base.

### DIFF
--- a/query.go
+++ b/query.go
@@ -1058,8 +1058,9 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// We'll gather all of the peers who replied to our query, along with
 	// the ones who rejected it and their reason for rejecting it. We'll use
 	// this to determine whether our transaction was actually rejected.
-	numReplied := 0
-	rejections := make(map[pushtx.BroadcastError]int)
+	replies := make(map[int32]struct{})
+	rejections := make(map[int32]*pushtx.BroadcastError)
+	rejectCodes := make(map[pushtx.BroadcastErrorCode]int)
 
 	// Send the peer query and listen for getdata.
 	s.queryAllPeers(
@@ -1077,7 +1078,11 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 							tx, nil, qo.encoding,
 						)
 
-						numReplied++
+						// Peers might send the INV
+						// request multiple times, we
+						// need to de-duplicate them
+						// using a map.
+						replies[sp.ID()] = struct{}{}
 
 						close(peerQuit)
 					}
@@ -1097,7 +1102,8 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 				broadcastErr := pushtx.ParseBroadcastError(
 					response, sp.Addr(),
 				)
-				rejections[*broadcastErr]++
+				rejections[sp.ID()] = broadcastErr
+				rejectCodes[broadcastErr.Code]++
 
 				log.Debugf("Transaction %v rejected by peer "+
 					"%v: code = %v, reason = %q", txHash,
@@ -1116,7 +1122,7 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// If none of our peers replied to our query, we'll avoid returning an
 	// error as the reliable broadcaster will take care of broadcasting this
 	// transaction upon every block connected/disconnected.
-	if numReplied == 0 {
+	if len(replies) == 0 {
 		log.Debugf("No peers replied to inv message for transaction %v",
 			txHash)
 		return nil
@@ -1133,21 +1139,36 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 	// our transaction but didn't send the response within our given timeout
 	// and certain other cases. Due to this, we should probably decide on a
 	// threshold of rejections instead.
-	if numReplied == len(rejections) {
+	log.Debugf("Got replies from %d peers and %d rejections", len(replies),
+		len(rejections))
+	if len(replies) == len(rejections) {
 		log.Warnf("All peers rejected transaction %v checking errors",
 			txHash)
 
-		mostRejectedCount := 0
-		var mostRejectedErr pushtx.BroadcastError
-
-		for broadcastErr, count := range rejections {
+		// First, find the reject code that was returned most often.
+		var (
+			mostRejectedCount = 0
+			mostRejectedCode  pushtx.BroadcastErrorCode
+		)
+		for code, count := range rejectCodes {
 			if count > mostRejectedCount {
 				mostRejectedCount = count
-				mostRejectedErr = broadcastErr
+				mostRejectedCode = code
 			}
 		}
 
-		return &mostRejectedErr
+		// Then return the first error we have for that code (it doesn't
+		// really matter which one, as long as our error code parsing is
+		// correct).
+		for _, broadcastErr := range rejections {
+			if broadcastErr.Code == mostRejectedCode {
+				return broadcastErr
+			}
+		}
+
+		// We can't really get here unless something is totally wrong in
+		// the above error mapping code.
+		return fmt.Errorf("invalid error mapping")
 	}
 
 	return nil
--- a/query.go
+++ b/query.go
@@ -32,6 +32,11 @@ var (
 	// to a peer that previously failed because of a timeout.
 	QueryPeerCooldown = time.Second * 5
 
+	// QueryRejectTimeout is the time we'll wait after sending a response to
+	// an INV query for a potential reject answer. If we don't get a reject
+	// before this delay, we assume the TX was accepted.
+	QueryRejectTimeout = time.Second
+
 	// QueryNumRetries specifies how many times to retry sending a query to
 	// each peer before we've concluded we aren't going to get a valid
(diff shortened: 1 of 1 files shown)


query.go
```diff
@@ -1077,9 +1106,27 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
+						// peer it means they accepted
+						// the tx. If we get a reject,
+						// things are clear as well. But
+						// what if they are just slow to
+						// respond? We'll give them
+						// another bit of time to
+						// respond.
+						closeEventually(
+							peerQuit, replyChan,
+						)
```
chappjc, line 1129: Hmm, sorry if I broke this.  I must have been mistaken in thinking that our `inv` request would be completed after receiving a `getdata` response (this case) and handling it by queuing the msgtx send.

Is the issue that after this first `getdata` response case handles it, that there may be a subsequent `reject` that should be handled by this same closure?  My (apparently bad) assumption was that further responses to the tx data was handled elsewhere since that msgtx send was queued up for asynchronous sending.

Looks like the `QueryRejectTimeout` is meant to handle any such second response following handling of the first `getdata` response?  1 second is certainly better than the full 5 second broadcast timeout that was always hit before!  What you're doing with `replyChan` makes sense in that context.

Thanks for the proper resolution to the broadcast timeout issue I was trying to solve.  :)

guggero: > Hmm, sorry if I broke this.

No problem. I guess that's what our unit and integration tests in lnd are there for.

The actual problem is that publishing a TX over the p2p network isn't a synchronous process.
What happens is the following:
 1. The Neutrino client sends out an INV message with just the TXID of the so far unpublished TX ("hey y'all, have you ever heard of this TX?")
 2. Every peer will look at their mempool and if they don't find it, they'll answer back with a GetData message ("this is news to me, mind sending over that full TX?")
    a) Now it could be that the peer just doesn't respond in time. In that case we just ignore that peer
    b) It could also be that the peer already knows of the TX (for example we published it before). In that case it won't react at all (I think).
  3. We send the full TX out to each peer that requested it as an MsgTx ("here it is, plz forward!").
  4. The peer now either accepts the TX and forwards it or rejects it.
    a) If the peer accepts, we don't get any ACK or other response. After some time we just need to assume they're okay with it
    b) If the peer rejects, they send back a MsgReject and we know something's up.

From what I understand, all of this happens in the closure above. I'm not 100% sure my solution works in all cases. But at least it should reduce the time for a normal, successful publish from 5 seconds to 1 second.

Why 1 second and not 5 as before?
The assumption here is that if a peer responds quickly (e.g. before the overall timeout), it should respond with a reject reasonably quickly too. If a peer fails to respond within a second, we still have the threshold detection to recognize invalid transactions.

query.go
```diff
@@ -1077,7 +1078,11 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 							tx, nil, qo.encoding,
 						)
 
-						numReplied++
+						// Peers might send the INV
```
Roasbeef, line 1109: In practice, won't the closing of `peerQuit` below mean that for each peer this'll only ever be incremented once?

query.go
```diff
@@ -1082,9 +1104,23 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
+						// sent out the TX. If we never
+						// hear anything back from the
+						// peer it means they accepted
+						// the tx. If we get a reject,
+						// things are clear as well. But
+						// what if they are just slow to
+						// respond? We'll give them
+						// another bit of time to
+						// respond.
+						closeEventually(
```
Roasbeef, line 1127: I think for this to be safe, we want to wrap it in a `sync.Once` to avoid a panic if we end up closing the channel twice.

query.go
```diff
@@ -1202,15 +1214,31 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
...
-				return broadcastErr
-			}
-		}
+		return firstRejectWithCode(mostRejectedCode)
+	}
 
-		// We can't really get here unless something is totally wrong in
-		// the above error mapping code.
-		return fmt.Errorf("invalid error mapping")
+	// We did get some rejections, but not from all peers. Perhaps that's
```
Roasbeef, line 1220: This is intended to _not_ return an error if say only one of our peers rejects (since we wait longer now?) the transaction?

Roasbeef (commented):
Yay integration tests for catching this issue!

Did an initial review of the patch set. The one thing that comes to mind is that in practice, given that most of the network is `bitcoind` and the newer versions actually never send a reject message, then this may not make a difference in practice. Would be interesting to check (based on advertised versions) the % of nodes that actually send a reject message. 

In the end, I wager most neutrino nodes are connected to at least one btcd node given the lineage

query.go
Roasbeef: > b) If the peer rejects, they send back a MsgReject and we know something's up.

FWIW, the latest `bitcoind` nodes won't send a reject message, as result we can end up in a more murky state where we don't know if they already have it, or think it's invalid.

Roasbeef: > This last bit is what I figured wasn't handled in this closure but rather whatever is sending the tx, but it does make sense it would be need to be to associate this reject with this inv message.

Yeah it ends up happening in this closure. This essentially hooks into every message received by the peer, with us only acting on certain messages. 

This section of the code is actually using an _older_ instance of this query code. The newer (re-designed) query interface is currently only used in the block manager: https://github.com/lightninglabs/neutrino/blob/master/blockmanager.go#L1052

We have some issues that track the process of moving over to the new interface uniformly.

query.go
guggero: I think so, yes. That's why I removed the counter and replaced it with a map that just tracks if we got _any_ reply or not.

query.go
guggero: You're right. This is also a bit hard to understand, so I did a bit more refactoring and added a `delayedCloser` helper struct.

query.go
guggero: This is intended to return an error in slightly more cases. Before, we'd only return an error if all the peers responded with an error. Now we also return an error if only 50% of the peers reject.

That doesn't handle the case where only one peer rejects (the case where we're connected to a bunch of `bitcoind` nodes and just one `btcd`). Not sure what to do there, since we also don't want to open ourselves up to a malicious peer always rejecting and us failing because of that.
Perhaps we need to take an even closer look at the _type_ of error we get? Like if we get one "input already spent", we give that 100% weight as with other rejects we use a threshold?

Roasbeef: > Not sure what to do there, since we also don't want to open ourselves up to a malicious peer always rejecting and us failing because of that.

Yeah there's not really a great solution here...which I think was part of the rationale that the bitcoind devs used 

Looking at the type of error would certainly help here. Ultimately, I think our rebroadcaster would help to fill the gap here: as long as we keep rebroadcasting until the transaction has been _confirmed_, then things should _eventually_ be mined.

query.go
```diff
@@ -32,6 +32,11 @@ var (
 	// to a peer that previously failed because of a timeout.
 	QueryPeerCooldown = time.Second * 5
 
+	// QueryRejectTimeout is the time we'll wait after sending a response to
+	// an INV query for a potential reject answer. If we don't get a reject
+	// before this delay, we assume the TX was accepted.
+	QueryRejectTimeout = time.Second
```
Roasbeef, line 38: We _may_ need to tune this later to increase the value (thinking about high latency links, mobile network switching, international connections, etc)

query.go
```diff
@@ -1173,3 +1206,42 @@ func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) e
 
 	return nil
 }
+
+// delayedCloser is a struct that makes sure a subject channel is closed at some
+// point, either after a delay or immediately.
+type delayedCloser struct {
```
Roasbeef, line 1240: 👍

query.go
```diff
@@ -37,6 +37,12 @@ var (
 	// before this delay, we assume the TX was accepted.
 	QueryRejectTimeout = time.Second
 
+	// QueryInvalidTxThreshold is the threshold for the fraction of peers
+	// that need to respond to a TX with a code of pushtx.Invalid to count
+	// it as invalid, even if not all peers respond. This currently
+	// corresponds to 50% of peers that need to reject.
+	QueryInvalidTxThreshold float32 = 0.5
```
Roasbeef, line 44: Perhaps we should raise this to 60%? Lofty reasoning, but then this would require a super majority for us to give up (keeping in mind that we select _all_ our peers manually as they're outbound connections).

Roasbeef (approved):
LGTM ⛩

query.go
guggero: Good point. I added this to the query options, making it configurable.

query.go
guggero: Hard to pick a value indeed. I was mostly thinking of a scenario with just two peers, then one would be enough. But anyway, bumped it to 60% and also made it configurable through the query options.

PR #249, opened 2022-04-22
go.mod+neutrino.go: use NetAddressV2 where appropriate

Crypt-iQ:
This will allow a caller that uses the neutrino library to receive AddrV2 messages and connect out to tor v3 addresses.


neutrino.go
```diff
@@ -1161,6 +1222,12 @@ func (s *ChainService) addrStringToNetAddr(addr string) (net.Addr, error) {
 		}
 	}
 
+	// Tor addresses cannot be resolved to an IP, so just return onionAddr
+	// instead.
+	if strings.HasSuffix(host, ".onion") {
```
Roasbeef, line 1227: 👍

Roasbeef (commented):
Slim diff, though I thought we would've needed some changes to the [newAddressFunc](https://github.com/lightninglabs/neutrino/blob/master/neutrino.go#L755)?

go.mod
```diff
@@ -1,16 +1,17 @@
 module github.com/lightninglabs/neutrino
 
 require (
-	github.com/btcsuite/btcd v0.22.0-beta.0.20220207191057-4dc4ff7963b4
-	github.com/btcsuite/btcd/btcec/v2 v2.1.0
-	github.com/btcsuite/btcd/btcutil v1.1.0
+	github.com/btcsuite/btcd v0.22.0-beta.0.20220413172512-bf64c8bdbbbf
```
Roasbeef, line 4: CI looks to be failing, I think we need to bump this further? 

Or we need to update everything to target go 1.18 as btcd is using _only_ those new directives?

Crypt-iQ:
> Slim diff, though I thought we would've needed some changes to the [newAddressFunc](https://github.com/lightninglabs/neutrino/blob/master/neutrino.go#L755)?

It calls addrStringToNetAddr which is all that needed to be changed (convert *NetAddressV2 into net.Addr)

Roasbeef (approved):
LGTM 🐘

ellemouton (approved):
LGTM!

Issue #211, opened 2021-01-09
Header doesn't pass sanity check: block target difficulty of 0000000000000000000eb1560000000000000000000000000000000000000000 is higher than max of 0000000000000000000000000000000000000000000000000000000000000000 -- disconnecting peer


JavierRSobrino:
My LND is unable to sync using Neutrino. I get this error for ever and ever:

```
2021-01-09 21:35:19.326 [INF] BTCN: Syncing to block height 665317 from peer 13.91.246.35:8333
2021-01-09 21:35:19.326 [INF] BTCN: Fetching set of headers from tip (height=659260) from peer 13.91.246.35:8333
2021-01-09 21:35:22.294 [WRN] BTCN: Header doesn't pass sanity check: block target difficulty of 0000000000000000000eb1560000000000000000000000000000000000000000 is higher than max of 0000000000000000000000000000000000000000000000000000000000000000 -- disconnecting peer

```
Obviusly something is wrong in the code, as the maximum difficulty cannot be zero.

lnd version 0.12.0-beta.rc5 commit=v0.12.0-beta.rc5

Relevant `lnd.conf` is:

```
bitcoin.node=neutrino
routing.assumechanvalid=1
neutrino.addpeer=btcd-mainnet.lightning.computer
neutrino.addpeer=btcd0.lightning.engineering
neutrino.addpeer=neutrino.shock.network
neutrino.addpeer=mainnet1-btcd.zaphq.io
neutrino.addpeer=mainnet2-btcd.zaphq.io
neutrino.addpeer=mainnet3-btcd.zaphq.io
neutrino.addpeer=mainnet4-btcd.zaphq.io
neutrino.banduration=1s
neutrino.banthreshold=100000
neutrino.maxpeers=256
neutrino.feeurl=https://nodes.lightning.computer/fees/v1/btc-fee-estimates.json
```

guggero:
Does the problem persist? Do you also get the error if you comment out the first `neutrino.addpeer`, because `13.91.246.35` is `btcd-mainnet.lightning.computer`.
I assume you have `bitcoin.mainnet=1` in the config as well? Or is this your complete config file?

This sounds like it could be related to a block reorg?
It looks to me like `calcNextRequiredDifficulty` returns `0`: https://github.com/lightninglabs/neutrino/blob/master/blockmanager.go#L2826

guggero:
@JavierRSobrino I looked at the code again and tried a few things to reproduce, but no luck.
[...]

Issue #255, opened 2022-08-02
Regtest: "block target difficulty is higher than max"


amanzag:
Whenever I try to use neutrino in regtest mode, I run into this error:
```
I/GoLog   (29209): 2022-08-02 14:49:05.236 [DBG] BTCN: Difficulty retarget at block height 4032
I/GoLog   (29209): 2022-08-02 14:49:05.236 [DBG] BTCN: Old target 207fffff (7fffff0000000000000000000000000000000000000000000000000000000000)
I/GoLog   (29209): 2022-08-02 14:49:05.236 [DBG] BTCN: New target 201fffff (1fffff0000000000000000000000000000000000000000000000000000000000)
I/GoLog   (29209): 2022-08-02 14:49:05.236 [DBG] BTCN: Actual timespan 33h35m0s, adjusted timespan 84h0m0s, target timespan 336h0m0s
I/GoLog   (29209): 2022-08-02 14:49:05.236 [WRN] BTCN: Header doesn't pass sanity check: block target difficulty of 7fffff0000000000000000000000000000000000000000000000000000000000 is higher than max of 1fffff0000000000000000000000000000000000000000000000000000000000 -- disconnecting peer
```
There must be some flaw in the difficulty calculation logic here: https://github.com/lightninglabs/neutrino/blob/master/blockmanager.go#L2795-L2834

Related issue (but I think it belongs here): https://github.com/btcsuite/btcd/issues/1819

Roasbeef:
Are you able to reproduce the same thing w/ simnet? IIRC, I think btcd will do difficulty retargets for regtest, while bitcoind never does. This is what historically led to the creation of simnet (which does the retargets).

amanzag:
I think you're right. I can't reproduce with simnet and that's probably because I'm running neutrino against a bitcoind which, as you say, doesn't do retargets, as can be seen here:
https://github.com/bitcoin/bitcoin/blob/master/src/chainparams.cpp#L407

Shouldn't btcd/neutrino disable retargets for regtest as well?

amanzag:
I created a pull request https://github.com/lightninglabs/neutrino/pull/256 that fixes the issue.
It's probably not the right way of fixing it, but if you point me in the right direction I can change it.

guggero:
[...]

PR #256, opened 2022-08-03
Disable retargeting on regtest to preserve compatibility with bitcoind

amanzag:


buck54321:
I think this could possibly resolve #211 and #255. @roasbeef, can we get some action on this?

blockmanager.go
```diff
@@ -2736,6 +2736,11 @@ func (b *blockManager) checkHeaderSanity(blockHeader *wire.BlockHeader,
 func (b *blockManager) calcNextRequiredDifficulty(newBlockTime time.Time,
 	reorgAttempt bool) (uint32, error) {
 
+	// bitcoind doesn't do retargeting on regtest, we do the same to ensure compatibility
```
guggero, line 2739: This looks a bit hacky indeed... `bitcoind` solved it by adding another boolean to the chain params: https://github.com/bitcoin/bitcoin/blob/e695d8536e534f1e59de4a62a0d87a93e7a73456/src/kernel/chainparams.cpp#L415
Perhaps we should do the same in `btcd`? Then we could solve it for both Neutrino and `btcd`.

guggero (changes requested):

blockmanager.go
guggero: I have a PR up in `btcd` that can be re-used here once merged: https://github.com/btcsuite/btcd/pull/1985

chappjc: Missed that, Great!

PR #263, opened 2022-11-29
blockmanager.go: use btcd libs to validate headers

Crypt-iQ:
This uses btcd's HeaderCtx and ChainCtx interfaces to be able to validate headers, both contextually and context-free. This allows neutrino to call blockchain.CheckBlockHeaderContext and blockchain.CheckBlockHeaderSanity. Also included is a check to assert that when neutrino receives a p2p headers message, that each header connects to the previous one.

TODO:

- [x] sync
- [ ] tests
- [x] depends on https://github.com/btcsuite/btcd/pull/1931

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -2383,6 +2383,33 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 	// atomically in order to improve peformance.
 	headerWriteBatch := make([]headerfs.BlockHeader, 0, len(msg.Headers))
 
+	// Explicitly check that each header in msg.Headers builds off of the
+	// previous one.
+	var (
+		lastHeader chainhash.Hash
+		emptyHash  chainhash.Hash
+	)
+
+	for _, blockHeader := range msg.Headers {
+		blockHash := blockHeader.BlockHash()
+
+		if lastHeader == emptyHash {
+			// If we haven't yet set lastHeader, set it now.
+			lastHeader = blockHash
+		} else {
+			// Ensure that blockHeader.PrevBlock matches
+			// lastHeader.
+			if blockHeader.PrevBlock != lastHeader {
+				log.Warnf("Headers received from peer don't " +
+					"connect")
+				hmsg.peer.Disconnect()
+				return
+			}
+
+			lastHeader = blockHash
+		}
+	}
+
 	// Process all of the received headers ensuring each one connects to
 	// the previous and that checkpoints match.
 	receivedCheckpoint := false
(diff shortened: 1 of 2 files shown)


blockmanager.go
```diff
@@ -2383,6 +2383,33 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 	// atomically in order to improve peformance.
 	headerWriteBatch := make([]headerfs.BlockHeader, 0, len(msg.Headers))
 
+	// Explicitly check that each header in msg.Headers builds off of the
```
Roasbeef, line 2386: Can be moved out into a some helper function. Also would have though the functions we exported from `btcd` would handle it so we don't need to do it again?

blockmanager.go
```diff
@@ -2411,8 +2438,10 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 		prevNode := prevNodeEl
 		prevHash := prevNode.Header.BlockHash()
 		if prevHash.IsEqual(&blockHeader.PrevBlock) {
+			prevNodeHeight := prevNode.Height
+			prevNodeHeader := prevNode.Header
 			err := b.checkHeaderSanity(blockHeader, maxTimestamp,
-				false)
+				false, prevNodeHeight, &prevNodeHeader)
```
Roasbeef, line 2444: Can fix line folding here.

blockmanager.go
```diff
@@ -2520,8 +2549,28 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
...
-				err = b.checkHeaderSanity(reorgHeader,
-					maxTimestamp, true)
+				// We have to get the parent's height and
+				// header to be able to contextually validate
+				// this header.
+				prevNodeHeight := backHeight + uint32(j)
+
+				var prevNodeHeader *wire.BlockHeader
+
+				if i+j == 0 {
```
Roasbeef, line 2559: So `i && j == 0`?

blockmanager.go
```diff
@@ -2706,29 +2755,44 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
...
 		return err
 	}
-	// Ensure the block time is not too far in the future.
-	if blockHeader.Timestamp.After(maxTimestamp) {
-		return fmt.Errorf("block timestamp of %v is too far in the "+
-			"future", blockHeader.Timestamp)
-	}
-	return nil
+
+	err = blockchain.CheckBlockHeaderSanity(
```
Roasbeef, line 2790: Can just do `return` here.

Roasbeef:
Can be rebased now!

guggero:
Replaced by https://github.com/lightninglabs/neutrino/pull/283, addressed all open review comments there.

PR #268, opened 2023-02-23
cache: add deletion and iteration methods

yyforyongyu:
This commit adds three methods, `Delete`, `LoadAndDelete`, and `Range` to the lru cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ breakpoints.txt
 
 # coverage output
 coverage.txt
+
+# go workspace
+go.work
--- a/cache/lru/lru.go
+++ b/cache/lru/lru.go
@@ -7,9 +7,6 @@ import (
 	"github.com/lightninglabs/neutrino/cache"
 )
 
-// elementMap is an alias for a map from a generic interface to a list.Element.
-type elementMap[K comparable, V any] map[K]V
-
 // entry represents a (key,value) pair entry in the Cache. The Cache's list
 // stores entries which let us get the cache key when an entry is evicted.
 type entry[K comparable, V cache.Value] struct {
@@ -33,7 +30,7 @@ type Cache[K comparable, V cache.Value] struct {
 
 	// cache is a generic cache which allows us to find an elements position
 	// in the ll list from a given key.
-	cache elementMap[K, *Element[entry[K, V]]]
+	cache syncMap[K, *Element[entry[K, V]]]
 
 	// mtx is used to make sure the Cache is thread-safe.
 	mtx sync.RWMutex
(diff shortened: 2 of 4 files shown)


cache/lru/lru.go
```diff
@@ -166,3 +166,52 @@ func (c *Cache[K, V]) Len() int {
...
+
+// Range iterates the cache.
+//
+// NOTE: the `visitor` is wrapped inside the Cache's mutex.
+func (c *Cache[K, V]) Range(visitor func(K, V) bool) {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	for k, v := range c.cache {
+		if !visitor(k, v.Value.value) {
```
Roasbeef, line 213: Hmm, one thing that pops up here is that since Go doesn't have reentrant locks, then if `visitor` tries to call any of these methods, it'll deadlock.

yyforyongyu: yeah I also think this design might be dangerous. Some thoughts,
- maybe we could copy the `cache` inside a lock above and then iterate? Meanwhile the `cache` could be updated tho by another goroutine tho.
- Remove the lock in `Range`, this means the caller needs to remember using the locks.
- Replace `cache` with a sync map, lower performance on write, but no more worries on locks.

Roasbeef: > Replace cache with a sync map, lower performance on write, but no more worries on locks.

I think this could work, w/o forcing us to reimplement the wheel too much.

yyforyongyu: Cool, used `SyncMap` with code copied from [here](https://github.com/lightningnetwork/lnd/blob/master/lnutils/sync_map.go). We may want to create a repo to host these commonly used utilities someday.

The `mutex` used by the `cache` is still needed tho, to update its internal list and size.

Roasbeef: > We may want to create a repo to host these commonly used utilities someday.

Alternatively, assuming we keep `lnutils` w/ a slim set of deps, we can make that into a sub-module so other repos can call into it w/o having to worry about weird import cycle.

.gitignore
```diff
@@ -28,3 +28,6 @@ breakpoints.txt
 
 # coverage output
 coverage.txt
+
+# go workspace
+go.work
```
Roasbeef, line 33: 😲

Do you use the go work space stuff? I read the docs a few times, but could never fully get my head around the utility....

Roasbeef (approved):
LGTM 🦄

.gitignore
yyforyongyu: yeah I used it cause vim kept complaining about the `cache` module, and my `go.work` is simply,

```
go 1.19

use ./cache
```

Meanwhile I don't understand the docs(they need to be improved tbh)😂, only trial and error got it working.

Issue #253, opened 2022-06-07
unban peer


bitbandi:
If a peer is banned for (48hours) becase he does bad things. But somethimes bans are unfair (btc daemon misconfigure, etc). there is no possible to unban the fixed peer.

Please add some func for this. and for lnd/lncli rpc too.

Roasbeef:
We have a gRPC server for neutrino in version 0.15 of lnd, which IMO should actually be moved to _this_ project, which'll allow us to easily add calls like this.

Chinwendu20:
plesae I would like to work on this

Chinwendu20:
I really do not think banning a peer does anything to it, it just disconnects it and puts it in database that is not even consulted when we want to reconnect to a peer. Please let me know if I am mistaken.

ellemouton:
> database that is not even consulted when we want to reconnect to a peer.

I dont think that is true. See all the calls to `(*ChainService) IsBanned`

Chinwendu20:
I don't think `IsBanned` has anything to do with reconnection?
https://github.com/lightninglabs/neutrino/blob/3275e33e5504daf33393ca8c888c8111a0595823/neutrino.go#L1102-L1124

ellemouton:
im saying look at where `IsBanned` is called from. There a a couple of call sites. We call this to check if we should connect to a peer or if we should disconnect from them if they connect to us

Chinwendu20:
Oh thanks you are totally right. I was looking at the logs and it seemed we connect to already banned peers but after following the code after you pointed it out, it seems in some cases we let the peers connect then in onconnected function disconnect them.

PR #270, opened 2023-04-18
neutrino: Added unban peer feature

Chinwendu20:
Changes:

- Added unbanpeer method to chainservice.
- Added unbanipnet method to banstore

Associated PR in LND: https://github.com/lightningnetwork/lnd/pull/7606 (With detailed explanation on how to test with LND)
Fixes issue: https://github.com/lightninglabs/neutrino/issues/253


neutrino.go
```diff
@@ -1078,6 +1078,33 @@ func (s *ChainService) BanPeer(addr string, reason banman.Reason) error {
 	return s.banStore.BanIPNet(ipNet, reason, BanDuration)
 }
 
+// UnBanPeer connects and unban's a previously banned peer.
+func (s *ChainService) UnBanPeer(addr string) error {
+	log.Warnf("UnBanning peer %v", addr)
+
+	defer func() {
+
+		if sp := s.PeerByAddr(addr); sp != nil {
```
Roasbeef, line 1087: I don't think it's necessary to also trigger a connection request in the background. Instead it should be enough to simply unban the peer. 

This also has a race condition: if this goroutine execute before the logic below, then we'll connect out to the peer but then tear down the connection as it's from a banned peer.

Roasbeef (changes requested):

neutrino.go
Chinwendu20: Thanks  a lot for the review but I would like to please clarify what goroutine that you are referring to. As I did not include a goroutine in the function. Are you saying that I should remove from line 1085-1098?

Roasbeef: Ah, I scanned over it quickly and didn't realize this was actually just a `defer`. Since the unban call is sync, this should be ok.

Roasbeef:
Has a linter error failure: 
```
neutrino.go:1085: unnecessary leading newline (whitespace)
	defer func() {
```

Chinwendu20:
Hello @Roasbeef I have fixed that

neutrino.go
```diff
@@ -1078,6 +1078,32 @@ func (s *ChainService) BanPeer(addr string, reason banman.Reason) error {
...
+func (s *ChainService) UnBanPeer(addr string) error {
+	log.Warnf("UnBanning peer %v", addr)
+
+	defer func() {
+		if sp := s.PeerByAddr(addr); sp != nil {
+			tcpAddr, err := s.addrStringToNetAddr(addr)
+			if err != nil {
+				log.Warnf("unable to lookup IP for "+
+					"%v: %v", addr, err)
+			}
```
positiveblue, line 1091: I think we should return when we get an error so we do not add it to the `connManager`

banman/store.go
```diff
@@ -66,6 +66,9 @@ type Store interface {
 
 	// Status returns the ban status for a given IP network.
 	Status(*net.IPNet) (Status, error)
+
+	// UnBanIPNet unbans previously banned peer
+	UnBanIPNet(ipNet *net.IPNet) error
```
positiveblue, line 71: nit: `unbanIPNet`? I am not an english speaker but I think it is the same as with the word "unable"?

positiveblue (changes requested):
I left one question about adding the `conn` to the `connManager` when we fail to get the  `tcpAddr` but the rest looks good  :rocket: 

Also, we should test it. Maybe we can simply update the `banPeer` test.

neutrino.go
Chinwendu20: Thank you so much for this. Fixed it.

banman/store.go
positiveblue: yes, I meant capitalizing only the first letter `UnbanIPNet` instead of `UnBanIPNet`. Like [here](https://github.com/bitcoin/bitcoin/blob/3132ec64d9e0f7af88e61ac35807f44315a5ca96/src/banman.h#L77)

Chinwendu20: I have effected the change @positiveblue

banman/store.go
```diff
@@ -66,6 +66,9 @@ type Store interface {
 
 	// Status returns the ban status for a given IP network.
 	Status(*net.IPNet) (Status, error)
+
```
Roasbeef, line 69: In commmit message: `banmen` -> `banman`

Roasbeef (changes requested):
I think this is only missing a test now as jorid noted.

banman/store.go
Chinwendu20: Oh okay thanks.

Chinwendu20:
Okay I would write the test.

Chinwendu20:
Hello, the commit issues here, I do not know how to fix them, so I would be creating a new PR for this.

PR #271, opened 2023-04-28
gottafixit 🤠

cubancorona:


blockmanager.go
```diff
@@ -2620,6 +2620,13 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 					hmsg.peer.Disconnect()
 					return
 				}
+				// Make sure headers being processed are connecting
```
guggero, line 2623: nit: can you please add a newline before the comment and after your added block of code? Also, sentences should always be complete, including punctuation.

blockmanager.go
```diff
@@ -2620,6 +2620,13 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
...
+				prevBlockHashInReorganizationList := b.reorgList.Back().Header.BlockHash()
+				if !reorgHeader.PrevBlock.IsEqual(&prevBlockHashInReorganizationList) {
```
guggero, line 2625: We try to keep lines at 80 characters max length (using 8 spaces for tabs). Maybe choose a shorter name?

guggero (commented):
Thanks for the PR. What's the context here? Did you run into this in the wild?

cubancorona:
> Thanks for the PR. What's the context here? Did you run into this in the wild?

No.  I simulated a block collision (two new valid blocks mined at approx. the same time, and at same height) using real block data (i.e., where an actual block race occurred) and then appended a BS chain to the de facto abandoned fork to show that neutrino would follow my garbage chain...

blockmanager.go
cubancorona: Could you assist?  I do not have a working dev env.

blockmanager.go
cubancorona: Could you assist?  I agree and many thanks for the review.

blockmanager.go
guggero: What do you mean by assist? Take over the PR? I can't push changes to your branch...

cubancorona: Yes, that's fine!

guggero:
Replaced by https://github.com/lightninglabs/neutrino/pull/272.

Issue #69, opened 2018-07-10
neutrino: implement BIP 151


Roasbeef:
Currently `neutrino` is used primarily within `lnd` as an optional chain backend. Within the current LN protocol, all traffic is encrypted by default. However, our Bitcoin p2p traffic would still currently be in plaintext and unauthenticated. In order to make our traffic a bit more uniform, we should implement [BIP 151](https://github.com/bitcoin/bips/blob/master/bip-0151.mediawiki) and switch to it for supported nodes. 

I'm tracking the issue here, but a bulk of the code will likely end up living in the `peer` package of `btcsuite/btcd`.

PR #273, opened 2023-05-05
query+neutrino: use query dispatcher for GetBlock and GetCFilter

ellemouton:
This PR refactors the `GetBlock` and `GetCFilter` methods so that they use the
work dispatcher for their queries instead of using the old `queryPeers` function. 
The function is now removed bringing us one step closer to removing all query 
logic from the main package.


query_test.go
```diff
@@ -286,57 +288,59 @@ func TestBlockCache(t *testing.T) {
...
-		}
+		errChan := make(chan error, 1)
+		defer close(errChan)
 
-		if len(getData.InvList) != 1 {
-			t.Fatalf("expected 1 element in inv list, found %v",
-				len(getData.InvList))
-		}
+		require.Len(t, reqs, 1)
+		require.IsType(t, &wire.MsgGetData{}, reqs[0].Req)
```
Roasbeef, line 306: TIl about `IsType`!

query.go
```diff
@@ -888,7 +888,7 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
 
 	// We didn't get the filter from the DB, so we'll try to get it from
 	// the network.
-	query, err := s.prepareCFiltersQuery(blockHash, filterType, options...)
```
Roasbeef, line 889: Commit message says `queue` instead of `query`.

query.go
```diff
@@ -888,7 +888,7 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
+	q, err := s.prepareCFiltersQuery(blockHash, filterType, options...)
```
Roasbeef: `filterQuery`?

query.go
```diff
@@ -268,20 +268,20 @@ func MaxBatchSize(maxSize int64) QueryOption {
 // Timeout functional option. The NumRetries option is set to 1 by default
 // unless overridden by the caller.
 func (s *ChainService) queryAllPeers(
-	// queryMsg is the message to broadcast to all peers.
+// queryMsg is the message to broadcast to all peers.
```
Roasbeef: I think we can remove those now along side the one below?

query.go
```diff
@@ -872,25 +907,28 @@ func (s *ChainService) GetCFilter(blockHash chainhash.Hash,
...
 		return nil, err
 	}
 
+	qo := defaultQueryOptions()
+	qo.applyQueryOptions(options...)
+
 	// We didn't get the filter from the DB, so we'll try to get it from
 	// the network.
-	q, err := s.prepareCFiltersQuery(blockHash, filterType, options...)
+	q, err := s.prepareCFiltersQuery(
```
Roasbeef, line 766: `filterQuery`

query.go
```diff
@@ -382,168 +382,6 @@ checkResponses:
 	}
 }
 
-// queryChainServicePeers is a helper function that sends a query to one or
-// more peers of the given ChainService, and waits for an answer. The timeout
-// for queries is set by the QueryTimeout package-level variable or the Timeout
-// functional option.
-func queryChainServicePeers(
```
Roasbeef, line 387: 👍

query.go
ellemouton: fixed 🤓

query.go
ellemouton: changed!

query.go
ellemouton: erg soz my formatter does its own thing with these commented parameters 🙃 

we still need the `queryAllPeers` func for the case where we want a certain message to be sent to each of our peers instead of just current best one for things like block header syncing & checking. The work manager doesnt currently support this "send to all" functionality.

guggero (commented):
Very nice improvement! Did a first pass just to load some context, didn't see anything that stood out as being problematic. I assume we'll battle test all these improvements mostly indirectly through the lnd itests?

neutrino.go
```diff
@@ -666,6 +666,7 @@ type ChainService struct { // nolint:maligned
 	broadcaster          *pushtx.Broadcaster
 	banStore             banman.Store
 	workManager          *query.WorkManager
+	queryDispatcher      query.Dispatcher
```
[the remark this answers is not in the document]
ellemouton: and cant switch out the current `*query.WorkManager` for the `query.Disptcher` interface since the interface doesnt have the `Start()` and `Stop()` methods.... although - I could just add those to the interface? Lemme know what you think

ellemouton:
> I assume we'll battle test all these improvements mostly indirectly through the lnd itests?

Ah, that's a good point - I should open an LND PR pointing to all these changes so that the neutrino itest suite can be run 👍

neutrino.go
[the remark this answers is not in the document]
ellemouton: cool I added a commit that hopefully makes this nicer. Lemme know what you think :)

guggero (approved):
Looks good to me :tada:

query.go
```diff
@@ -576,29 +576,179 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
...
+			"elements. Consider increasing the cache size...",
+			numFilters)
+	}
+
+	if q.cs.persistToDisk {
+		err = q.cs.FilterDB.PutFilter(
+			&response.BlockHash, filter, dbFilterType,
+		)
+		if err != nil {
+			log.Warnf("Couldn't write filter to filterDB: %v", err)
```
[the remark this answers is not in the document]
ellemouton: yeah good point!

The reason I'm hesitant to change this is cause the DB write actually _really_ slows down the process of fetching the filters from the network and so in the next PR, I basically completely divorce these two things so that the work manager doesnt get slowed down by the db writes at all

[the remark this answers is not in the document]
guggero: Ah okay. Fine with me to just ignore it until the follow-up PRs. But I think in general error handling in these async calls could be something to think about.

query.go
```diff
@@ -576,29 +576,179 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
...
+
+// handleResponse validates that the cfilter response we get from a peer is
+// sane given the getcfilter query that we made.
+func (q *cfiltersQuery) handleResponse(req, resp wire.Message,
+	_ string) query.Progress {
+
+	// The request must have been a "getcfilters" msg.
+	request, ok := req.(*wire.MsgGetCFilters)
+	if !ok {
+		return query.Progress{
```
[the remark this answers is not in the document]
ellemouton: good idea - updated :)

ellemouton:
hmmm - has anyone seen this test error before? i'm wondering if it is a flake or related to this change... it passes locally (and on prior pushes to this branch)

positiveblue (approved):
LGTM :100: 


I saw that the unit tests were failing in [#69](https://github.com/ellemouton/lnd/pull/69) but they all passed locally.

query.go
```diff
@@ -574,29 +421,155 @@ func (s *ChainService) putFilterToCache(blockHash *chainhash.Hash,
...
+			log.Warnf("Couldn't write filter to filterDB: %v", err)
+		}
+
+		log.Tracef("Wrote filter for block %s, type %d",
+			&response.BlockHash, dbFilterType)
+	}
+
+	// We delete the entry for this filter from the headerIndex to indicate
+	// that we have received it.
+	delete(q.headerIndex, response.BlockHash)
```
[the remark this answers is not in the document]
ellemouton: yeah that sounds correct 🎉

ellemouton:
> I saw that the unit tests were failing in https://github.com/ellemouton/lnd/pull/69 but they all passed locally.

ah - let me open one that only points to this and not the others. Just to isolate it a bit.

ellemouton:
ok, I've updated https://github.com/ellemouton/lnd/pull/69 to just point to this PR. fingers crossed for unit tests 🤞

ellemouton:
erg - realised we will need to make an update in btcwallet aswell since it uses the old WorkManager struct which is now an interface 🙃

ellemouton:
alrighty! after searching for waaaayyy too many hours - I found the issue 🙈 

The issue is that: the worker manager, as it currently is, doesnt have any form of "maximum number of retries" and so if any query fails, it just reschedules the job so that it can be tried again later. This is not the behaviour that the rescan code expects especially during a re-org. The rescan function wants the query to error so that it can retry itself. The reason we want the calling function to control retries is cause perhaps the query failed due to a re-org and so the actual query itself needs to be changed. So what was happening in the tests that were failing is: a  query was started to get a filter for a certain block hash, then a re-org happens, and then the query manager just keeps trying the same query with the old block hash forever. 

So now I have added a commit that allows the caller to set the maximum number of retries. 

gonna re-request review since a new commit was added (it is the second commit)

query/workmanager.go
```diff
@@ -321,11 +322,48 @@ Loop:
 				}
 
 			// If the query ended with any other error, put it back
-			// into the work queue.
+			// into the work queue if it has not reached the
+			// maximum number of retries.
 			case result.err != nil:
 				// Punish the peer for the failed query.
 				w.cfg.Ranking.Punish(result.peer.Addr())
```
ellemouton, line 335: pre-existing but wondering if we should remove this Punish here and let the caller handle it instead since this could just be erroring cause we requested something that is not there anymore (like in a re-org)

query/workmanager.go
```diff
@@ -321,11 +323,50 @@ Loop:
...
+						result.peer.Addr(), result.err)
+
+					// Decrement the number of queries
+					// remaining in the batch.
+					batch.rem--
+					log.Tracef("Remaining jobs for batch "+
+						"%v: %v ", batchNum, batch.rem)
+
+					if batch.rem > 0 {
+						break
```
guggero, line 362: `break` in a nested loop/select is always a bit tricky to understand. Wouldn't this have the same effect as `continue Loop` and should therefore be replaced with that to make the behavior more clear?

query/workmanager_test.go
```diff
@@ -384,7 +386,10 @@ func TestWorkManagerWorkRankingScheduling(t *testing.T) {
 	const numQueries = 4
 	const numWorkers = 8
 
-	wm, workers := startWorkManager(t, numWorkers)
+	workMgr, workers := startWorkManager(t, numWorkers)
+
+	wm, ok := workMgr.(*workManager)
```
guggero, line 391: nit: could use `require.IsType(t, &workManager{}, workMgr)` for a nicer failure message if this ever is `false`. Also, there's a (pre-existing) typo in the godoc comment of this test function.

guggero (approved):
Very nice, LGTM :tada: 

Can we turn https://github.com/ellemouton/lnd/pull/69 into an upstream PR so we can track CI there as well?

query/workmanager.go
Roasbeef: A _named_ break (so using a label), might also be useful here.

query/workmanager.go
```diff
@@ -321,11 +323,50 @@ Loop:
...
+					log.Warnf("Query(%d) from peer %v "+
+						"failed and reached maximum "+
+						"number of retries, not "+
+						"rescheduling: %v",
+						result.job.index,
+						result.peer.Addr(), result.err)
+
+					// Decrement the number of queries
+					// remaining in the batch.
+					batch.rem--
```
Roasbeef, line 357: Shouldn't this instead report back as an error since the number of retries was exceeded?

query/interface.go
```diff
@@ -43,7 +56,7 @@ func defaultQueryOptions() *queryOptions {
 	return &queryOptions{
 		timeout:    defaultQueryTimeout,
 		encoding:   defaultQueryEncoding,
-		cancelChan: nil,
+		numRetries: defaultNumRetries,
```
Roasbeef, line 59: This appears to be breaking behavior for _all_ callers: previously the query functions would retry indefinitely, while now they'll give up after two attempts.

query/workmanager.go
```diff
@@ -96,10 +96,11 @@ type Config struct {
...
 }
 
-// WorkManager is the main access point for outside callers, and satisfies the
+// workManager is the main access point for outside callers, and satisfies the
 // QueryAccess API. It receives queries to pass to peers, and schedules them
-// among available workers, orchestrating where to send them.
-type WorkManager struct {
+// among available workers, orchestrating where to send them. It implements the
+// WorkManager interface.
+type workManager struct {
```
Roasbeef, line 103: Another naming suggestion: `peerWorkManager`?

query/workmanager_test.go
Roasbeef, line 391: Alternatively looks like we can just modify `startWorkManager` to return a pointer to the struct itself?

query.go
```diff
@@ -61,6 +63,13 @@ var (
 	// ErrFilterFetchFailed is returned in case fetching a compact filter
 	// fails.
 	ErrFilterFetchFailed = fmt.Errorf("unable to fetch cfilter")
+
+	// noProgress will be used to indicate to a query.WorkManager that a
+	// response makes no progress towards the completion of the query.
+	noProgress = query.Progress{
```
Roasbeef, line 69: 👍

query.go
```diff
@@ -992,74 +1001,94 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
...
+
+	// Prepare the query request.
+	request := &query.Request{
+		Req:        getData,
+		HandleResp: handleResp,
+	}
+
+	// Prepare the query options.
+	queryOpts := []query.QueryOption{
+		query.Encoding(qo.encoding),
```
Roasbeef, line 913: At a glance the option mapping here looks to be lossy, but need to dig deeper to see if it'll matter in practice (eg: we don't need things like `PeerConnectTimeout` here.

query.go
```diff
@@ -646,18 +769,13 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
...
-		options = append(options, Timeout(QueryBatchTimeout))
-
 	// Reverse batch, fetch as many of the preceding filters as possible.
 	case reverseBatch:
 		stopHeight = int64(height)
 		startHeight = stopHeight - batchSize + 1
 
-		// We need a longer timeout, since we are going to receive more
-		// than a single response.
-		options = append(options, Timeout(QueryBatchTimeout))
```
Roasbeef, line 651: Why do we no longer need the batch timeout?

query.go
```diff
@@ -619,23 +745,20 @@ func (s *ChainService) prepareCFiltersQuery(blockHash chainhash.Hash,
 	}
 	bestHeight := int64(bestBlock.Height)
 
-	qo := defaultQueryOptions()
```
Roasbeef, line 613: I think rather than remove the old set of options, we should instead seek to unify them. If we need to add extra options that are relevant to the main package, but not the `query` package, then we can use a type alias here.

Roasbeef (changes requested):

Roasbeef:
Ah ok, I see the distinction now: 

https://github.com/lightninglabs/neutrino/blob/d209c5e8abdb6d10b03b088a1f0bf40540fcf097/query.go#L140-L151 

vs 

https://github.com/lightninglabs/neutrino/blob/d209c5e8abdb6d10b03b088a1f0bf40540fcf097/query/interface.go#L35-L48

query.go
ellemouton: Yeah it's defs lossy. See my overall review comment on the trickiness of the mapping between these two sets of
functional options

query.go
ellemouton: > If we need to add extra options that are relevant to the main package, but not the query package, then we can use a type alias here.

can you elaborate on "we can use a type alias here"?

query.go
ellemouton: imo the dynamic timeout is way nicer cause then we dont wait for 30 seconds on a non-responsive peer when all we want is 2 filters.

ellemouton (commented):
>  but is it not possible to create a 1:1 mapping between them?

I've tried various things to do this but it always ends up being quite contrived. So for example: 
- the query dispatcher makes use of a dynamic query timeout meaning that it tries a low timeout and increases it if needed. This also means that this timeout is not exposed to the caller. But the neutrino query options have a `Timeout` option used to set the static timeout that was used in the now deleted query method. So does this then mean we add a `WithStaticQueryTimout` option to the query dispatcher and lose the nice dynamic timeout feature there? 

Also, just in general: mapping one set of functional options to another is tricky cause it is hard to tell what the user specifically set/requested and what was set by default. So it is hard to know what is ok to override in the original set of options. 

> Or is this about documenting a potential breaking API change?

So im trying to determine if we should do a whole thing to make the options mappable or if we should rather just add extra documentation around the functional options saying "This one wont apply for GetBlock" or w/e. And then the other option is to break the API completely and have a new set of functional options.

ellemouton:
@Roasbeef  - re-requesting just to get your opinion on how to move forward with the 2 separate/conflicting query option sets

Roasbeef:
> So im trying to determine if we should do a whole thing to make the options mappable or if we should rather just add extra documentation around the functional options saying "This one wont apply for GetBlock" or w/e. And then the other option is to break the API completely and have a new set of functional options.

So I think we may be able to unify things by inserting a type param here: there'd be a set of "static" values (say the timeout, etc), then a value `T` that's inserted to allow the specific package/call site to insert the extra information needed for the error. So something along the lines of: https://golang.design/research/generic-option/

I don't think this should block the rest of this PR series though, as upfront we've been able to identify some option abstraction leaks while ensuring the existing behavior is properly capture as is.

Roasbeef:
> So does this then mean we add a WithStaticQueryTimout option to the query dispatcher and lose the nice dynamic timeout feature there?


Agreed that in most cases we do want a dynamic timeout vs a static one. In the past, we had some hard coded timeouts tuned by devs in SF, then those in South America needed to retune things as certain timeouts were no longer valid across the higher latency environment (was mobile also).

Roasbeef (approved):
LGTM 🦠

PR #274, opened 2023-05-05
multi: introduce a batch filter writer

ellemouton:
This PR adds a generic `BatchWriter` which can be used to batch write items to a db.
It persists the batch once the size reaches a certain threshold _or_ if a ticker goes off 
if at least one items is in the batch. 

This `BatchWriter` is then used by the `GetCFilter` method to persist filters that it
gets during a query. This will improve the speed of the query since the `BatchWriter` is
non-blocking. 

## Performance Improvement

On current master branch, syncing 3200 filters on my local regtest network takes 4m25s
With this PR, syncing the same number of filters takes 4.2s

## LND CI

LND CI with this PR is being tested here: https://github.com/lightningnetwork/lnd/pull/7788

### DIFF
--- a/chanutils/batch_writer.go
+++ b/chanutils/batch_writer.go
@@ -1,0 +1,149 @@
+package chanutils
+
+import (
+	"sync"
+	"time"
+)
+
+// BatchWriterConfig holds the configuration options for BatchWriter.
+type BatchWriterConfig[T any] struct {
+	// QueueBufferSize sets the buffer size of the output channel of the
+	// concurrent queue used by the BatchWriter.
+	QueueBufferSize int
+
+	// MaxBatch is the maximum number of filters to be persisted to the DB
+	// in one go.
+	MaxBatch int
+
+	// DBWritesTickerDuration is the time after receiving a filter that the
+	// writer will wait for more filters before writing the current batch
+	// to the DB.
+	DBWritesTickerDuration time.Duration
+
+	// PutItems will be used by the BatchWriter to persist filters in
+	// batches.
+	PutItems func(...T) error
+}
+
+// BatchWriter manages writing Filters to the DB and tries to batch the writes
+// as much as possible.
+type BatchWriter[T any] struct {
+	started sync.Once
+	stopped sync.Once
+
+	cfg *BatchWriterConfig[T]
+
+	queue *ConcurrentQueue[T]
+
+	quit chan struct{}
+	wg   sync.WaitGroup
+}
+
+// NewBatchWriter constructs a new BatchWriter using the given
+// BatchWriterConfig.
+func NewBatchWriter[T any](cfg *BatchWriterConfig[T]) *BatchWriter[T] {
+	return &BatchWriter[T]{
+		cfg:   cfg,
+		queue: NewConcurrentQueue[T](cfg.QueueBufferSize),
+		quit:  make(chan struct{}),
+	}
+}
+
+// Start starts the BatchWriter.
+func (b *BatchWriter[T]) Start() {
+	b.started.Do(func() {
+		b.queue.Start()
+
+		b.wg.Add(1)
+		go b.manageNewItems()
+	})
+}
+
+// Stop stops the BatchWriter.
+func (b *BatchWriter[T]) Stop() {
+	b.stopped.Do(func() {
+		close(b.quit)
+		b.wg.Wait()
+
+		b.queue.Stop()
+	})
+}
+
+// AddItem adds a given item to the BatchWriter queue.
+func (b *BatchWriter[T]) AddItem(item T) {
+	b.queue.ChanIn() <- item
+}
+
+// manageNewItems manages collecting filters and persisting them to the DB.
+// There are two conditions for writing a batch of filters to the DB: the first
+// is if a certain threshold (MaxBatch) of filters has been collected and the
+// other is if at least one filter has been collected and a timeout has been
+// reached.
+//
+// NOTE: this must be run in a goroutine.
+func (b *BatchWriter[T]) manageNewItems() {
+	defer b.wg.Done()
+
+	batch := make([]T, 0, b.cfg.MaxBatch)
+
+	// writeBatch writes the current contents of the batch slice to the
+	// filters DB.
+	writeBatch := func() {
+		if len(batch) == 0 {
+			return
+		}
+
+		err := b.cfg.PutItems(batch...)
+		if err != nil {
+			log.Errorf("Could not write filters to filterDB: %v",
+				err)
+		}
+
+		// Empty the batch slice.
+		batch = make([]T, 0, b.cfg.MaxBatch)
+	}
+
+	ticker := time.NewTicker(b.cfg.DBWritesTickerDuration)
+	defer ticker.Stop()
+
+	// Stop the ticker since we don't want it to tick unless there is at
+	// least one item in the queue.
+	ticker.Stop()
+
+	for {
+		select {
+		case filter, ok := <-b.queue.ChanOut():
+			if !ok {
+				return
+			}
+
+			batch = append(batch, filter)
+
+			switch len(batch) {
+			// If the batch slice is full, we stop the ticker and
+			// write the batch contents to disk.
+			case b.cfg.MaxBatch:
+				ticker.Stop()
+				writeBatch()
+
+			// If an item is added to the batch, we reset the timer.
+			// This ensures that if the batch threshold is not met
+			// then items are still persisted in a timely manner.
+			default:
+				ticker.Reset(b.cfg.DBWritesTickerDuration)
+			}
+
+		case <-ticker.C:
+			// If the ticker ticks, then we stop it and write the
+			// current batch contents to the db. If any more items
+			// are added, the ticker will be reset.
+			ticker.Stop()
+			writeBatch()
+
+		case <-b.quit:
+			writeBatch()
+
+			return
+		}
+	}
+}
--- a/chanutils/batch_writer_test.go
+++ b/chanutils/batch_writer_test.go
@@ -1,0 +1,210 @@
+package chanutils
+
+import (
+	"fmt"
+	"math/rand"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+const waitTime = time.Second * 5
+
+// TestBatchWriter tests that the BatchWriter behaves as expected.
+func TestBatchWriter(t *testing.T) {
+	t.Parallel()
+	rand.Seed(time.Now().UnixNano())
+
+	// waitForItems is a helper function that will wait for a given set of
+	// items to appear in the db.
+	waitForItems := func(db *mockItemsDB, items ...*item) {
+		err := waitFor(func() bool {
+			return db.hasItems(items...)
+		}, waitTime)
+		require.NoError(t, err)
+	}
+
+	t.Run("filters persisted after ticker", func(t *testing.T) {
+		t.Parallel()
+
+		// Create a mock filters DB.
+		db := newMockItemsDB()
+
+		// Construct a new BatchWriter backed by the mock db.
+		b := NewBatchWriter[*item](&BatchWriterConfig[*item]{
+			QueueBufferSize:        10,
+			MaxBatch:               20,
+			DBWritesTickerDuration: time.Millisecond * 500,
+			PutItems:               db.PutItems,
+		})
+		b.Start()
+		t.Cleanup(b.Stop)
+
+		fs := genFilterSet(5)
+		for _, f := range fs {
+			b.AddItem(f)
+		}
+		waitForItems(db, fs...)
+	})
+
+	t.Run("write once threshold is reached", func(t *testing.T) {
+		t.Parallel()
+
+		// Create a mock filters DB.
+		db := newMockItemsDB()
+
+		// Construct a new BatchWriter backed by the mock db.
+		// Make the DB writes ticker duration extra long so that we
+		// can explicitly test that the batch gets persisted if the
+		// MaxBatch threshold is reached.
+		b := NewBatchWriter[*item](&BatchWriterConfig[*item]{
+			QueueBufferSize:        10,
+			MaxBatch:               20,
+			DBWritesTickerDuration: time.Hour,
+			PutItems:               db.PutItems,
+		})
+		b.Start()
+		t.Cleanup(b.Stop)
+
+		// Generate 30 filters and add each one to the batch writer.
+		fs := genFilterSet(30)
+		for _, f := range fs {
+			b.AddItem(f)
+		}
+
+		// Since the MaxBatch threshold has been reached, we expect the
+		// first 20 filters to be persisted.
+		waitForItems(db, fs[:20]...)
+
+		// Since the last 10 filters don't reach the threshold and since
+		// the ticker has definitely not ticked yet, we don't expect the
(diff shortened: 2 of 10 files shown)


filterdb/db.go
```diff
@@ -155,30 +167,22 @@ func putFilter(bucket walletdb.ReadWriteBucket, hash *chainhash.Hash,
...
-func (f *FilterStore) PutFilter(hash *chainhash.Hash,
-	filter *gcs.Filter, fType FilterType) error {
-
+func (f *FilterStore) PutFilter(filterData *FilterData) error {
 	return walletdb.Update(f.db, func(tx walletdb.ReadWriteTx) error {
 		filters := tx.ReadWriteBucket(filterBucket)
 
 		var targetBucket walletdb.ReadWriteBucket
-		switch fType {
+		switch filterData.Type {
```
Roasbeef: I think we can also start to use `Batch` here as well instead of `Update`.

chanutils/batch_writer.go
```diff
@@ -0,0 +1,157 @@
...
+		close(b.quit)
+		b.wg.Wait()
+
+		b.queue.Stop()
+	})
+}
+
+// AddItem adds a given item to the BatchWriter queue.
+func (b *BatchWriter[T]) AddItem(item T) {
+	b.queue.ChanIn() <- item
```
Roasbeef, line 79: Do we need to synchronously respond to the caller here? Still getting through the diff, but wondering if there can be a consistency issue here if we think we've written some filters, but then try to process the next batch in turn.

chanutils/batch_writer.go
```diff
@@ -0,0 +1,157 @@
...
+				b.cfg.Logger.Warnf("Couldn't write filters to "+
+					"filterDB: %v", err)
+			}
+
+			return
+		}
+	}
+
+	ticker := time.NewTicker(b.cfg.DBWritesTickerDuration)
+	ticker.Stop()
```
Roasbeef, line 121: Missing a `defer`.

Roasbeef: Also doesn't look like the ticker is ever restarted.

query.go
```diff
@@ -551,16 +551,11 @@ func (q *cfiltersQuery) handleResponse(req, resp wire.Message,
 	}
 
 	if q.cs.persistToDisk {
-		filterData := &filterdb.FilterData{
+		q.cs.filterBatchWriter.AddItem(&filterdb.FilterData{
```
Roasbeef, line 554: 👍

filterdb/db.go
ellemouton: cool - added this in the next commit 👍

filterdb/db.go
```diff
@@ -163,26 +163,38 @@ func putFilter(bucket walletdb.ReadWriteBucket, hash *chainhash.Hash,
...
 
-// PutFilter stores a filter with the given hash and type to persistent
-// storage.
+// PutFilters stores a set of filters to persistent storage.
 //
 // NOTE: This method is a part of the FilterDatabase interface.
-func (f *FilterStore) PutFilter(filterData *FilterData) error {
-	return walletdb.Update(f.db, func(tx walletdb.ReadWriteTx) error {
+func (f *FilterStore) PutFilters(filterList ...*FilterData) error {
+	return walletdb.Batch(f.db, func(tx walletdb.ReadWriteTx) error {
```
yyforyongyu, line 170: To make the best usage of `Batch`, we'd need to distinguish two types of errors,
- database error: the error which causes the whole `Batch` to fail, which means all other batched transactions will fail, such as a bucket is not found.
- transaction error: the error returned from a specific transaction, which only causes the current tx to fail.

[Here's a test](https://gist.github.com/yyforyongyu/fc82b46314d3ee7128dd8312182de785) I wrote before to check the `Batch` behavior. And we use [this pattern in lnd](https://github.com/lightningnetwork/lnd/blob/master/channeldb/payment_control.go#L130).

chanutils/queue.go
```diff
@@ -0,0 +1,132 @@
...
+	// DefaultQueueSize is the default size to use for concurrent queues.
+	DefaultQueueSize = 10
+)
+
+// ConcurrentQueue is a typed concurrent-safe FIFO queue with unbounded
+// capacity. Clients interact with the queue by pushing items into the in
+// channel and popping items from the out channel. There is a goroutine that
+// manages moving items from the in channel to the out channel in the correct
+// order that must be started by calling Start().
+type ConcurrentQueue[T any] struct {
```
yyforyongyu, line 18: I'm assuming the package `module github.com/lightningnetwork/lnd/queue` is no longer maintained right?

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// DBWritesTickerDuration is the time after receiving a filter that the
+	// writer will wait for more filters before writing the current batch
+	// to the DB.
+	DBWritesTickerDuration time.Duration
+
+	// Logger is the logger that the BatchWriter should use for any logs.
+	Logger btclog.Logger
+
+	// PutFilters will be used by the BatchWriter to persist filters in
```
yyforyongyu, line 28:
```suggestion
	// PutItems will be used by the BatchWriter to persist filters in
```

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+			select {
+			case filter := <-batch:
+				filterList = append(filterList, filter)
+				continue
+			case <-b.quit:
+				return
+			default:
+			}
+
+			err := b.cfg.PutItems(filterList...)
```
yyforyongyu, line 112: Looks like we need to `PutItems` outside the `for` loop to drain the `batch` channel?

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+			return
+		}
+	}
+
+	ticker := time.NewTicker(b.cfg.DBWritesTickerDuration)
+	defer ticker.Stop()
+
+	// Stop the ticker since we don't want it to tick unless there is at
+	// least one item in the queue.
+	ticker.Stop()
```
yyforyongyu, line 127: uncleaned code?

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+
+		case <-ticker.C:
+			// If the ticker ticks, then we stop it and write the
+			// current batch contents to the db. If any more items
+			// are added, the ticker will be reset.
+			ticker.Stop()
+			writeBatch()
+
+		case <-b.quit:
```
yyforyongyu, line 164: Think we need to flush `batch` to disk when quitting, in case there are items there.

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+	for {
+		select {
+		case filter, ok := <-b.queue.ChanOut():
+			if !ok {
+				return
+			}
+
+			// Push the new filter onto the batch channel.
+			select {
+			case batch <- filter:
```
yyforyongyu, line 138: Don't quite get the idea of using `chan` instead of a `slice` here, looks like there's no async communication here. Could we use a slice here instead?

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/btcsuite/btclog"
+)
+
+// BatchWriterConfig holds the configuration options for BatchWriter.
+type BatchWriterConfig[T any] struct {
```
yyforyongyu, line 11: Do we plan to use this writer for other data types?

chanutils/batch_writer.go
```diff
@@ -0,0 +1,168 @@
...
+		b.queue.Stop()
+	})
+}
+
+// AddItem adds a given item to the BatchWriter queue.
+func (b *BatchWriter[T]) AddItem(item T) {
+	b.queue.ChanIn() <- item
+}
+
+// manageFilters manages collecting filters and persisting them to the DB.
```
yyforyongyu, line 82:
```suggestion
// manageNewItems manages collecting filters and persisting them to the DB.
```

yyforyongyu (changes requested):
Very cool optimization⚡️ Left some comments on the usage of `Batch` and questions about the method `manageNewItems`.

filterdb/db.go
```diff
@@ -163,26 +163,38 @@ func putFilter(bucket walletdb.ReadWriteBucket, hash *chainhash.Hash,
...
-		case RegularFilter:
-			targetBucket = filters.NestedReadWriteBucket(regBucket)
-		default:
-			return fmt.Errorf("unknown filter type: %v",
-				filterData.Type)
+		for _, filterData := range filterList {
+			var targetBucket walletdb.ReadWriteBucket
+			switch filterData.Type {
+			case RegularFilter:
+				targetBucket = filters.NestedReadWriteBucket(
```
positiveblue, line 177: nit: instead of calling `NestedReadWriteBucket` in every iteration we can call it once before the loop and change targetBucket in the switch statement. Now we have only two optio: regBucket or error but this would work even if we add more in the future

positiveblue (commented):
` from 4m25s to  4.2s` :fire:

Those are amazing improvements Elle :tada: 

 It looks like we can be optimistic when we fetch the filters and assume "eventually they will be persisted". Worst case scenario we will ask for them again, but if we use sensitive default for flushing to disk it should not be a problem.

My main concern is about the concurrent queue. Isn't a simple struct with a slice + a lock enough for this?

chanutils/batch_writer.go
ellemouton: noice catch

ellemouton (commented):
Thanks for the review everyone! 🙏 

You guys were completely right about just using a slice for the batch writer :) I was too trigger happy with the channel 😂 

Updated accordingly!

Chinwendu20:
Hello @ellemouton may I please ask how you timed this?

ellemouton:
@yyforyongyu  & @positiveblue - this is ready! The dependant PR has been merged. PTAL 🙏

chanutils/queue.go
```diff
@@ -0,0 +1,132 @@
...
+	started sync.Once
+	stopped sync.Once
+
+	chanIn   chan T
+	chanOut  chan T
+	overflow *list.List
```
Roasbeef, line 24: Can use this generic list to inherit type safety (or sort of push down the concerns to another layer): https://github.com/lightninglabs/neutrino/blob/5aac983d75cc02fce159b592dab7940df7d494a7/cache/lru/list.go#L62-L63

chanutils/batch_writer.go
Roasbeef: Was thinking of it more from the perspective of filter headers, but I think it's fine here: at this point, we already have all the filter headers and our effective "rescan state" is saved as is on the caller level with the height hint attribute.

chanutils/queue.go
ellemouton: oh sweet - that is perfect. Updated!

ellemouton (commented):
Thanks @Roasbeef! updated. Also updated the LND PR pointing to this 👍

chanutils/batch_writer.go
```diff
@@ -0,0 +1,154 @@
...
+		// Empty the batch slice.
+		batch = make([]T, 0, b.cfg.MaxBatch)
+	}
+
+	ticker := time.NewTicker(b.cfg.DBWritesTickerDuration)
+	defer ticker.Stop()
+
+	// Stop the ticker since we don't want it to tick unless there is at
+	// least one item in the queue.
+	ticker.Stop()
```
yyforyongyu, line 116: nit: there's a near-impossible case, that if the `DBWritesTickerDuration` is extremely small, we may end up having a tick already inside `ticker`, causing an immediate write - no big deal here. But in a hindsight it looks like the `lnd/ticker` package is more suitable for this case.

chanutils/batch_writer_test.go
```diff
@@ -0,0 +1,210 @@
...
+		t.Cleanup(b.Stop)
+
+		fs := genFilterSet(5)
+		for _, f := range fs {
+			b.AddItem(f)
+		}
+		waitForItems(db, fs...)
+	})
+
+	t.Run("write once threshold is reached", func(t *testing.T) {
```
yyforyongyu, line 52: 👍

yyforyongyu (approved):
LGTM🥳 and def impressive optimization😻

chanutils/queue.go
```diff
@@ -0,0 +1,133 @@
...
+			nextElement := cq.overflow.Front()
+			if nextElement == nil {
+				// Overflow queue is empty so incoming items can
+				// be pushed directly to the output channel. If
+				// output channel is full though, push to
+				// overflow.
+				select {
+				case item, ok := <-cq.chanIn:
+					if !ok {
+						break readLoop
```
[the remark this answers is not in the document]
ellemouton: added :+1:

chanutils/batch_writer.go
```diff
@@ -0,0 +1,154 @@
...
+	// writeBatch writes the current contents of the batch slice to the
+	// filters DB.
+	writeBatch := func() {
+		if len(batch) == 0 {
+			return
+		}
+
+		err := b.cfg.PutItems(batch...)
+		if err != nil {
+			b.cfg.Logger.Warnf("Couldn't write filters to "+
```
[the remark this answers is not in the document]
ellemouton: changed 👍

chanutils/batch_writer.go
ellemouton: cool yeah im happy to update to the lnd ticker. 

The only reason not to would be for performance reasons: this ticker stops and starts the same one the whole time while the LND one keeps creating new ones & discarding old ones. 
So maybe makes sense to keep as is? Agreed re the possible immediate write but also agree that it is not a big deal 👍 

lemme know :)

ellemouton: also doesnt really make sense to make the `DBWritesTickerDuration` too small cause then the batching benefits are lost

ellemouton (commented):
Thanks @yyforyongyu !

chanutils/batch_writer.go
yyforyongyu: yeah i think it's fine to leave it as-is, in the future we can add some config sanity check to make sure it won't happen - or let it happen, worse case we get an immediate call to the batch write.

chanutils/queue.go
```diff
@@ -0,0 +1,143 @@
...
+			nextElement := cq.overflow.Front()
+			if nextElement == nil {
+				// Overflow queue is empty so incoming items can
+				// be pushed directly to the output channel. If
+				// output channel is full though, push to
+				// overflow.
+				select {
+				case item, ok := <-cq.chanIn:
+					if !ok {
+						log.Warnf("ConcurrentQueue " +
```
yyforyongyu, line 78: 🙏

yyforyongyu (approved):
LGTM⛵️

Chinwendu20:
It seems like this change has not been reflected in btcwallet. I just rebased my branch and I could not run lnd with this new change because of this line of code in btcwallet @ellemouton 
https://github.com/btcsuite/btcwallet/blob/23eda3cadf717dbf79d191873fc524178492aef9/chain/pruned_block_dispatcher.go#L194

ellemouton:
@Chinwendu20 see https://github.com/btcsuite/btcwallet/pull/884 and https://github.com/lightningnetwork/lnd/pull/7788

PR #275, opened 2023-05-05
rescan: use batch filter fetching

ellemouton:
With this PR, we ensure that `rescan` can make use of batch filter fetching by 
waiting until the header chain is either current or until it is ahead of the specified 
end height before starting the scan. This greatly improves the speed of a rescan 
because before this commit, the filters would be fetched one-by-one. 

The first few commits of this PR are just refactor commits that clean-up the `rescan`
function and then the actual behaviour change is added in the final commit. 

Fixes https://github.com/lightninglabs/neutrino/issues/68
Replaces https://github.com/lightninglabs/neutrino/pull/236

## Performance Improvement

On current master branch, syncing 3200 filters on my local regtest network takes 4m25s
With this PR, syncing the same number of filters takes 653ms

## LND CI

https://github.com/lightningnetwork/lnd/pull/7788

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -925,10 +925,7 @@ rescanLoop:
 				)
 			}
 
-			err = notifyBlock(
-				chain, ro, rs.curHeader, rs.curStamp,
-				rs.scanning,
-			)
+			err = rs.notifyBlock()
 			if err != nil {
 				return err
 			}
@@ -937,19 +934,19 @@ rescanLoop:
 }
 
 // notifyBlock calls appropriate listeners based on the block filter.
-func notifyBlock(chain ChainSource, ro *rescanOptions,
-	curHeader wire.BlockHeader, curStamp headerfs.BlockStamp,
-	scanning bool) error {
+func (rs *rescanState) notifyBlock() error {
+	chain := rs.chain
+	ro := rs.opts
 
 	// Find relevant transactions based on watch list. If scanning is
 	// false, we can safely assume this block has no relevant transactions.
 	var relevantTxs []*btcutil.Tx
-	if len(ro.watchList) != 0 && scanning {
+	if len(ro.watchList) != 0 && rs.scanning {
 		// If we have a non-empty watch list, then we need to see if it
 		// matches the rescan's filters, so we get the basic filter
 		// from the DB or network.
 		matched, filter, err := blockFilterMatches(
-			chain, ro, &curStamp.Hash,
+			chain, ro, &rs.curStamp.Hash,
 		)
 		if err != nil {
 			return err
@@ -957,7 +954,7 @@ func notifyBlock(chain ChainSource, ro *rescanOptions,
 
 		if matched {
 			relevantTxs, err = extractBlockMatches(
-				chain, ro, &curStamp, filter,
+				chain, ro, &rs.curStamp, filter,
 			)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


rescan.go
```diff
@@ -289,9 +289,29 @@ func updateChan(update <-chan *updateOptions) RescanOption {
 	}
 }
 
-// rescan is a single-threaded function that uses headers from the database and
-// functional options as arguments.
-func rescan(chain ChainSource, options ...RescanOption) error {
+// rescanState hold the state used throughout a rescan.
+type rescanState struct {
```
Roasbeef, line 293: 👍

rescan.go
```diff
@@ -937,27 +934,27 @@ rescanLoop:
 }
 
 // notifyBlock calls appropriate listeners based on the block filter.
-func notifyBlock(chain ChainSource, ro *rescanOptions,
-	curHeader wire.BlockHeader, curStamp headerfs.BlockStamp,
-	scanning bool) error {
+func (rs *rescanState) notifyBlock() error {
```
Roasbeef, line 852: 👍

rescan.go
```diff
@@ -1090,13 +1090,13 @@ func notifyBlockWithFilter(chain ChainSource, ro *rescanOptions,
...
 	if ro.ntfn.OnFilteredBlockConnected != nil {
-		ro.ntfn.OnFilteredBlockConnected(curStamp.Height, curHeader,
+		ro.ntfn.OnFilteredBlockConnected(stamp.Height, header,
 			relevantTxs)
 	}
 
 	if ro.ntfn.OnBlockConnected != nil { // nolint:staticcheck
-		ro.ntfn.OnBlockConnected(&curStamp.Hash, // nolint:staticcheck
-			curStamp.Height, curHeader.Timestamp)
+		ro.ntfn.OnBlockConnected(&stamp.Hash, // nolint:staticcheck
```
Roasbeef, line 1136: Let's fix the line folding here while we're at it?

rescan.go
```diff
@@ -442,6 +442,39 @@ func (rs *rescanState) rescan() error {
 		return err
 	}
 
+	// To ensure that we batch as many filter queries as possible, we also
+	// wait for the header chain to either be current or for it to at least
+	// have caught up with the specified end block.
+	r, ok := chain.(*RescanChainSource)
```
Roasbeef, line 448: Perhaps we can extend the interface to permit these operations w/o needing to type cast?

Roasbeef: Looks like all we need is the `IsCurrent()` method.

Roasbeef (changes requested):

rescan.go
ellemouton: good idea :) updated!

rescan.go
ellemouton: 😎

Roasbeef (approved):
LGTM 🦜

saubyk:
cc @ProofOfKeags for review request

positiveblue (approved):
Epic performance improvements @ellemouton :tada:

GTG

ProofOfKeags:
post merge ACK [b38fa7f](https://github.com/lightninglabs/neutrino/pull/275/commits/b38fa7f50176dca35b7fabbb475eceae7d6283f6)

It took me a sec to realize most of the diff was just nesting some values under the state structure. Seems like the only major change was blocking on waiting for the header chain to get far enough ahead to maximize the benefits of batching.

PR #277, opened 2023-05-15
verification: mute taproot related err to dbg

guggero:
We can't re-create the pkScript from a keyspend p2tr witness, as we're missing the public key. So this warning will trigger for each block that contains a p2tr key spend, which isn't really useful and confuses users. So we're downgrading the error to a debug message.


positiveblue (approved):

Roasbeef (approved):
LGTM ♻️

PR #278, opened 2023-06-03
pushtx: map different backend err to internal err.

ziggie1984:
Relates to lightningnetwork/lnd#7156.

~~Looking for a Concept ACK here.~~

This makes sure our rebroadcaster continues broadcasting transactions when they are for example rejected because the `min mempool fee` is not met. ~~It becomes a bit messy to now include some other backend conversions into the `pushtx` package which is mainly a neutrino package? So the most cleanest way would be to maybe reimplement the rebroadcaster for the different backends?~~
~~This solution gets the job done, so open for suggestions.~~

### DIFF
--- a/pushtx/broadcaster.go
+++ b/pushtx/broadcaster.go
@@ -52,6 +52,12 @@ type Config struct {
 	// RebroadcastInterval is the interval that we'll continually try to
 	// re-broadcast transactions in-between new block arrival.
 	RebroadcastInterval time.Duration
+
+	// MapCustomBroadcastError allows the Rebroadcaster to map broadcast
+	// errors from other backends to the neutrino internal BroadcastError.
+	// This allows the Rebroadcaster to behave consistently over different
+	// backends.
+	MapCustomBroadcastError func(error) error
 }
 
 // Broadcaster is a subsystem responsible for reliably broadcasting transactions
@@ -171,10 +177,19 @@ func (b *Broadcaster) broadcastHandler(sub *blockntfns.Subscription) {
 		// A new broadcast request was submitted by an external caller.
 		case req := <-b.broadcastReqs:
 			err := b.cfg.Broadcast(req.tx)
-			if err != nil && !IsBroadcastError(err, Mempool) {
-				log.Errorf("Broadcast attempt failed: %v", err)
-				req.errChan <- err
-				continue
+			if err != nil {
+				// We apply the custom err mapping function if
+				// it was supplied which allows to map other
+				// backend errors to the neutrino BroadcastError.
+				if b.cfg.MapCustomBroadcastError != nil {
+					err = b.cfg.MapCustomBroadcastError(err)
+				}
+				if !IsBroadcastError(err, Mempool) {
+					log.Errorf("Broadcast attempt "+
+						"failed: %v", err)
+					req.errChan <- err
+					continue
+				}
 			}
 
 			transactions[req.tx.TxHash()] = req.tx
@@ -231,6 +246,12 @@ func (b *Broadcaster) rebroadcast(txs map[chainhash.Hash]*wire.MsgTx,
 		}
 
 		err := b.cfg.Broadcast(tx)
+		// We apply the custom err mapping function if it was supplied
+		// which allows to map other backend errors to the neutrino
+		// BroadcastError.
+		if err != nil && b.cfg.MapCustomBroadcastError != nil {
+			err = b.cfg.MapCustomBroadcastError(err)
+		}
 		switch {
 		// If the transaction has already confirmed on-chain, we can
 		// stop broadcasting it further.
--- a/pushtx/broadcaster_test.go
+++ b/pushtx/broadcaster_test.go
@@ -1,10 +1,13 @@
 package pushtx
 
 import (
+	"fmt"
 	"math/rand"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pushtx/error.go
```diff
@@ -150,3 +155,72 @@ func ParseBroadcastError(msg *wire.MsgReject, peerAddr string) *BroadcastError {
 	reason := fmt.Sprintf("rejected by %v: %v", peerAddr, msg.Reason)
 	return &BroadcastError{Code: code, Reason: reason}
 }
+
+// ParseBackendError maps an error received by another backend for example
+// bitcoind and maps it to the neutrino backend error.
```
hieblmi, line 160: nit: ...for example bitcoind ~and maps it~ to the neutrino backend error.

pushtx/error.go
```diff
@@ -150,3 +159,72 @@ func ParseBroadcastError(msg *wire.MsgReject, peerAddr string) *BroadcastError {
...
+		fallthrough
+
+	// This error is returned when broadcasting a transaction that has
+	// already confirmed to a bitcoind node over the P2P network.
+	case match(err, "txn-already-known"):
+		fallthrough
+
+	// This error is returned when broadcasting/sending a transaction to a
+	// btcd node that already has it in their mempool.
+	case match(err, "already have transaction"):
```
bitromortac, line 215: Should we copy over most of https://github.com/btcsuite/btcwallet/blob/4383930e9c4c07dba5cf7444b0c08c47bd2be30a/wallet/wallet.go#L3655-L3795 @Roasbeef?

I think that this case should belong to `code = Mempool` as the comment describes

pushtx/broadcaster.go
```diff
@@ -171,7 +171,17 @@ func (b *Broadcaster) broadcastHandler(sub *blockntfns.Subscription) {
...
-			if err != nil && !IsBroadcastError(err, Mempool) {
+			// The Rebroadcaster is not only used by neutrino but
+			// also with other backends now, so we need to make sure
+			// to map the error to the internal BroadcastError.
+			if err != nil {
+				if _, ok := err.(*BroadcastError); !ok {
+					err = MapBackendError(err)
+				}
+			}
+			if err != nil && !IsBroadcastError(err, Mempool) &&
```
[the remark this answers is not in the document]
ziggie1984: hmm if I do so I kinda mess with the 80 line-limit we have, so I think I keep as is ?

pushtx/error.go
ziggie1984: I agree, I copied the complete error list over feels like the right way from me too.

pushtx/broadcaster_test.go
```diff
@@ -187,16 +189,60 @@ func TestRebroadcast(t *testing.T) {
...
+	// Manually mark the third transaction as confirmed.
 	broadcaster.MarkAsConfirmed(txs[2].TxHash())
 
+	// Now, we'll modify the Broadcast method to mark the fourth transaction
+	// as confirmed but with a bitcoind backend notification to test
+	// that the mapping between different backend errors and the
+	// neutrino BroadcastError works as expected.
+	// We also mark the last transaction with with the bitcoind backend
+	// error for not having enough fees to be included in the mempool. We
+	// expected that it gets rebroadcasted as well.
```
bitromortac, line 204: typos: `with with`, `expected`

bitromortac (approved):
LGTM :tada:, only minor nits

Roasbeef:
>  It becomes a bit messy to now include some other backend conversions into the pushtx package which is mainly a neutrino package? So the most cleanest way would be to maybe reimplement the rebroadcaster for the different backends?

Agree would be ideal to actually move this entire package into `btcwallet` itself. There's another path where we make a small cut out here for custom error detection, then hook that up to the existing function (basically the switch case we'd copied over here) in `btcwallet` when we create the rebroadcaster in `lnd`. At least that way we are able to re-use code, and keep all the error parsing/interpretation in a single place.

pushtx/broadcaster.go
Roasbeef, line 182: Re my suggestion in the comments, we'd then pass in that error detection predicate into `IsBroadcastErrror`.

ziggie1984: Thanks for your suggestion, I implemented it with cut out. My new design removes the duplicated code, but not sure if supplying a `MapCustomBroadcastError` feels intuitive? Maybe you had something even more elegant in mind?

Roasbeef (approved):
LGTM 👑

just needs a rebase

bitromortac (approved):
LGTM :+1:! The only thing we may consider is to update the description of the `Mempool`
error code to something like:

```
        // Mempool is the code used when a transaction already exists in a
        // peer's mempool or when the transaction was rejected not meeting the
        // mempool's minimum fee.
        Mempool
```

ziggie1984:
My intention was to leave the neutrino code and therefore the Error Definition mostly untouched because it's still taylored for the neutrino backend and there this kind of description could be maybe confusing? But  happy to include your comment after your final thought.

bitromortac:
> My intention was to leave the neutrino code and therefore the Error Definition mostly untouched because it's still taylored for the neutrino backend and there this kind of description could be maybe confusing? But happy to include your comment after your final thought.

Yes, good argument, then let's keep it like it is. It was confusing me (in the context of the lnd PR) because I thought it would not include the case of min mempool fee (but it does), so just wanted to mention it here.

PR #281, opened 2023-06-22
Disable retargeting on regtest to preserve compatibility with bitcoind

guggero:
Replaces https://github.com/lightninglabs/neutrino/pull/256/, now that https://github.com/btcsuite/btcd/pull/1985 was merged.

Fixes https://github.com/lightninglabs/neutrino/issues/255.

cc @chappjc, @amanzag

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -2736,6 +2736,11 @@ func (b *blockManager) checkHeaderSanity(blockHeader *wire.BlockHeader,
 func (b *blockManager) calcNextRequiredDifficulty(newBlockTime time.Time,
 	reorgAttempt bool) (uint32, error) {
 
+	// Some networks don't support difficulty retargeting.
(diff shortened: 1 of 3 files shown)


blockmanager.go
```diff
@@ -2736,6 +2736,11 @@ func (b *blockManager) checkHeaderSanity(blockHeader *wire.BlockHeader,
 func (b *blockManager) calcNextRequiredDifficulty(newBlockTime time.Time,
 	reorgAttempt bool) (uint32, error) {
 
+	// Some networks don't support difficulty retargeting.
+	if b.cfg.ChainParams.PoWNoRetargeting {
```
Roasbeef, line 2740: I think we want to go with this PR instead: https://github.com/lightninglabs/neutrino/pull/263

Then we just share all this code, and don't need to make changes in two areas.

Roasbeef (changes requested):

PR #283, opened 2023-07-14
blockmanager.go: use btcd libs to validate headers

guggero:
Replaces https://github.com/lightninglabs/neutrino/pull/263.

This uses btcd's HeaderCtx and ChainCtx interfaces to be able to validate headers, both contextually and context-free. This allows neutrino to call blockchain.CheckBlockHeaderContext and blockchain.CheckBlockHeaderSanity. Also included is a check to assert that when neutrino receives a p2p headers message, that each header connects to the previous one.

Fixes https://github.com/lightninglabs/neutrino/issues/255


Crypt-iQ (approved):

blockmanager_test.go
```diff
@@ -865,3 +869,79 @@ func TestBlockManagerDetectBadPeers(t *testing.T) {
...
+	bm, _, _, err := setupBlockManager(t)
+	require.NoError(t, err)
+
+	fakePeer, err := peer.NewOutboundPeer(&peer.Config{}, "fake:123")
+	require.NoError(t, err)
+
+	assertPeerDisconnected := func(shouldBeDisconnected bool) {
+		// This is quite hacky but works: We expect the peer to be
+		// disconnected, which sets the unexported "disconnected" field
+		// to 1.
```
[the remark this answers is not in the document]
guggero: Unfortunately not, since we're never actually connected during the unit test (didn't find a way to fake being connected either).

Roasbeef:
Small linter error:
```
Error: blockmanager.go:33:2: `maxTimeOffset` is unused (deadcode)
```

guggero:
> 

Oops, fixed.

blockmanager.go
```diff
@@ -2529,8 +2530,27 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
...
 			for j, reorgHeader := range msg.Headers[i:] {
-				err = b.checkHeaderSanity(reorgHeader,
-					maxTimestamp, true)
+				// We have to get the parent's height and
+				// header to be able to contextually validate
+				// this header.
+				prevNodeHeight := backHeight + uint32(j)
+
+				var prevNodeHeader *wire.BlockHeader
+				if i+j == 0 {
```
Roasbeef, line 2535: 🤔

Roasbeef: Aren't these both positive integers? or this is simply detecting the first iteration?

blockmanager.go
```diff
@@ -2743,29 +2763,42 @@ func areHeadersConnected(headers []*wire.BlockHeader) bool {
...
+
+	// Create a lightChainCtx as well.
+	chainCtx := newLightChainCtx(
+		&b.cfg.ChainParams, b.blocksPerRetarget, b.minRetargetTimespan,
+		b.maxRetargetTimespan,
+	)
+
+	var emptyFlags blockchain.BehaviorFlags
+	err := blockchain.CheckBlockHeaderContext(
+		blockHeader, parentHeaderCtx, emptyFlags, chainCtx, true,
```
Roasbeef, line 2788: Don't we also want to continue to verify the checkpoints here as well? Then we can remove the existing code to verify the checkpoints.

blockmanager.go
guggero: Yes, this just means "the very first iteration".

blockmanager.go
Crypt-iQ: I think I found it more complicated to also verify the checkpoints in btcd & would've rather had it in a different PR

guggero: Yeah, we would need to implement `lightChainCtx.VerifyCheckpoint` and `lightChainCtx.FindPreviousCheckpoint`. I can do that, but it'll delay this PR since I'd also want to add unit tests around that.

Roasbeef (approved):
LGTM 🪬

PR #286, opened 2023-08-22
Added unban peer feature

Chinwendu20:
Changes:

- Added unbanpeer method to chainservice.
- Added unbanipnet method to banstore

Link to issue: https://github.com/lightninglabs/neutrino/issues/253
Link to previous PR for context:
https://github.com/lightninglabs/neutrino/pull/270/

### DIFF
--- a/banman/store.go
+++ b/banman/store.go
@@ -66,6 +66,9 @@ type Store interface {
 
 	// Status returns the ban status for a given IP network.
 	Status(*net.IPNet) (Status, error)
+
+	// UnbanIPNet removes the ban imposed on the specified peer.
+	UnbanIPNet(ipNet *net.IPNet) error
 }
 
 // NewStore returns a Store backed by a database.
@@ -134,6 +137,38 @@ func (s *banStore) BanIPNet(ipNet *net.IPNet, reason Reason, duration time.Durat
 
 		return addBannedIPNet(banIndex, reasonIndex, k, reason, duration)
 	})
+}
+
+// UnbanIPNet removes a ban record for the IP network within the store.
+func (s *banStore) UnbanIPNet(ipNet *net.IPNet) error {
+	err := walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		banStore := tx.ReadWriteBucket(banStoreBucket)
+		if banStore == nil {
+			return ErrCorruptedStore
+		}
+
+		banIndex := banStore.NestedReadWriteBucket(banBucket)
+		if banIndex == nil {
+			return ErrCorruptedStore
+		}
+
+		reasonIndex := banStore.NestedReadWriteBucket(reasonBucket)
+		if reasonIndex == nil {
+			return ErrCorruptedStore
+		}
+
+		var ipNetBuf bytes.Buffer
+		if err := encodeIPNet(&ipNetBuf, ipNet); err != nil {
+			return fmt.Errorf("unable to encode %v: %v", ipNet,
+				err)
+		}
+
+		k := ipNetBuf.Bytes()
+
+		return removeBannedIPNet(banIndex, reasonIndex, k)
+	})
+
+	return err
 }
 
 // addBannedIPNet adds an entry to the ban store for the given IP network.
--- a/banman/store_test.go
+++ b/banman/store_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/btcsuite/btcwallet/walletdb"
 	_ "github.com/btcsuite/btcwallet/walletdb/bdb"
 	"github.com/lightninglabs/neutrino/banman"
+	"github.com/stretchr/testify/require"
 )
 
 // createTestBanStore creates a test Store backed by a boltdb instance.
@@ -123,4 +124,10 @@ func TestBanStore(t *testing.T) {
 	// We'll query for second IP network again as it should now be unknown
 	// to the BanStore. We should expect not to find anything regarding it.
 	checkBanStore(ipNet2, false, 0, 0)
+
+	// Test UnbanIPNet.
+	require.NoError(t, banStore.UnbanIPNet(ipNet1))
+
+	// We would now check that ipNet1 is indeed unbanned.
+	checkBanStore(ipNet1, false, 0, 0)
 }
--- a/neutrino.go
+++ b/neutrino.go
@@ -1081,6 +1081,24 @@ func (s *ChainService) BanPeer(addr string, reason banman.Reason) error {
 	return s.banStore.BanIPNet(ipNet, reason, BanDuration)
 }
 
+// UnbanPeer connects and unbans a previously banned peer.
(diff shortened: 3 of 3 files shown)


Chinwendu20:
Hello @Roasbeef and @positiveblue . I have added the test

banman/store_test.go
```diff
@@ -123,4 +123,10 @@ func TestBanStore(t *testing.T) {
 	// We'll query for second IP network again as it should now be unknown
 	// to the BanStore. We should expect not to find anything regarding it.
 	checkBanStore(ipNet2, false, 0, 0)
+
+	//Unban ipNet1
```
Roasbeef, line 127: Style nit: comments should be complete sentences.

neutrino.go
```diff
@@ -1081,6 +1081,33 @@ func (s *ChainService) BanPeer(addr string, reason banman.Reason) error {
...
+
+	defer func() error {
+		if sp := s.PeerByAddr(addr); sp != nil {
+			tcpAddr, err := s.addrStringToNetAddr(addr)
+			if err != nil {
+				return err
+			}
+			s.connManager.Connect(&connmgr.ConnReq{
+				Addr:      tcpAddr,
+				Permanent: true,
```
Roasbeef, line 1096: Perhaps this should be optional? So a bool that controls if we'll try to permanently connecvt to them after banning.

Roasbeef (changes requested):

banman/store.go
```diff
@@ -136,6 +139,34 @@ func (s *banStore) BanIPNet(ipNet *net.IPNet, reason Reason, duration time.Durat
...
 }
 
+// UnbanIPNet removes a ban record for the IP network within the store.
+func (s *banStore) UnbanIPNet(ipNet *net.IPNet) error {
+	err := walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		banStore := tx.ReadWriteBucket(banStoreBucket)
+		if banStore == nil {
+			return ErrCorruptedStore
+		}
+		banIndex := banStore.NestedReadWriteBucket(banBucket)
```
Crypt-iQ, line 149: nit: newlines after braces

banman/store.go
```diff
@@ -136,6 +139,34 @@ func (s *banStore) BanIPNet(ipNet *net.IPNet, reason Reason, duration time.Durat
...
+			return ErrCorruptedStore
+		}
+		reasonIndex := banStore.NestedReadWriteBucket(reasonBucket)
+		if reasonIndex == nil {
+			return ErrCorruptedStore
+		}
+
+		var ipNetBuf bytes.Buffer
+		if err := encodeIPNet(&ipNetBuf, ipNet); err != nil {
+			return fmt.Errorf("unable to encode %v: %v", ipNet, err)
```
Crypt-iQ, line 160: nit: line length

neutrino.go
Crypt-iQ: I don't think this defer is necessary?

Roasbeef:
@Chinwendu20 still interested in finishing this up?

Chinwendu20:
Thanks @Roasbeef wow, did not notice the comments

neutrino.go
Chinwendu20: Are you saying we should skip the defer function and just connect the peer that has just been unbanned?

banman/store.go
Roasbeef: Should be wrapped to 80 columns.

Roasbeef (approved):
LGTM 👑

banman/store_test.go
```diff
@@ -123,4 +124,10 @@ func TestBanStore(t *testing.T) {
 	// We'll query for second IP network again as it should now be unknown
 	// to the BanStore. We should expect not to find anything regarding it.
 	checkBanStore(ipNet2, false, 0, 0)
+
+	// Test UnbanIPNet.
+	require.NoError(t, banStore.UnbanIPNet(ipNet1))
```
Crypt-iQ, line 129: it was unbanned above, should ban it again and then check that `UnbanIPNet` unbans it

neutrino.go
```diff
@@ -1081,6 +1081,24 @@ func (s *ChainService) BanPeer(addr string, reason banman.Reason) error {
...
+		return fmt.Errorf("unable to parse IP network for peer %v: %v",
+			addr, err)
+	}
+
+	err = s.banStore.UnbanIPNet(ipNet)
+	if err != nil {
+		return fmt.Errorf("unable to unban peer: %v", err)
+	}
+
+	return s.ConnectNode(addr, parmanent)
```
Crypt-iQ, line 1099: sorry for my earlier suggestion, I think this should be in a goroutine so we aren't blocking?

cc @Roasbeef

Crypt-iQ (commented):
looks really close, some comments

banman/store_test.go
Chinwendu20: I do not think that it is up to an hour, here?

Crypt-iQ: I just mean that we should all `BanIPNet` again before we call `UnbanIPNet` again. I guess I don't understand the question?

Chinwendu20: I mean, it hould still be banned because it is not up to one hour:
https://github.com/lightninglabs/neutrino/blob/eb5942f8b86afe0a51026bfa3bc218fac554e9f5/banman/store_test.go#L114

Crypt-iQ: ah I see, the below `checkBanStore` then needs to be `checkBanStore(ipNet1, false, 0, 0)`

Chinwendu20: Thanks

neutrino.go
Chinwendu20: I think the blocking you are referring to might emanate from here?
https://github.com/lightninglabs/neutrino/blob/9130a5e9308d15f2dd9306974d98deffda277b4d/notifications.go#L351-L360
From what I have seen from other exported functions on the chainservice, I do not think they take into account this blocking?

Crypt-iQ: See above in `BanPeer`, the disconnect is in a goroutine

Chinwendu20: Thanks for pointing this out. The `Disconnect` in that function is quite different as it is not a chain service method like what I was referring to. I was referring to other [exported chainservice methods](https://github.com/lightninglabs/neutrino/blob/9130a5e9308d15f2dd9306974d98deffda277b4d/notifications.go#L282-L327) that send a message to the query channel and how I have seen that they do not take into account a potential blocking.

I am not sure if we should continue with that practice.

Crypt-iQ: It depends on how it's used in practice, which it's not here. So I'm fine with it as-is and we can adjust it later based on usage

Crypt-iQ (approved):

Chinwendu20:
@guggero I think we can merge this now?
Here is the upstream PR for lnd: https://github.com/lightningnetwork/lnd/pull/7606

PR #288, opened 2023-09-02
neutrino: Added ResetRanking method to PeerRanking.

### LINKED ISSUE #280: refresh peer's rank on reconnect
In `peer_rank.go` we never refresh a peer's rank if we reconnect. It will use the old ranking we had for it. If we try to query a tor node and penalize it due to high latency, we should allow reconnecting to refresh the rank:
https://github.com/lightninglabs/neutrino/blob/cb61d7f9ff0381415acd88b841f74bfeb4b87e80/query/peer_rank.go#L57-L61

Chinwendu20:
This PR addresses https://github.com/lightninglabs/neutrino/issues/280

Fixes #280

### DIFF
--- a/query/peer_rank.go
+++ b/query/peer_rank.go
@@ -91,3 +91,13 @@ func (p *peerRanking) Reward(peer string) {
 
 	p.rank[peer] = score - 1
 }
+
+// ResetRanking sets the score of the passed peer to the defaultScore.
+func (p *peerRanking) ResetRanking(peer string) {
+	_, ok := p.rank[peer]
+	if !ok {
+		return
+	}
+
+	p.rank[peer] = defaultScore
+}
--- a/query/workmanager.go
+++ b/query/workmanager.go
@@ -65,8 +65,11 @@ type PeerRanking interface {
 	// queries.
 	Punish(peer string)
 
-	// Order sorst the slice of peers according to their ranking.
+	// Order sorts the slice of peers according to their ranking.
 	Order(peers []string)
+
+	// ResetRanking sets the score of the passed peer to the defaultScore.
+	ResetRanking(peerAddr string)
 }
 
 // activeWorker wraps a Worker that is currently running, together with the job
@@ -375,6 +378,11 @@ Loop:
 						newTimeout = maxQueryTimeout
 					}
 					result.job.timeout = newTimeout
+				}
+
+				// Refresh peer rank on disconnect.
+				if result.err == ErrPeerDisconnected {
+					w.cfg.Ranking.ResetRanking(result.peer.Addr())
 				}
 
 				heap.Push(work, result.job)
--- a/query/peer_rank_test.go
+++ b/query/peer_rank_test.go
@@ -46,14 +46,35 @@ func TestPeerRank(t *testing.T) {
 		}
 	}
 
-	// Lastly, reward the lowest scored one a bunch, which should move it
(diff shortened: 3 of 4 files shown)


Chinwendu20:
@Crypt-iQ, this is the PR.

query/workmanager.go
```diff
@@ -67,6 +67,8 @@ type PeerRanking interface {
 
 	// Order sorst the slice of peers according to their ranking.
 	Order(peers []string)
+
+	ResetRanking(peerAddr string)
```
Roasbeef, line 71: Missing a `godoc` comment.

query/workmanager.go
```diff
@@ -377,6 +379,11 @@ Loop:
 					result.job.timeout = newTimeout
 				}
 
+				// Refresh peer rank on disconnect.
+				if result.err == ErrPeerDisconnected {
+					w.cfg.Ranking.ResetRanking(result.peer.Addr())
```
Roasbeef, line 384: Line should fold to 80 char columns.

query/peer_rank.go
```diff
@@ -91,3 +91,12 @@ func (p *peerRanking) Reward(peer string) {
 
 	p.rank[peer] = score - 1
 }
+
+func (p *peerRanking) ResetRanking(peer string) {
```
Roasbeef, line 95: Missing a `godoc` comment.

Roasbeef (changes requested):

Crypt-iQ (approved):
LGTM once @Roasbeef comments addressed 🎉

Chinwendu20:
Thank you guys but I am.not sure if this PR is still valid. Since there is no need to reset ranking?

Roasbeef:
> Since there is no need to reset ranking?

What makes you think that? IIUC, the original intent was the a peer's ranking isn't necessarily the same between reconnects (changes in network, implementation, hardware, etc, etc). I'd consider the fact that they were never reset to be a bug.

Chinwendu20:
> > Since there is no need to reset ranking?
> 
> What makes you think that? IIUC, the original intent was the a peer's ranking isn't necessarily the same between reconnects (changes in network, implementation, hardware, etc, etc). I'd consider the fact that they were never reset to be a bug.

Oh okay, this comment: https://github.com/lightninglabs/neutrino/issues/280#issuecomment-1703914567, I was just wondering since work would still be distributed, there would be no need for fetch from that from that particular peer which disconnected. 

But I guess the current setup makes for a bug in itself, thanks for pointing it out.

query/workmanager.go
Chinwendu20: Please I checked this in my IDE and it is less than 80 char column

Roasbeef (approved):
LGTM 🍬

query/workmanager.go
guggero: @Chinwendu20 did you configure your IDE to use 8 spaces for a tab character? If not, this will probably pop up in future PRs as well (in `lnd` the line length is even enforced by a linter).

Chinwendu20: Thanks for pointing this out, this was one of my earlier PRs. I have now configured my IDE to this. Does it still appear as going beyond the limit?

guggero: This particular line as it was merged, yes. But no big deal, just wanted to let you know for future PRs.

PR #295, opened 2024-02-18
Add Browser Support

linden:
This PR adds support for the browser using [indexeddb](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) as a storage backend for `headerfs` instead of the file system.

## Why IndexedDB?
I'd initially chosen to go with [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) for simplicity but due to recent rate limiting changes it made syncing past a couple 100 blocks freeze, indexddb does not have share these limitations. I'd also looked into emulating the file system but it turned out to add a lot of complexity with a performance loss over just using indexeddb directly. 

## Browser Testing.
Since Go only provides an executor for your native platform testing needs to be done via a custom executor targeting the browser. I've written my own called [`wasmexec`](https://github.com/linden/wasmexec) which also provides file system access for ease of testing, you can run tests via the following:
```sh
# install wasmexec from the add-file-system branch while #1 is in progress.
go install github.com/linden/wasmexec@add-file-system

# target javascript/wasm and execute via wasmexec.
GOOS=js GOARCH=wasm go test -exec wasmexec ./...
```
I'm planning on provider this as part of a Github Action in the future.

## `btcd` Changes.
Running this in the browser without `wasmexec`'s file system access requires a couple of changes to `btcd`'s `addrmgr` & `btcutil` as they require a file system. Changes made in https://github.com/btcsuite/btcd/pull/2124.

### DIFF
--- a/bamboozle_unit_test.go
+++ b/bamboozle_unit_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/lightninglabs/neutrino/headerfs"
+	_ "github.com/linden/tempdb"
 )
 
 func decodeHashNoError(str string) *chainhash.Hash {
@@ -525,9 +526,7 @@ func runCheckCFCheckptSanityTestCase(t *testing.T, testCase *cfCheckptTestCase) 
 	}
 	defer os.RemoveAll(tempDir)
 
-	db, err := walletdb.Create(
-		"bdb", tempDir+"/weks.db", true, dbOpenTimeout,
-	)
+	db, err := walletdb.Create("tempdb", "weks.db")
 	if err != nil {
 		t.Fatalf("Error opening DB: %s", err)
 	}
--- a/headerfs/file.go
+++ b/headerfs/file.go
@@ -1,18 +1,211 @@
+//go:build !windows && !js && !wasm
+
 package headerfs
 
 import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
+	"sync"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/walletdb"
 )
 
-// ErrHeaderNotFound is returned when a target header on disk (flat file) can't
-// be found.
-type ErrHeaderNotFound struct {
-	error
+// headerBufPool is a pool of bytes.Buffer that will be re-used by the various
+// headerStore implementations to batch their header writes to disk. By
+// utilizing this variable we can minimize the total number of allocations when
+// writing headers to disk.
+var headerBufPool = sync.Pool{
+	New: func() interface{} { return new(bytes.Buffer) },
+}
+
+// headerStore combines a on-disk set of headers within a flat file in addition
+// to a database which indexes that flat file. Together, these two abstractions
+// can be used in order to build an indexed header store for any type of
+// "header" as it deals only with raw bytes, and leaves it to a higher layer to
+// interpret those raw bytes accordingly.
+//
+// TODO(roasbeef): quickcheck coverage.
+type headerStore struct {
+	mtx sync.RWMutex // nolint:structcheck // false positive because used as embedded struct only
+
+	fileName string
+
+	file *os.File
+
+	hType HeaderType
+
+	*headerIndex
+}
+
+// newHeaderStore creates a new headerStore given an already open database, a
+// target file path for the flat-file and a particular header type. The target
+// file will be created as necessary.
+func newHeaderStore(db walletdb.DB, filePath string,
+	hType HeaderType) (*headerStore, error) {
+
+	var flatFileName string
+	switch hType {
+	case Block:
+		flatFileName = "block_headers.bin"
+	case RegularFilter:
+		flatFileName = "reg_filter_headers.bin"
+	default:
+		return nil, fmt.Errorf("unrecognized filter type: %v", hType)
+	}
+
+	flatFileName = filepath.Join(filePath, flatFileName)
+
+	// We'll open the file, creating it if necessary and ensuring that all
+	// writes are actually appends to the end of the file.
+	fileFlags := os.O_RDWR | os.O_APPEND | os.O_CREATE
+	headerFile, err := os.OpenFile(flatFileName, fileFlags, 0644)
+	if err != nil {
+		return nil, err
+	}
+
+	// With the file open, we'll then create the header index so we can
+	// have random access into the flat files.
+	index, err := newHeaderIndex(db, hType)
+	if err != nil {
+		return nil, err
+	}
+
+	return &headerStore{
+		fileName:    flatFileName,
+		file:        headerFile,
+		hType:       hType,
+		headerIndex: index,
+	}, nil
+}
+
+// WriteHeaders writes a set of headers to disk and updates the index in a
+// single atomic transaction.
+//
+// NOTE: Part of the BlockHeaderStore interface.
+func (h *blockHeaderStore) WriteHeaders(hdrs ...BlockHeader) error {
+	// Lock store for write.
+	h.mtx.Lock()
+	defer h.mtx.Unlock()
+
+	// First, we'll grab a buffer from the write buffer pool so we an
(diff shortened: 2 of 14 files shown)


bamboozle_unit_test.go
```diff
@@ -525,9 +526,7 @@ func runCheckCFCheckptSanityTestCase(t *testing.T, testCase *cfCheckptTestCase)
 	}
 	defer os.RemoveAll(tempDir)
 
-	db, err := walletdb.Create(
-		"bdb", tempDir+"/weks.db", true, dbOpenTimeout,
-	)
+	db, err := walletdb.Create("tempdb", "weks.db")
```
Roasbeef, line 529: Perhaps we can instead modify the tests to use a build tag to specify `tempdb` vs `bdb`?

Roasbeef: We could have a `NewTestDB` func defined for the build w/ and w/o the build tag defined.

headerfs/index_test.go
```diff
@@ -173,12 +173,18 @@ func TestHeaderStorageFallback(t *testing.T) {
 		t.Fatalf("error setting new tip: %v", err)
 	}
 	for _, header := range newHeaderEntries {
-		if err := hIndex.truncateIndex(&header.hash, true); err != nil {
+		// Copy the header hash so we can create a iteration-specific pointer.
+		// https://github.com/golang/go/discussions/56010#discussion-4441437.
```
Roasbeef, line 177: If we bump the Go version to 1.22 here, then we can bind the new behavior directly.

headerfs/file.go
```diff
@@ -1,18 +1,211 @@
...
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/walletdb"
 )
 
-// ErrHeaderNotFound is returned when a target header on disk (flat file) can't
-// be found.
-type ErrHeaderNotFound struct {
-	error
+// headerBufPool is a pool of bytes.Buffer that will be re-used by the various
```
Roasbeef, line 17: I think this would be easier to review if one commit moved the contents into this new file, then subsequent commits start to add the build tag and additional refactoring.

headerfs/store.go
```diff
@@ -74,71 +94,33 @@ type BlockHeaderStore interface {
...
 	RollbackLastBlock() (*BlockStamp, error)
-}
 
-// headerBufPool is a pool of bytes.Buffer that will be re-used by the various
-// headerStore implementations to batch their header writes to disk. By
-// utilizing this variable we can minimize the total number of allocations when
-// writing headers to disk.
-var headerBufPool = sync.Pool{
-	New: func() interface{} { return new(bytes.Buffer) },
+	// Close and delete the BlockHeaderStore.
```
Roasbeef, line 98: `godoc` comment should start with the method name

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+
+const headersStore = "headers"
+
+// headerStore combines a IndexedDB store of headers within a flat file in addition
+// to a database which indexes that flat file. Together, these two abstractions
+// can be used in order to build an indexed header store for any type of
+// "header" as it deals only with raw bytes, and leaves it to a higher layer to
+// interpret those raw bytes accordingly.
+//
+// TODO(roasbeef): quickcheck coverage.
```
Roasbeef, line 28: stray todo/copy pasta

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+type headerStore struct {
+	mtx sync.RWMutex // nolint:structcheck // false positive because used as embedded struct only
+
+	idb *indexeddb.DB
```
Roasbeef, line 32: What if we had this package implement the `walletdb` interface? Then we could just drop it in here, as well as some other locations/projects that use that central KV-store abstraction.

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+	tx, err := h.idb.NewTransaction([]string{headersStore}, indexeddb.ReadMode)
+	if err != nil {
+		return 0, false, err
+	}
+
+	// Get the headers store.
+	hdrs := tx.Store(headersStore)
+
+	// Count the amount of headers.
+	count, err := hdrs.Count()
```
Roasbeef, line 94: Under the hood, does this hit a fast index, or will it need to enumerate all the contents to get a final sum?

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+	"github.com/btcsuite/btcwallet/walletdb"
+	"github.com/linden/indexeddb"
+)
+
+const headersStore = "headers"
+
+// headerStore combines a IndexedDB store of headers within a flat file in addition
+// to a database which indexes that flat file. Together, these two abstractions
+// can be used in order to build an indexed header store for any type of
+// "header" as it deals only with raw bytes, and leaves it to a higher layer to
```
Roasbeef, line 25: I think we should revise this `godoc` a bit to note the high level structure of this impl. IIUC, it maps `height => header` within `indexeddb`. It's the based off a similar bucket structure as boltdb, with the top level key here being  `headersStore`.

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+	// Get the height store.
+	hdrs := tx.Store(headersStore)
+
+	// Delete the genesis block.
+	if genesis {
+		return hdrs.Delete("genesis")
+	}
+
+	// Delete the header.
+	return hdrs.Delete(height + 1)
```
Roasbeef, line 131: not zero indexed?

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+		return 0, false, err
+	}
+
+	// Subtract one as the block height does not include the genesis block.
+	return uint32(count - 1), false, nil
+}
+
+func (h *headerStore) singleTruncate() error {
+	// Get the current height.
+	height, genesis, err := h.height()
```
Roasbeef, line 111: Need to check out the underlying implementation, but: should the height be obtained _after_ the transaction is created? Otherwise, two writers may start the transaction with an inconsistent view.

Roasbeef (changes requested):
Nice work! This is super exciting, we have a number of other projects in the background that stand to tremendously benefit from this line of work. 

Did an initial round of review, will also check out the source of the `indexeddb` wrapper you created so I can get a handle on what type of concurrency control we can count on.

headerfs/file.go
```diff
@@ -1,18 +1,211 @@
+//go:build !windows && !js && !wasm
```
Chinwendu20, line 1: Why is windows excluded?

headerfs/store.go
```diff
@@ -595,170 +438,91 @@ func (h *blockHeaderStore) ChainTip() (*wire.BlockHeader, uint32, error) {
...
-	fileInfo, err := fStore.file.Stat()
+	height, genesis, err := hStore.height()
 	if err != nil {
 		return nil, err
 	}
 
-	fhs := &FilterHeaderStore{
-		fStore,
+	bhs := &blockHeaderStore{
+		headerStore: hStore,
```
Chinwendu20, line 462: Do you think we can make this headerStore field an interface? since we want to interact with the methods only and there are currently two implementations for this. Each implementation smartly separated by your use of build tags.

headerfs/js.go
```diff
@@ -0,0 +1,402 @@
...
+	idb *indexeddb.DB
+
+	*headerIndex
+}
+
+// readHeader reads a full block header from the flat-file. The header read is
+// determined by the hight value.
+func (h *headerStore) readHeader(height uint32) (wire.BlockHeader, error) {
+	// Create a new read/write transaction, scoped to headers.
+	tx, err := h.idb.NewTransaction([]string{headersStore}, indexeddb.ReadWriteMode)
```
Chinwendu20, line 41: Do you think this should transaction should be created in ReadMode only?

Chinwendu20 (commented):
Thanks for the PR. I left some comments and suggestions. The commit title should be:
package name: Message.

guggero:
!lightninglabs-deploy mute

linden:
closing this PR for now, Go's lack of multi-threading in browser wasm has made a large amount of `github.com/lightninglabs/taproot-assets` unrunnable - which was our business justification. 

we plan to pick this back up as soon as Go and any mainstream browser supports this.

PR #296, opened 2024-03-24
headerfs: sync file after writing raw header in `appendRaw` function

### LINKED ISSUE #276: fsync instead of calling `Write`
`Write` doesn't sync to disk so a crash can corrupt data. One example of this is here: https://github.com/lightninglabs/neutrino/blob/4bc2c998c726f6438017769da5222cbe1c88ca92/headerfs/file.go#L19-L25

I forget if there are other call-sites

mohamedawnallah:
## Change Description

Closes #276

### DIFF
--- a/headerfs/file.go
+++ b/headerfs/file.go
@@ -20,6 +20,9 @@ func (h *headerStore) appendRaw(header []byte) error {
 	if _, err := h.file.Write(header); err != nil {
 		return err
 	}
+	if err := h.file.Sync(); err != nil {
+		return err
+	}
 
 	return nil
 }


guggero:
While this implementation is likely correct and probably the thing we _should_ do, I worry about the performance hit this will take. Do we have a benchmark test to see how much of an impact this change has? Perhaps we need to also refactor things a bit to make sure we only call the `appendRaw` method for big chunks of data where it's worth syncing and not for every single entry for example.

mohamedawnallah:
> Do we have a benchmark test to see how much of an impact this change has?

Yes, you're right! It may cause overhead of syncing unless a given benchmark proves otherwise 👍

> Perhaps we need to also refactor things a bit to make sure we only call the appendRaw method for big chunks of data where it's worth syncing and not for every single entry for example.

I will look into this. Thanks for sharing!

Roasbeef:
Doesn't the truncate logic take care of this? https://github.com/lightninglabs/neutrino/blob/27977e1af13774a556fd2e62ebef1f32667a0fdb/headerfs/store.go#L241-L248

Roasbeef:
On second thought, there's a TODO above there that mentions we don't expect the index to be ahead of the file. It's possible to have a distinct loop that'll replace the current index tip with the latest header in the file.

PR #297, opened 2024-03-27
query: fix retry query case.

ziggie1984:
Fixes https://github.com/lightningnetwork/lnd/issues/8593

In general I will rework this code area in LND 19 because this needs an overhaul and relates to https://github.com/btcsuite/btcwallet/issues/904 

In the current code there is an edge case which manifests itself when the used backend has very unstable connections and also not a lot of outbound peers which results in queryJobs waiting a long time in the queue. Then we would timeout the batch here:

https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/workmanager.go#L420-L433

but the queryJob was already registered for a second try here:

https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/workmanager.go#L388-L389

Now this query will be tried all over again and never purged from the queue and due to how the heap map works it will always retry the query with the lowest index number. So this will cause us to never try new queries but always retry the old ones.

### DIFF
--- a/query/workmanager.go
+++ b/query/workmanager.go
@@ -183,6 +183,7 @@ func (w *peerWorkManager) workDispatcher() {
 		timeout    <-chan time.Time
 		rem        int
 		errChan    chan error
+		cancelChan chan struct{}
 	}
 
 	// We set up a batch index counter to keep track of batches that still
@@ -309,7 +310,20 @@ Loop:
 			// turns out to be an error.
 			batchNum := currentQueries[result.job.index]
 			delete(currentQueries, result.job.index)
-			batch := currentBatches[batchNum]
+
+			// In case the batch is already canceled we return
+			// early.
+			batch, ok := currentBatches[batchNum]
+			if !ok {
+				log.Warnf("Query(%d) result from peer %v "+
+					"discarded with retries %d, because "+
+					"batch already canceled: %v",
+					result.job.index,
+					result.peer.Addr(),
+					result.job.tries, result.err)
+
+				continue Loop
+			}
 
 			switch {
 			// If the query ended because it was canceled, drop it.
@@ -322,30 +336,34 @@ Loop:
 				// was canceled, forward the error on the
 				// batch's error channel.  We do this since a
 				// cancellation applies to the whole batch.
-				if batch != nil {
-					batch.errChan <- result.err
-					delete(currentBatches, batchNum)
+				batch.errChan <- result.err
+				delete(currentBatches, batchNum)
 
-					log.Debugf("Canceled batch %v",
-						batchNum)
-					continue Loop
-				}
+				log.Debugf("Canceled batch %v", batchNum)
+				continue Loop
 
 			// If the query ended with any other error, put it back
 			// into the work queue if it has not reached the
 			// maximum number of retries.
 			case result.err != nil:
-				// Punish the peer for the failed query.
-				w.cfg.Ranking.Punish(result.peer.Addr())
+				// Refresh peer rank on disconnect.
+				if result.err == ErrPeerDisconnected {
+					w.cfg.Ranking.ResetRanking(
+						result.peer.Addr(),
+					)
+				} else {
+					// Punish the peer for the failed query.
+					w.cfg.Ranking.Punish(result.peer.Addr())
+				}
 
-				if batch != nil && !batch.noRetryMax {
+				if !batch.noRetryMax {
 					result.job.tries++
 				}
 
 				// Check if this query has reached its maximum
 				// number of retries. If so, remove it from the
 				// batch and don't reschedule it.
-				if batch != nil && !batch.noRetryMax &&
+				if !batch.noRetryMax &&
 					result.job.tries >= batch.maxRetries {
 
 					log.Warnf("Query(%d) from peer %v "+
@@ -380,11 +398,6 @@ Loop:
 					result.job.timeout = newTimeout
 				}
 
-				// Refresh peer rank on disconnect.
-				if result.err == ErrPeerDisconnected {
-					w.cfg.Ranking.ResetRanking(result.peer.Addr())
-				}
-
 				heap.Push(work, result.job)
 				currentQueries[result.job.index] = batchNum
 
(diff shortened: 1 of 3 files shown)


ziggie1984:
Maybe we should also increase the timeouts/retries while we are here, so that we can mitigate cases like this: https://github.com/lightningnetwork/lnd/pull/8497. 
Where the default 2 retries with a timeout of 2 and then 4 seconds just aren't enough also taking the overall timeout of the batch of 30 seconds into account.

2 sec timeout: https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/workmanager.go#L14

4 sec timeout after first failed retry: 

https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/workmanager.go#L375-L379

Overall batchtimeout 30 sec:
https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/interface.go#L10-L12

Retries 2:

https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/interface.go#L18-L20

Chinwendu20 (commented):
Well, thanks a lot for handling this edge case. The fix looks good. I think we should have checked earlier on if batch == nil here: https://github.com/lightninglabs/neutrino/blob/43f5a588eb4a53cb36d118420e186629154086da/query/workmanager.go#L312
 instead of duplicating that check over and over but maybe that can be included in the rework that you suggested.

query/workmanager.go
```diff
@@ -335,8 +335,27 @@ Loop:
 			// into the work queue if it has not reached the
 			// maximum number of retries.
 			case result.err != nil:
-				// Punish the peer for the failed query.
-				w.cfg.Ranking.Punish(result.peer.Addr())
+				// Refresh peer rank on disconnect.
+				if result.err == ErrPeerDisconnected {
+					w.cfg.Ranking.ResetRanking(
+						result.peer.Addr(),
+					)
```
ProofOfKeags, line 342: Do we not want to punish unstable peers?

query/workmanager.go
```diff
@@ -335,8 +335,27 @@ Loop:
...
+					log.Warnf("Query(%d) from peer %v "+
+						"failed with retries %d, NOT "+
+						"rescheduling batch already"+
+						"canceled: %v",
+						result.job.index,
+						result.peer.Addr(),
+						result.job.tries, result.err)
+
+					continue Loop
+				}
```
ProofOfKeags, line 358: 🫡

query/workmanager.go
```diff
@@ -423,11 +437,32 @@ Loop:
...
+					// Otherwise these queries might never
+					// be removed (in case they fail because
+					// of a timeout for example).
+					//
+					// NOTE: We do NOT cancel ongoing
+					// queries with the workers because it's
+					// easier to let them timeout and just
+					// be dropped because their
+					// corresponding batch is already
+					// removed.
```
ProofOfKeags, line 452: Are we sure we want to leave dangling queries like this? I think I understand that it's not necessary for us to kill them, but I wonder if we have the capacity to deal with the results if we leave the queries. Are the outstanding queries guaranteed to timeout? If not, what happens to the results?

ProofOfKeags (commented):
I have a couple of questions. Other than that looks good.

query/workmanager.go
ProofOfKeags: Right but consider how much cognitive load is required to ascertain that. Instead we should just cancel that because now we are relying on behavior defined 100+ LOC away to mutate state in exactly the right way and check those conditions in exactly the right way.

ProofOfKeags: Oof. I don't want to scope creep this, but I want it on the record that relying on shared state for this kind of thing that isn't part of the API of the thing being operated on is not good practice.

ziggie1984: I think we can definitely remove the ongoing jobs as well (we just need to track the active work), is this what you referring to with the shared state ? I am not sure if I understood the concern in detail, may you elaborate ?

ziggie1984:
Ok I came up with a slightly different approach, we will now use a cancel channel and close iit nstead of looping through the queryJob array, maybe we should do both ?. Let me know if its even a more hacky way, tho have to think about a new test for this.

query/workmanager.go
ProofOfKeags: I would disregard what I said. It is out of scope for this change. My concern is that we haven't layered components in neutrino properly and the code is very delicate and requires "whole program understanding" as opposed to being able to reason locally about what is going on. I'm fine with the choices made here as the parts I take issue with precede your changes.

ProofOfKeags:
> Ok I came up with a slightly different approach, we will now use a cancel channel and close it instead of looping through the queryJob array, maybe we should do both (close channel to remove active requests and remove all the one still queued, I think just the first one is enough) ?. Let me know if its even a more hacky way, tho have to think about a new test for this.

Don't worry about it, this is fine enough for a change this small. I would like to see neutrino undergo some refactoring at some point but I don't think we should turn your 1-commit fix into that project.

ziggie1984:
Sounds good, I am writing a test for the new approach and then we ship it until we get the new refactor into the codebase

ziggie1984:
@ProofOfKeags actually the former version had a bug in it, it would crash in some circumstances which was only revealed to me while adding a unit test for this case. We cannot just remove the heap entries for the queries because they might already be registered with the workers. So the prior version would crash.

Good reminder to never ack something without tests...

ProofOfKeags:
> Good reminder to never ack something without tests...

Good catch. I'll admit that the way we do testing makes it really hard for me to tell whether the tests are actually good or not. I'm actively trying to improve our library infrastructure. The better we factorize things, the smaller the tests will be and the easier they will be to evaluate too. It requires a lot of discipline though and time though.

query/workmanager.go
ziggie1984: introduced another design so this comment is obsolete.

Chinwendu20:
> @ProofOfKeags actually the former version had a bug in it, it would crash in some circumstances which was only revealed to me while adding a unit test for this case. We cannot just remove the heap entries for the queries because they might already be registered with the workers. So the prior version would crash.

Interesting, I thought if they had already been registered with a worker, it would have been deleted from the heap, then and so won't be present in the purge you did earlier.
https://github.com/lightninglabs/neutrino/blob/1ef869ffea1a88333dbb2c8e846e384ea6e29390/query/workmanager.go#L244
> 
> Good reminder to never ack something without tests... 😅

A learning point for me as well.

This approach works too, just that we would have to wait to schedule the job that has the batch cancelled already before it gets a cancel signal.

ziggie1984:
> Interesting, I thought if they had already been registered with a worker, it would have been deleted from the heap, then and so won't be present in the purge you did earlier.

Yes so the main problem is that as soon as a job is `Pop()`-ed from the heap it is not immediately removed from the `currentQueries` map, so when the batch times out and registered queries are still ongoing with the workers I would try to delete queries which are already `Pop()`-ed from the heap. I think the way its now, is better because it uses channels in a clean way. But as said at the beginning and also underlined by @ProofOfKeags this code part this a refactoring.

ProofOfKeags (approved):

ellemouton (changes requested):
Great find!

I think this is the incorrect use of the `cancelChannel` api though. A cancel channel given by the caller is a way for the caller to signal to us (the system) that we should cancel early. Ie, we the system should only ever listen on this channel. Else we risk running into a "panic: close of closed channel" error. 

I think we should instead of an internal-only way of canceling queries that is separate from the way that a caller cancels queries

ProofOfKeags:
> I think we should instead have an internal-only way of canceling queries that is separate from the way that a caller cancels queries

Sounds like I should revive my work on the Async API. 😅

ziggie1984:
Nice input @ellemouton 🙏, introduced a new `internalCancelChannel` for the query to distinguish both cases.

query/interface.go
```diff
@@ -57,6 +57,7 @@ func defaultQueryOptions() *queryOptions {
 		timeout:    defaultQueryTimeout,
 		encoding:   defaultQueryEncoding,
 		numRetries: defaultNumRetries,
+		cancelChan: make(chan struct{}),
```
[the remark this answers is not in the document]
ellemouton: in fact, i dont think we want to set it. Cause if the user chooses not to pass in a cancel channel, then we dont want to hang on this accidentally (ie, we might depend on it being nil if the user does not set it)

ellemouton (commented):
Thanks for the quick update! Looks good! One little thing left over from previous version I think though! After that, LGTM!

ellemouton (approved):
LGTM 🚀

ellemouton:
@ziggie1984  - before we merge this, can you open an LND PR that points to this so we can make sure the CI passes?

ziggie1984:
@guggero I think we can merge this now, itests passed on the LND PR (https://github.com/lightningnetwork/lnd/pull/8621)

PR #300, opened 2024-05-29
neutrino: improve perf in high latency networks

djkazic:
After working with @hsjoberg and @niteshbalusu11 to troubleshoot Blixt performance we realized two things were consistently going wrong in networks with > around 300ms latency:

1. Retries would be exhausted quickly
2. Due to that --^ timeouts being doubled did not help much

After investigating, we decided to experiment with changing some constants in neutrino to work better.

This PR addresses both issues by:
- Decreasing maxBatch to 10
- Increasing query retries from 2 to 8
- Increasing rescan retries from 1 to 2
- Increasing broadcast query retries from 1 to 8

The result is that we can now use Blixt with networks that we could not previously.

### DIFF
--- a/neutrino.go
+++ b/neutrino.go
@@ -753,7 +753,7 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	if s.persistToDisk {
 		cfg := &chanutils.BatchWriterConfig[*filterdb.FilterData]{
 			QueueBufferSize:        chanutils.DefaultQueueSize,
-			MaxBatch:               1000,
+			MaxBatch:               10,
 			DBWritesTickerDuration: time.Millisecond * 500,
 			PutItems:               s.FilterDB.PutFilters,
 		}
--- a/rescan.go
+++ b/rescan.go
@@ -956,7 +956,7 @@ func (rs *rescanState) handleBlockConnected(ntfn *blockntfns.Connected) error {
 
 	// Otherwise, we'll attempt to fetch the filter to retrieve the relevant
 	// transactions and notify them.
-	queryOptions := NumRetries(0)
+	queryOptions := NumRetries(2)
 	blockFilter, err := chain.GetCFilter(
 		newStamp.Hash, wire.GCSFilterRegular, queryOptions,
 	)
--- a/query.go
+++ b/query.go
@@ -49,7 +49,7 @@ var (
 	// each peer before we've concluded we aren't going to get a valid
 	// response. This allows to make up for missed messages in some
 	// instances.
-	QueryNumRetries = 2
+	QueryNumRetries = 8
 
 	// QueryPeerConnectTimeout specifies how long to wait for the
 	// underlying chain service to connect to a peer before giving up


heri16 (approved):

djkazic:
OK, it appears that `queryAllPeers` breaks when the `numRetries`  of the `defaultQueryOptions` struct is > 1.

djkazic:
Cool, so I fixed the unit test.

Now, the Blixt team will do some testing to make sure that we haven't also rolled back the perf improvements.

rescan.go
```diff
@@ -956,7 +956,7 @@ func (rs *rescanState) handleBlockConnected(ntfn *blockntfns.Connected) error {
 
 	// Otherwise, we'll attempt to fetch the filter to retrieve the relevant
 	// transactions and notify them.
-	queryOptions := NumRetries(0)
+	queryOptions := NumRetries(2)
```
ellemouton, line 959: would it not make sense for high latency environments to instead make the timeout longer instead of bumping num retries? (or both lol)

also - wondering if we should make this configurable rather than hard code it?

hsjoberg: > would it not make sense for high latency environments to instead make the timeout longer instead of bumping num retries? (or both lol)

Increasing timeout is kinda a bigger change as the number gets increased exponentially after each try.
I think increasing both retry and timeout makes sense. But I could look into that in a follow-up PR instead.

> also - wondering if we should make this configurable rather than hard code it?

Yeah, I've thought of that too. It would be a very nice thing to have.

ellemouton: gotcha gotcha

ellemouton (commented):
cool LGTM - before the final ✅ though, can you open an LND PR that points to this so that we can just check the CI on that side?

djkazic:
> cool LGTM - before the final ✅ though, can you open an LND PR that points to this so that we can just check the CI on that side?

Created here: https://github.com/lightningnetwork/lnd/pull/8797

ellemouton:
cool thanks - will ACK this once the neutrino test on that side is green

ellemouton (approved):
Looking good on the LND side - thanks!

neutrino.go
```diff
@@ -753,7 +753,7 @@ func NewChainService(cfg Config) (*ChainService, error) {
 	if s.persistToDisk {
 		cfg := &chanutils.BatchWriterConfig[*filterdb.FilterData]{
 			QueueBufferSize:        chanutils.DefaultQueueSize,
-			MaxBatch:               1000,
+			MaxBatch:               10,
```
guggero, line 756: Won't this significantly reduce performance on fast systems? Since we now need to commit 100 times more often to disk, always potentially hitting a file system sync?
If this indeed fixes a problem in a slow environment, that makes me think we should make the value configurable through the `lnd` config. That way mobile users can tune the value down while Neutrino users on powerful machines would leave the original value.

guggero (commented):
One question about disk performance on fast systems, otherwise LGTM.

neutrino.go
djkazic: In theory yes but we tested on a desktop neutrino node and observed that in practice, batch sizes were often 1.

So 10 gives us some breathing room.

Sync times were not affected.

This only sets the cap anyway, I think the root of the problem is actually how we determine batch size at runtime.

guggero (approved):

PR #308, opened 2025-02-17
query: add OnMaxTries

kcalvinalvin:
OnMaxTries allows an outside caller to dictate what should happen to a given worker in case of a timeout.
This allows for peer selection strategies to be made outside of the package, allowing for greater flexibility.

EDIT:
This PR is a dependency for https://github.com/btcsuite/btcd/pull/2226

### DIFF
--- a/query/workmanager.go
+++ b/query/workmanager.go
@@ -94,6 +94,11 @@ type Config struct {
 	// make this configurable to easily mock the worker used during tests.
 	NewWorker func(Peer) Worker
 
+	// OnMaxTries gives the caller access to the Peer object once a maximum
+	// number of retries have been attempted. The caller can then access the
+	// Peer's address and can choose to punish the peer accordingly.
+	OnMaxTries func(Peer)
+
 	// Ranking is used to rank the connected peers when determining who to
 	// give work to.
 	Ranking PeerRanking
@@ -380,6 +385,14 @@ Loop:
 
 					log.Debugf("Canceled batch %v",
 						batchNum)
+
+					// Since we've reached this query's
+					// maximum number of retries, now is the
+					// time to call the OnMaxTries callback
+					// function if it isn't nil.
+					if w.cfg.OnMaxTries != nil {
+						w.cfg.OnMaxTries(result.peer)
+					}
 
 					continue Loop
 				}


saubyk:
cc: @mohamedawnallah for review

query/workmanager.go
```diff
@@ -94,6 +94,10 @@ type Config struct {
 	// make this configurable to easily mock the worker used during tests.
 	NewWorker func(Peer) Worker
 
+	// OnMaxTries is function closure that's called on max retries on
+	// workers.
+	OnMaxTries func(string)
```
ellemouton, line 99: think we should instead use a stricter type here (net.Addr) and mention what is being passed to this call-back so that the caller doesnt have to go look at the code to figure out what is being passed to the call-back it is setting. 

The commit message isnt very accurate imo. `OnMaxTries config lets the caller choose what happens when a query goes over the maximum allowed tries` isnt really true - more so: `OnMaxTries` gives the caller access to the peer's address once a maxmum number of retries have been attempted" or something like that

ellemouton: it's worth linking the PR that makes use of this new config option in the PR description so that reviewers can understand the context a bit more

query/worker.go
```diff
@@ -161,6 +161,18 @@ func (w *worker) Run(results chan<- *jobResult, quit <-chan struct{}) {
...
+				select {
+				// If the peer disconnects before giving us a valid
+				// answer, we'll also exit with an error.
+				case <-peer.OnDisconnect():
+					log.Debugf("Peer %v for worker disconnected, "+
+						"cancelling job %v", peer.Addr(),
+						job.Index())
+
+					jobErr = ErrPeerDisconnected
+					break Loop
```
ellemouton, line 173: i dont think this adds anything. This select is already covered in below and so if peer.OnDisconnect() triggers before the message is received on the msgChan, we will already catch the case below.

I think what you instead want is for `job.HandleResp`'s function to know if it can exit or not. 

The idiomatic way of doing this would be to pass it a `ctx context.Context` which gets cancelled when the peer disconnects. An intermediate step can just be to pass it a `quit chan struct{}` which you can derive from the `peer`s `OnDisconnect` method and then listen on that in the HandleResp call-back

ellemouton (changes requested):

query/worker.go
kcalvinalvin: So the reason for adding this is because I saw that there are cases where `case resp := <- msgChan:` gets triggered first before the below `case <-peer.OnDisconnect():` gets triggered. 

> I think what you instead want is for job.HandleResp's function to know if it can exit or not.

Yeah that would also solve the problem I was seeing.

> An intermediate step can just be to pass it a quit chan struct{} which you can derive from the peers OnDisconnect method

That would be ok as well except now we force every implementation of `HandleResp` to handle the case where `quit chan struct{}` is passed.

kcalvinalvin: Well, I'll also add that I don't really have a preference on how that's handled so I'll push changes so that `HandleResp` takes a `quti chan struct{}` as an argument.

query/workmanager.go
kcalvinalvin: > think we should instead use a stricter type here (net.Addr) and mention what is being passed to this call-back so that the caller doesnt have to go look at the code to figure out what is being passed to the call-back it is setting.
> 
> The commit message isnt very accurate imo. `OnMaxTries config lets the caller choose what happens when a query goes over the maximum allowed tries` isnt really true - more so: `OnMaxTries` gives the caller access to the peer's address once a maxmum number of retries have been attempted" or something like that

Will address both points.

kcalvinalvin: > think we should instead use a stricter type here (net.Addr) and mention what is being passed to this call-back so that the caller doesnt have to go look at the code to figure out what is being passed to the call-back it is setting.

On a second look, I don't think I should address this since all of the existing code in the neutrino codebase is based off of the peer as a string value.

You could modify the `Peer` interface to have a `NetAddr() net.Addr` but I think it's more prudent to modify the entire neutrino code to use `net.Addr` instead of having this exception just for this function.

ellemouton: > On a second look, I don't think I should address this since all of the existing code in the neutrino codebase is based off of the peer as a string value.

I dont think we should design off of that. Rather use the proper type here and then in the neutrino code that uses this we can convert to a string. No need to update the Peer interface there, just convert what this call-back provides into a string at call-time

blockmanager.go
```diff
@@ -829,7 +829,13 @@ func (c *checkpointedCFHeadersQuery) requests() []*query.Request {
...
 // CFHeaders query.
 func (c *checkpointedCFHeadersQuery) handleResponse(req, resp wire.Message,
-	peerAddr string) query.Progress {
+	peerAddr string, quit <-chan struct{}) query.Progress {
+
+	select {
+	case <-quit:
+		return noProgress
+	default:
+	}
```
ellemouton, line 838: it doesnt need to be checked here, it should instead be checked on the `select` below on line #918. As it currently stands, things will still hang if `quit` is closed right after this `select`. 

None of the code between here and there is blocking, so can just check there which is the actual "hang" point.

query.go
```diff
@@ -839,7 +845,13 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
 	// handleResp will be called for each message received from a peer. It
 	// will be used to signal to the work manager whether progress has been
 	// made or not.
-	handleResp := func(req, resp wire.Message, peer string) query.Progress {
+	handleResp := func(req, resp wire.Message, peer string, quit <-chan struct{}) query.Progress {
+		select {
+		case <-quit:
+			return noProgress
```
ellemouton, line 851: same here, rather move this to the select on L #940 where things might actually hang

ellemouton (changes requested):
almost there 🙏

query/workmanager.go
```diff
@@ -381,6 +386,18 @@ Loop:
 					log.Debugf("Canceled batch %v",
 						batchNum)
 
+					if w.cfg.OnMaxTries != nil {
+						// We should not error out here
+						// because the peer has already
+						// been connected to. This is
+						// just converting string to
+						// net.Addr.
```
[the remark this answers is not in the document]
kcalvinalvin: > aren't we calling this specifically cause we perhaps havent been able to connect to the peer? ie we could expect an error below.

No that would never happen as an error for not being able to connect to a peer would've been caught long before this callback can trigger. We're specifically looking for slow nodes that don't respond to our message in time.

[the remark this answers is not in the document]
kcalvinalvin: > I think to avoid needing to do this resolution here, just let `OnMaxTries` take a `Peer` object itself. then we are also not restricting the caller to only getting access to an address (it might not easily be able to tell which address that peer belongs to - so if we provide it the whole peer, it can then do as it pleases)

Sure. I'll change it

PR #313, opened 2025-05-08
headerfs: fail gracefully on header write

mohamedawnallah:
## Description

This PR improves error handling by failing gracefully when header writes fail. This helps prevent unexpected `EOF` errors during the process.

### DIFF
--- a/headerfs/file_test.go
+++ b/headerfs/file_test.go
@@ -1,0 +1,179 @@
+package headerfs
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestAppendRow(t *testing.T) {
+	tests := []struct {
+		name          string
+		initialData   []byte
+		headerToWrite []byte
+		writeFn       func([]byte, FileInterface) (int, error)
+		truncFn       func(size int64) error
+		expected      []byte
+		wantErr       bool
+		errMsg        string
+	}{
+		{
+			name:          "NormalWrite_ValidHeader_DataAppendedSuccessfully",
+			initialData:   []byte{0x01, 0x02, 0x03},
+			headerToWrite: []byte{0x04, 0x05, 0x06},
+			expected:      []byte{0x01, 0x02, 0x03, 0x04, 0x05, 0x06},
+		},
+		{
+			name:          "WriteError_ZeroBytesWritten_OriginalDataPreserved",
+			initialData:   []byte{0x01, 0x02, 0x03},
+			headerToWrite: []byte{0x04, 0x05, 0x06},
+			writeFn: func(p []byte, _ FileInterface) (int, error) {
+				return 0, errors.New("simulated write failure")
+			},
+			expected: []byte{0x01, 0x02, 0x03},
+			wantErr:  true,
+			errMsg:   "simulated write failure",
+		},
+		{
+			name:          "PartialWrite_WriteErrorMidway_RollsBackToOriginal",
+			initialData:   []byte{0x01, 0x02, 0x03},
+			headerToWrite: []byte{0x04, 0x05, 0x06},
+			writeFn: func(p []byte, file FileInterface) (int, error) {
+				// Mock a partial write - write the first two bytes.
+				n, _ := file.Write(p[:2])
+				return n, errors.New("simulated partial write failure")
+			},
+			expected: []byte{0x01, 0x02, 0x03},
+			wantErr:  true,
+			errMsg:   "simulated partial write failure",
+		},
+		{
+			name:          "PartialWrite_TruncateFailure_ReportsCompoundError",
+			initialData:   []byte{0x01, 0x02, 0x03},
+			headerToWrite: []byte{0x04, 0x05, 0x06},
+			writeFn: func(p []byte, file FileInterface) (int, error) {
+				// Mock a partial write - write just the first byte.
+				n, _ := file.Write(p[:1])
+				return n, errors.New("simulated partial write failure")
+			},
+			truncFn: func(size int64) error {
+				return errors.New("simulated truncate failure")
+			},
+			expected: []byte{0x01, 0x02, 0x03, 0x04},
+			wantErr:  true,
+			errMsg: "write failed with simulated partial write failure and " +
+				"recovery failed with: simulated truncate failure",
+		},
+		{
+			name:          "NormalWrite_ValidHeader_DataAppendedSuccessfully",
+			initialData:   []byte{},
+			headerToWrite: []byte{0x01, 0x02, 0x03},
+			expected:      []byte{0x01, 0x02, 0x03},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// Create a temporary file for testing.
+			tmpFile, err := os.CreateTemp(t.TempDir(), "header_store_test")
+			if err != nil {
+				t.Fatalf("Failed to create temp file: %v", err)
+			}
+			defer os.Remove(tmpFile.Name())
+
+			// Write initial data.
+			if _, err := tmpFile.Write(test.initialData); err != nil {
+				t.Fatalf("Failed to write initial data: %v", err)
+			}
+
+			// Reset the file position to the end of initial data.
+			_, err = tmpFile.Seek(int64(len(test.initialData)), io.SeekStart)
+			if err != nil {
+				t.Fatalf("Failed to seek: %v", err)
+			}
+
+			// Create a mock file that wraps the real file.
+			mockFile := &mockFile{
+				File:    tmpFile,
+				writeFn: test.writeFn,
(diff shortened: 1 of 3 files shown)


headerfs/store.go
```diff
@@ -84,6 +85,23 @@ var headerBufPool = sync.Pool{
 	New: func() interface{} { return new(bytes.Buffer) },
 }
 
+// FileInterface defines the minimum file operations needed by headerStore.
+type FileInterface interface {
```
Roasbeef, line 89: Can just call it `File` perhaps?

headerfs/file_test.go
```diff
@@ -0,0 +1,179 @@
...
+		initialData   []byte
+		headerToWrite []byte
+		writeFn       func([]byte, FileInterface) (int, error)
+		truncFn       func(size int64) error
+		expected      []byte
+		wantErr       bool
+		errMsg        string
+	}{
+		{
+			name:          "NormalWrite_ValidHeader_DataAppendedSuccessfully",
```
Roasbeef, line 24: Fwiw if you add spaces in the naem here, then the `go test` tool will automatically add underscores there.

headerfs/file_test.go
```diff
@@ -0,0 +1,179 @@
...
+			if _, err := tmpFile.Seek(0, io.SeekStart); err != nil {
+				t.Fatalf("Failed to seek to start: %v", err)
+			}
+
+			// Read the file contents.
+			actualData, err := io.ReadAll(tmpFile)
+			if err != nil {
+				t.Fatalf("Failed to read file: %v", err)
+			}
+
```
Roasbeef, line 135: Related to my other suggestion: can we also test that if we go down in a partially truncated state, then once we come up we're able to recover? May warrant an entirely distinct change.

headerfs/file_test.go
```diff
@@ -0,0 +1,179 @@
...
+import (
+	"bytes"
+	"errors"
+	"io"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestAppendRow(t *testing.T) {
```
Roasbeef, line 12: Missing a `godoc` comment.

headerfs/file_test.go
mohamedawnallah: My current understanding is:

- A `Truncate` operation is a fatal error.
- If a `Truncate` error occurs, it likely indicates a serious issue.
- The probability of both a partial write failure **and** a truncate failure occurring together is extremely low and would typically require manual intervention.

Possible mitigation:

- Implement a recovery check during reads:
    - If the data read is not a complete header (32 bytes for the filter header, 80 bytes for the index), remove it from the binary file.
    - The idea is to handle incomplete or invalid entries immediately (i.e., perform a delete operation during read), so the same issue doesn't recur on the next startup. This also include a delete operation in read operation
    - This serves as a recovery mechanism if the data is recoverable, depending on the mix of operations.
    - However, if the delete operation itself fails, it's unclear how recovery would be handled in that scenario.
- If we simply choose to ignore the incomplete or invalid entry, we risk leaving the binary file in an inconsistent state: the invalid data would remain in the headers binary file, yet we might still be able to read it successfully.
- This approach assumes the partial write entry was not fully truncated, since a truncate failure should have already been reported as an error.
- My concern is that this method potentially mask the root cause of the issue, rather than addressing it directly.

What do you think?

Roasbeef: > Request missing or invalid/incomplete tail header from peers upon detection on reads and index it in the store.

I don't think this makes sense. At this point, we have already fetched the headers. 

> The probability of both a partial write failure and a truncate failure occurring together is probably unlikely and would typically require manual intervention.

This is precisely what we're attempting to solve. Consider that a user running w/o this patch might have a partial write. Today we require them to go in and delete the file manually. On a mobile platform, this isn't feasible for end users. 

If we can make sure that both we write properly, and we can recover from botched writes (rn we fail on read), then we're able to cover all bases.

mohamedawnallah: > This is precisely what we're attempting to solve. Consider that a user running w/o this patch might have a partial write. Today we require them to go in and delete the file manually. On a mobile platform, this isn't feasible for end users. If we can make sure that both we write properly, and we can recover from botched writes (rn we fail on read), then we're able to cover all bases.

Thanks that makes sense. I have opened this issue so we could track the recovery on read for partial written headers #315. That said if you wanna me to approach that recovery on read in certain way would be happy to hear

mohamedawnallah:
> Can we make a basic benchmark here to gauge if this affects write perf in a traditional set up?

That makes sense, thanks. I have added that `BenchmarkHeaderStoreAppendRaw` benchmarking test. Here is the report:

## Without this PR (commit afcfeb1)
$ go test -bench=BenchmarkHeaderStoreAppendRaw -benchmem
goos: linux
goarch: amd64
pkg: github.com/lightninglabs/neutrino/headerfs
cpu: AMD EPYC-Rome-v4 Processor (no XSAVES)
BenchmarkHeaderStoreAppendRaw-4          1083670              1214 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/lightninglabs/neutrino/headerfs      2.521s


## With this PR
$ go test -bench=BenchmarkHeaderStoreAppendRaw -benchmem
goos: linux
goarch: amd64
pkg: github.com/lightninglabs/neutrino/headerfs
cpu: AMD EPYC-Rome-v4 Processor (no XSAVES)
BenchmarkHeaderStoreAppendRaw-4           951910              1305 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/lightninglabs/neutrino/headerfs      2.401s


## Benchmarking Report - Generated by Gemini

Based on the benchmark results:

**Before the change:**
- 1,083,670 operations
- 1,214 ns/op
- 0 B/op (no memory allocations)

**After the change:**
- 951,910 operations
- 1,305 ns/op
- 0 B/op (no memory allocations)

**Analysis:**
1. **Performance impact:** The new implementation is about 7.5% slower (1,305 ns vs 1,214 ns per operation)
2. **Throughput decrease:** The number of operations completed decreased by about 12.2%
3. **Memory usage:** Both implementations are equally memory-efficient with zero allocations

**Conclusion:**
The added error handling with the `Seek` operation does introduce a small performance overhead of approximately 91 nanoseconds per operation. Given that:

1. This is a very modest performance impact (~7.5%)
2. The code now properly handles partial writes and maintains data integrity
3. The error messages are much more informative
4. There's still no memory allocation overhead

The tradeoff appears reasonable. The performance cost is minimal while the reliability improvements are significant. For a storage operation where data integrity is critical, this seems like an acceptable performance trade-off to gain much better error handling and recovery capability.

mohamedawnallah:
From the benchmarking results, it appears there's overhead on the seek operation. However, this is somehow misleading because the benchmark calls `appendRaw` on only a single header (creating a one-to-one mapping between seek operations and headers). In the actual codebase, this operation is performed on batches, meaning one seek operation handles multiple headers. Therefore, the observed 7.5% performance impact is significantly lower in real-world usage.

I think integrating this code change in the codebase would be fine. What do you think @Roasbeef?

Roasbeef (approved):
LGTM 🌾

ziggie1984:
Added the column linter here: https://github.com/lightninglabs/neutrino/pull/318

headerfs/file_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// Create a temporary file for testing.
+			tmpFile, err := os.CreateTemp(t.TempDir(), "header_store_test")
+			if err != nil {
+				t.Fatalf("Failed to create temp file: %v", err)
+			}
+			defer os.Remove(tmpFile.Name())
```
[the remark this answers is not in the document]
ziggie1984: add a function to create the file something like this:

```
func createFile(t *testing.T, filename string) (*os.File, func()) {
    tmpFile, err := os.CreateTemp(t.TempDir(), filename)
    if err != nil {
        t.Fatalf("Failed to create temp file: %v", err)
    }

    cleanup := func() {
        tmpFile.Close()
        os.Remove(tmpFile.Name())
    }

    return tmpFile, cleanup
}
```

then use it like this:

```
tmpFile, cleanup := createFile(t, "header_store_test")
defer cleanup()
```

headerfs/file_test.go
```diff
@@ -0,0 +1,262 @@
...
+				t.Fatalf("Expected file data: %v, "+
+					"got: %v", test.expected, actualData)
+			}
+		})
+	}
+}
+
+// BenchmarkHeaderStoreAppendRaw measures performance of headerStore.appendRaw
+// by writing 80-byte headers to a file and resetting position between writes
+// to isolate raw append performance from file size effects.
```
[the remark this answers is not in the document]
ziggie1984: do you know what the defaults for bitcoin-core are ?

ziggie1984 (approved):
LGTM, pending godoc.

PR #314, opened 2025-05-08
headerfs: improve error messages for header store

mohamedawnallah:
## Description

Improve error messages for header store regards the write and chain tip operations. This would be useful in debugging for example when the user has corrupt headers data.

Follow-up on #313.

### DIFF
--- a/headerfs/store.go
+++ b/headerfs/store.go
@@ -584,12 +584,12 @@ func (h *blockHeaderStore) ChainTip() (*wire.BlockHeader, uint32, error) {
 
 	_, tipHeight, err := h.chainTip()
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, fmt.Errorf("unable to fetch chain tip: %v", err)
 	}
 
 	latestHeader, err := h.readHeader(tipHeight)
 	if err != nil {
-		return nil, 0, err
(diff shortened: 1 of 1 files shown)


bhandras (approved):
LGTM 🎉

headerfs/store.go
```diff
@@ -890,7 +890,7 @@ func (f *FilterHeaderStore) ChainTip() (*chainhash.Hash, uint32, error) {
 
 	latestHeader, err := f.readHeader(tipHeight)
 	if err != nil {
-		return nil, 0, fmt.Errorf("unable to read header: %v", err)
+		return nil, 0, fmt.Errorf("unable to read filter header: %v", err)
```
hieblmi, line 893: did you mean `block` instead of `filter`?

mohamedawnallah: `filter` as it is currently. This is related to `FilterHeaderStore`?

mohamedawnallah: `filter` as it is currently. This is related to `FilterHeaderStore.readHeader`?

hieblmi: Ah I see, looks good.

hieblmi (approved):

PR #316, opened 2025-05-17
fix(query/worker): replace ineffective break with continue in job can…

GarmashAlex:
…cel check

Replaced inefficient break statements with continue statements inside the first select block in the worker's Run function. Now, when a job is canceled (job.cancelChan or job.internalCancelChan), execution correctly moves to the next iteration of the outer loop, which prevents possible logical errors.


saubyk:
Inconsequential pr, private profile, likely a commit farmer.

PR #320, opened 2025-06-14
chainimport[1/3]: import block and filter headers on startup before falling back to P2P synchronization

mohamedawnallah:
## Description

This PR serves two main purposes:
1. Refactor related code and remove duplicated logic
2. Fail gracefully on DB header writes

Towards https://github.com/lightninglabs/neutrino/issues/70.
Next #317.

## Motivation and Context

This opportunity for improvement was identified while addressing Oli's feedback https://github.com/lightninglabs/neutrino/pull/317#discussion_r2132103740.


guggero:
Yeah, looks ready. Assigned myself and @ziggie1984 for review.

guggero:
@mohamedawnallah it looks like in addition to the config file you added, the App also needs to be installed from the marketplace. And I'm not sure if that will do it for the whole organization or not, so I'd rather ask @Roasbeef to look into that (I might not have the permissions anyway).

headerfs/store.go
```diff
@@ -538,7 +538,35 @@ func (h *blockHeaderStore) WriteHeaders(hdrs ...BlockHeader) error {
...
+	// to roll back any changes to the header file to maintain consistency.
+	// The rollback process bases on the number of header serialized. If
+	// both the initial operation and the rollback fail, we return
+	// a detailed error explaining both failures to aid in debugging.
+	if err := h.addHeaders(headerLocs); err != nil {
+		// Since the probability of failing to write to the database is
+		// very low, it is mostly worth the cost of file sync operation
+		// to make sure truncate headers does it correctly.
+		if err := h.file.Sync(); err != nil {
+			return fmt.Errorf("failed to sync block headers "+
```
guggero, line 551: If we can't sync, then we return here and don't truncate. I guess there's not much we can do anyway as this would hint at file system issues on the machine. But I'm wondering if we should just log this error and try to truncate anyway? Same question in `filterHeaderStore.WriteHeaders()`.

guggero (commented):
Very nice, thanks a lot for the improved commit structure. Now ready very well!
Just one final question, other than that LGTM :tada:

headerfs/store.go
guggero: Okay, makes sense. Thanks for the explanation.

guggero (approved):
LGTM 🎉

headerfs/truncate.go
```diff
@@ -22,18 +22,27 @@ func (h *headerStore) singleTruncate() error {
...
-	default:
-		return fmt.Errorf("unknown index type: %v", h.indexType)
+	truncateLength, err := h.indexType.Size()
+	if err != nil {
+		return err
+	}
+	newSize := fileSize - int64(truncateLength)
+
+	// On Windows, we need to close, truncate, and reopen the file.
+	if runtime.GOOS == "windows" {
```
Roasbeef: Not bad, saves some extra files, while also putting more of the logic in line vs normal build tags.

headerfs/index.go
```diff
@@ -417,26 +432,66 @@ func getHeaderEntryFallback(rootBucket walletdb.ReadBucket,
...
-		return ErrHashNotFound
+	// Group hashes by their sub-bucket for more efficient deletion.
+	bySubBucket := make(map[string][]*chainhash.Hash)
+	rootBucketHashes := make([]*chainhash.Hash, 0, len(headerHashes))
+
+	// Check which hashes are in the root bucket and group the rest by their
+	// sub-bucket prefix.
+	for _, hash := range headerHashes {
+		// Convert hash to bytes for DB operations.
+		hashBytes := hash.CloneBytes()
```
Roasbeef, line 454: 👍

Roasbeef (approved):
LGTM 🦧

PR #335, opened 2026-02-04
blockmanager: reset header state after chainimport

Roasbeef:
When headers are imported via `chainimport` in `ChainService.Start()`, the block manager's internal state (`headerList`, `headerTip`, `headerTipHash`, `filterHeaderTip`, `filterHeaderTipHash`) still reflects the pre-import chain tips from when it was constructed in `NewChainService()`. Since the block manager is created before `Start()` runs the import, its in-memory tracking state becomes stale once the import writes headers to the stores. This causes the block manager to build P2P getheaders locators using the genesis hash, preventing it from fetching any remaining blocks via P2P after import completes.

This was discovered while wiring the `chainimport` feature into lnd's integration tests. A fresh neutrino node configured with both `HeadersImport` and peer connections would successfully import headers from file but then fail to sync any additional blocks mined after the import snapshot. The neutrino logs showed `getheaders` requests with a genesis locator and `cfheaders` fetch failures, confirming the block manager was unaware of the imported headers.

The fix adds `ResetHeaderState()` to the block manager, which re-reads the block and filter header chain tips from the stores and reinitializes all internal tracking fields (`headerList`, `headerTip`, `headerTipHash`, `nextCheckpoint`, `filterHeaderTip`, `filterHeaderTipHash`). This method is called in `Start()` immediately after a successful import, before the block manager begins its P2P sync loop.

A new test `TestNeutrinoImportThenP2PSync` exercises this exact scenario: a single `ChainService` with both `HeadersImport` and `AddPeers` configured imports 800 blocks from file, then must sync 50 additional blocks via P2P. Without the fix, the test times out waiting for sync; with the fix, the service correctly builds locators from the imported tip and catches up via P2P.

Depends on the `headerlist.ResetHeaderState()` method added in the chainimport PRs (#317, #320, #324).

---

## Follow-on import performance work

While testing the reset-state fix end-to-end against an lnd local-replace build on mainnet and testnet3, profiling surfaced two hotspots that dominated the import-plus-catch-up wall time well after the locator bug was fixed:

1. The bbolt-backed header-index spent most of its write time creating and re-resolving per-hash sub-buckets, one `CreateBucketIfNotExists` call per imported header.
2. Once the import completed, `lightHeaderCtx.RelativeAncestorCtx` ran a linear walk back from the chain tip for every header validated. On testnet that meant a 2016-step scan per block in every difficulty-adjustment window, and the profile pinned `findPrevTestNetDifficulty` and `checkHeaderSanity` at roughly 10s of cumulative CPU each.

The five commits stacked on top of the original fix address both:

- **`chainimport: bump default header import batch size to 65536`** — 4× the import write batch size.
- **`headerfs: pre-create header index sub-buckets`** — eagerly create all 65 536 two-byte hash-prefix sub-buckets once per database (gated by a new `index-sub-buckets-ready` marker key), then specialize the hot `addHeaders` loop to cache the resolved sub-bucket across consecutive sorted entries.
- **`headerlist: add skip-list ancestor lookups to Node`** — add a single skip pointer to `headerlist.Node` and a matching `Ancestor(height int32)` method in the style of Bitcoin Core's `CBlockIndex`, with the wrap-around-safe staleness checks needed for `BoundedMemoryChain`. Covered by deterministic plus `rapid` property tests and head-to-head skip-vs-prev benchmarks. Also fixes a latent `maxSize == 1` self-cycle in `PushBack` that the property test exposed.
- **`blockmanager: use skip-list ancestor in lightHeaderCtx`** — rewrite `RelativeAncestorCtx` to traverse the new skip-list, threading the resolved `headerlist.Node` into the returned context so subsequent ancestor walks keep amortizing along the skip-list instead of restarting from the tip.
- **`blockmanager: test lightHeaderCtx skip-list and store-fallback paths`** — unit tests that lock down the skip-list path (in-memory only, store never touched, node propagated) and the documented store-fallback path (store consulted, node `nil`).
- **`sync_test: align headers-import benchmark to fast-add validation`** — pass `blockchain.BFFastAdd` in the headers-import benchmark so the measured path matches the block-dn flow chainimport actually runs.

### Measured impact

**Header-import write path** — `BenchmarkHeadersImport` on the bundled 100k mainnet fixture, with `BFFastAdd` validation:

| Batch size | Baseline (no eager buckets) | Eager sub-buckets |
| ---: | ---: | ---: |
| 16 384 | 195 940 hdrs/s | 193 031 hdrs/s |
| 32 768 | 220 237 hdrs/s | 235 562 hdrs/s |
| **65 536** | 213 697 hdrs/s | **260 131 hdrs/s** |
| 131 072 | 203 261 hdrs/s | 279 914 hdrs/s |

Eager sub-bucket pre-creation is what unlocks the larger batch: at 16 384 it is roughly neutral, but at 65 536 it lifts throughput by ~22% over the same batch without the change. 65 536 captures most of the win at roughly half the working set of 131 072.

**Skip-list ancestor lookup** — focused micro-benchmark over a 10 000-node `BoundedMemoryChain` with a spread of target heights:

| Path | ns/op | allocs/op |
| --- | ---: | ---: |
| `Node.Ancestor` (skip-list) | ~1 200 | 0 |
| `Node.Prev()` linear walk | ~96 000 | 0 |

About 80× faster with zero allocations on either path.

**End-to-end CPU profile (lnd, testnet3 → block-dn 4 700 000 source, skip-list + 65 536 batch + eager sub-buckets)**:

| Function | Pre-skip cum CPU | Post-skip cum CPU |
| --- | ---: | ---: |
| `lightHeaderCtx.RelativeAncestorCtx` | 10.52 s | 0.08 s |
| `findPrevTestNetDifficulty` | ~10.47 s | 0.09 s |
| `blockManager.checkHeaderSanity` | ~10.73 s | 0.09 s |

The remaining CPU is back in the bbolt write path inside `headerfs.(*blockHeaderStore).WriteHeaders`.

**End-to-end lnd time-to-synced** (HTTP source `block-dn.org`, local-file source after caching):

| Scenario | Before* | After |
| --- | ---: | ---: |
| Mainnet, 900k headers (HTTP) | 54 s | 41 s |
| Mainnet, 900k headers (local file) | n/a | 17 s |
| Testnet3, 4.7M headers (HTTP) | 5 m 17 s | 3 m 11 s |
| Testnet3, 4.7M headers (local file) | n/a | 2 m 39 s |

*"Before" is the same branch with the original 16 384 batch size and no skip-list. Local-file numbers remove HTTP variance and isolate the write/validation path.

### Verification

- `go test ./headerlist ./headerfs ./chainimport`
- `go test . -run '^TestNeutrinoImportThenP2PSync$' -count=1`
- `go test . -run '^TestLightHeaderCtxRelativeAncestor' -count=1`
- `go test ./...`
- `golangci-lint run --new-from-rev=master --timeout=10m ./...` (clean against the pinned `v1.64.6` from `tools/go.mod`)
- lnd sanity with local neutrino replace: `go test . ./lncfg` and `go test -tags=integration ./itest -run '^$'`

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -282,6 +282,49 @@ func newBlockManager(cfg *blockManagerCfg) (*blockManager, error) {
 	return &bm, nil
 }
 
+// ResetHeaderState re-reads the chain tips from the header stores and
+// reinitializes the block manager's internal tracking state. This must be
+// called after headers have been imported into the stores outside of the
+// block manager (e.g., via chainimport) but before the block manager is
+// started, so that it begins syncing from the correct chain tip rather
+// than the stale state captured at construction time.
+func (b *blockManager) ResetHeaderState() error {
+	// Re-read the block header chain tip from the store.
+	header, height, err := b.cfg.BlockHeaders.ChainTip()
+	if err != nil {
+		return fmt.Errorf("failed to read block header chain "+
+			"tip: %w", err)
+	}
+	b.nextCheckpoint = b.findNextHeaderCheckpoint(int32(height))
+	b.headerList.ResetHeaderState(headerlist.Node{
+		Header: *header,
+		Height: int32(height),
+	})
+	b.headerTip = height
+	b.headerTipHash = header.BlockHash()
+
+	// Re-read the filter header chain tip from the store.
+	_, b.filterHeaderTip, err = b.cfg.RegFilterHeaders.ChainTip()
+	if err != nil {
+		return fmt.Errorf("failed to read filter header chain "+
+			"tip: %w", err)
+	}
+
+	// Ensure the filter header tip hash is set to the block hash at the
+	// filter tip height.
+	fh, err := b.cfg.BlockHeaders.FetchHeaderByHeight(b.filterHeaderTip)
+	if err != nil {
+		return fmt.Errorf("failed to fetch block header at filter "+
+			"tip height %d: %w", b.filterHeaderTip, err)
+	}
+	b.filterHeaderTipHash = fh.BlockHash()
+
+	log.Infof("Block manager header state reset: block tip=%d, "+
+		"filter tip=%d", height, b.filterHeaderTip)
+
+	return nil
+}
+
 // Start begins the core block handler which processes block and inv messages.
 func (b *blockManager) Start() {
 	// Already started?
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -2989,6 +2989,12 @@ type lightHeaderCtx struct {
 
 	store      headerfs.BlockHeaderStore
 	headerList headerlist.Chain
+
+	// node, if non-nil, is the headerList node corresponding to this
+	// header. When set, RelativeAncestorCtx will use the node's skip-list
+	// pointer for O(log n) ancestor lookups instead of walking back from
+	// the chain tip on every call.
+	node *headerlist.Node
 }
 
 // newLightHeaderCtx returns an instance of a lightHeaderCtx to be used when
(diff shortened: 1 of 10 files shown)


blockmanager.go
```diff
@@ -282,6 +282,49 @@ func newBlockManager(cfg *blockManagerCfg) (*blockManager, error) {
 	return &bm, nil
 }
 
+// ResetHeaderState re-reads the chain tips from the header stores and
+// reinitializes the block manager's internal tracking state. This must be
+// called after headers have been imported into the stores outside of the
+// block manager (e.g., via chainimport) but before the block manager is
+// started, so that it begins syncing from the correct chain tip rather
```
haanhvu, line 289: Following the comment that this must be called before the block manager is started, should we explicitly enforce that here?
```
if atomic.LoadInt32(&b.started) != 0 {
    return fmt.Errorf("ResetHeaderState must be called before Start")
}
```

blockmanager.go
```diff
@@ -282,6 +282,49 @@ func newBlockManager(cfg *blockManagerCfg) (*blockManager, error) {
...
+	// filter tip height.
+	fh, err := b.cfg.BlockHeaders.FetchHeaderByHeight(b.filterHeaderTip)
+	if err != nil {
+		return fmt.Errorf("failed to fetch block header at filter "+
+			"tip height %d: %w", b.filterHeaderTip, err)
+	}
+	b.filterHeaderTipHash = fh.BlockHash()
+
+	log.Infof("Block manager header state reset: block tip=%d, "+
+		"filter tip=%d", height, b.filterHeaderTip)
```
haanhvu, line 323: It might be useful to include the tip hashes here as well. Since this is rehydrating state from separate header stores, having both height and hash would make it easier to spot inconsistencies between the block and filter header state, and help debug reorg or partial update scenarios.
```
log.Infof("Block manager header state reset: "+
    "block tip=%d (%s), filter tip=%d (%s)",
    height, b.headerTipHash, b.filterHeaderTip, b.filterHeaderTipHash)
```

litbot-9000:
@yyforyongyu: review reminder

PR #336, opened 2026-02-10
blockmanager: recover from WriteHeaders failure to prevent permanent sync stall

ziggie1984:
## Summary

- When `WriteHeaders` fails in `handleHeadersMsg`, the block manager enters a permanently stuck state: the in-memory `headerList` diverges from disk, no `getheaders` is sent, `cfHandler` blocks forever on `newHeadersSignal.Wait()`, and the sync peer stays connected preventing automatic recovery.
- Fix both the main header sync path and the reorg path by resetting `headerList` from the on-disk chain tip and disconnecting the sync peer, which triggers the existing recovery flow in `handleDonePeerMsg`.
- Add logging to the silent `ChainTip` failure in `handleDonePeerMsg` so recovery failures are observable.

## Test plan

- [x] `TestHandleHeadersWriteFailure` verifies that on `WriteHeaders` failure the peer is disconnected and `headerList` is rolled back to the on-disk chain tip.
- [x] Verify existing tests still pass (`go test ./...`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -2694,6 +2694,26 @@ func (b *blockManager) handleHeadersMsg(hmsg *headersMsg) {
 		err := b.cfg.BlockHeaders.WriteHeaders(headerWriteBatch...)
 		if err != nil {
 			log.Errorf("Unable to write block headers: %v", err)
+
+			// Roll back the in-memory header list to match the
+			// on-disk state, since headers were already appended
+			// to headerList before the write attempt.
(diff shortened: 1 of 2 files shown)


ziggie1984:
to test this on LND version based on LND 20 you need to run the following command to use the right version which is compatible with it:

```
go get github.com/lightninglabs/neutrino@test-rebase-on-v0.16.1
go mod tidy
```

PR #337, opened 2026-02-18
rescan: Retry transient errors during catchup.

JoeGruffins:
When the rescan is in catchup mode (not current with the chain tip), errors from GetBlockHeaderByHeight and notifyBlock were fatal — they caused the rescan goroutine to return the error and exit. This meant that a transient issue like a peer timeout fetching a block would permanently kill the rescan. The btcwallet notification handler would log "Neutrino rescan ended with error" and cancel the block subscription, after which the wallet would never process another block until a manual rescan was triggered.

This was observed on mainnet where a peer timeout during block processing caused the wallet to stall for 15+ hours until the user manually initiated a rescan.

Fix this by retrying both GetBlockHeaderByHeight and notifyBlock errors with a 5-second delay instead of returning them. For notifyBlock failures, the rescan state is rolled back so the same block is retried on the next iteration. Both retry paths respect the quit channel for clean shutdown.

### DIFF
--- a/rescan.go
+++ b/rescan.go
@@ -507,6 +507,13 @@ func (rs *rescanState) rescan() error {
 		// TODO(roasbeef): add exponential back-off
 		blockRetryInterval = time.Millisecond * 100
 		blockRetryQueue    = newBlockRetryQueue()
+
+		// catchupRetryInterval is the interval used to retry block
+		// processing errors during catchup mode (e.g. timeouts
+		// fetching blocks or filters from peers). This uses a longer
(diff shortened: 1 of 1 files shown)


Roasbeef:
Can we add test coverage here to demonstrate the fix?

JoeGruffins:
Oh sorry, will come back to this. We are using a dumber method of just watching if we see a stall and doing a rescan if it does, automatically. Busy atm but will clean this up when I can.

PR #339, opened 2026-03-27
blockmanager: validate BIP 157 stop hashes to prevent peer disconnects during reorgs

### LINKED ISSUE #338: requesting compact filters/headers from bitcoind peers leads to a disconnect
This happened with the recent Foundry reorg. Neutrino does not check that the peer has the header / block and the bitcoind peer will disconnect if it doesn't have it. Then it will reconnect and get into a loop. From BIP157: 
- "StopHash MUST be known to belong to a block accepted by the receiving peer. This is the case if the peer had previously sent a headers or inv message with that block or any descendents. A node that receives getcfilters with an unknown StopHash SHOULD NOT respond."

Neutrino code: https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1841-L1842
Bitcoind code: https://github.com/bitcoin/bitcoin/blob/99f99c989e737cae003cbf848196a70ab472f8bc/src/net_processing.cpp#L3278-L3289
Pointed out on IRC: https://bitcoin-irc.chaincode.com/bitcoin-core-dev/2026-03-25#1206087

Roasbeef:
In this PR, we fix a bug where neutrino could enter a tight
disconnect/reconnect loop with bitcoind peers during a chain reorg
(#338). The root cause is a race between `blockHandler` (which
processes reorg headers via `handleHeadersMsg` → `rollBackToHeight`)
and `cfHandler` which reads from the header store concurrently to
construct filter queries. When `cfHandler` captures a block hash for
use as a `StopHash` in `GetCFilters`/`GetCFHeaders`/`GetCFCheckpt`
messages, that hash can become stale if `blockHandler` rolls back
the chain in between.

Per BIP 157, the `StopHash` MUST be known to the receiving peer.
bitcoind's `PrepareBlockFilterRequest` enforces this strictly: if the
stop hash isn't in the active chain (i.e.
`!stop_index || !BlockRequestAllowed`), it sets `fDisconnect = true`.
So sending a stale hash from a reorged-out block causes an immediate
disconnect, and since neutrino retries with the same stale state, we
get a tight reconnect loop. This was observed in the wild during the
recent Foundry reorg.

We introduce `errReorgDetected` as a sentinel error and add validation
to all four query construction paths. `getCFHeadersForAllPeers`
snapshots the chain tip hash before the query and re-checks after; if
the tip changed mid-query, we return `errReorgDetected` so the caller
retries with fresh state. `fetchFilterFromAllPeers` and `getCheckpts`
both validate their target block hash still exists in the header store
via `FetchHeader` before sending any query to peers.
`getCheckpointedCFHeaders` replaces a `panic` on
`FetchHeaderByHeight` failure with a graceful `errReorgDetected`
return, since a rolled-back chain makes previously valid heights
unreachable. We also change the signature from `void` to `error` so
callers can react.

All callers in `cfHandler`, `getUncheckpointedCFHeaders`,
`resolveConflict`, and `detectBadPeers` are updated to propagate the
new error. The existing retry loops in `cfHandler` (with
`retryTimeout` backoff and `goto waitForHeaders`) handle the recovery
naturally. When `getCheckpts` returns `errReorgDetected` specifically,
we go to `waitForHeaders` to re-snapshot chain state rather than
retrying with the same potentially stale `bestHash`. Similarly,
`getCheckpointedCFHeaders` failures clear `allCFCheckpoints` and
apply a retry backoff before going back to `waitForHeaders` to prevent
tight loops on persistent store errors.

As a secondary fix, we also replace the remaining `panic` in the
`writeCFHeadersMsg` error path with a proper error return, so store
write failures (DB errors, alignment mismatches from concurrent
reorgs) are recoverable rather than crashing the process.

See each commit message for a detailed description w.r.t the
incremental changes.

Fixes #338.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -5,6 +5,7 @@ package neutrino
 import (
 	"bytes"
 	"container/list"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -47,6 +48,11 @@ const (
 
 // zeroHash is the zero value hash (all zeros).  It is defined as a convenience.
 var zeroHash chainhash.Hash
+
+// errReorgDetected is returned when a chain reorg is detected during a
+// filter-related query, indicating the caller should retry with updated chain
+// state.
+var errReorgDetected = errors.New("reorg detected during filter query")
 
 // newPeerMsg signifies a newly connected peer to the block handler.
 type newPeerMsg struct {
@@ -569,7 +575,29 @@ waitForHeaders:
 
 			log.Debugf("Getting filter checkpoints up to "+
 				"height=%v, hash=%v", bestHeight, bestHash)
-			allCFCheckpoints = b.getCheckpts(&bestHash, fType)
+
+			var checkptsErr error
+			allCFCheckpoints, checkptsErr = b.getCheckpts(
+				&bestHash, fType,
+			)
+			if checkptsErr != nil {
+				log.Warnf("Error fetching checkpoints: "+
+					"%v, trying again...", checkptsErr)
+
+				// If a reorg was detected, we need to
+				// re-snapshot the chain state since our
+				// bestHash may reference a rolled-back block.
+				if errors.Is(checkptsErr, errReorgDetected) {
+					goto waitForHeaders
+				}
+
+				select {
+				case <-time.After(retryTimeout):
+				case <-b.quit:
+					return
+				}
+				continue
+			}
 			if len(allCFCheckpoints) == 0 {
 				log.Warnf("Unable to fetch set of " +
 					"candidate checkpoints, trying again...")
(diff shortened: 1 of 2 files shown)


saubyk:
Hi @Crypt-iQ would you consider reviewing this pr?

Crypt-iQ:
> Hi @Crypt-iQ would you consider reviewing this pr?

I've lost a lot of context here, so probably better to have a different reviewer. I think? the fix here needs to be updated to instead track what headers the peer has sent us.

Roasbeef:
Considering closing this based on this comment chain: https://github.com/lightninglabs/neutrino/issues/338#issuecomment-4145330952

PR #344, opened 2026-04-09
chainimport: fix median time validation in lightHeaderCtx

m0wer:
- `lightHeaderCtx.Parent()` returned nil, causing `CalcPastMedianTime` to use only one timestamp instead of the median of the last 11 blocks. This made chainimport reject valid blocks with non-monotonic timestamps (common on mainnet/signet at heights ~2299 and ~2238 respectively).
- Fix `Parent()` to delegate to `RelativeAncestorCtx(1)` and propagate the `validator` field so the full ancestor chain can be walked.
- Add tests for parent chain traversal and non-monotonic timestamp validation.

### DIFF
--- a/chainimport/block_headers_validator.go
+++ b/chainimport/block_headers_validator.go
@@ -239,6 +239,7 @@ func (l *lightHeaderCtx) RelativeAncestorCtx(
 			height:    int32(ancestorHeight),
 			bits:      ancestor.Bits,
 			timestamp: ancestor.Timestamp.Unix(),
+			validator: l.validator,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


m0wer:
Applied Gemini suggestion and linter.

m0wer:
Ready for review @Roasbeef 

Currently used in https://github.com/m0wer/neutrino-api and https://github.com/joinmarket-ng/joinmarket-ng/ Working fine.

litbot-9000:
@m0wer, remember to re-request review from reviewers when ready

Roasbeef (approved):
LGTM 🐵


Added on another bug fix

PR #351, opened 2026-05-06
feeest: statistical fee estimator from observed block fee rates

Roasbeef:
In this PR, we add a statistical fee estimator to neutrino. Today the chain
service has no fee estimation of its own, so lnd-on-neutrino and other
downstream consumers are stuck with "fee estimation not supported" or have
to wire in an external API. The estimator added here is fully self-contained:
it samples fee data from every block we already fetch (via filter-match
rescans, the utxoscanner, or cfheader-mismatch resolution), combines those
samples with the BIP133 feefilter rates we receive from connected peers, and
produces a fee-rate recommendation parameterised by confirmation target.

This is intentionally a coarse estimator and won't beat `estimatesmartfee`
running against a full node with mempool visibility. The aim is to give
neutrino-backed consumers a real number with a calibrated confidence score
instead of a sentinel error, and to do so at zero extra bandwidth: the
sampler piggybacks on blocks we'd already be downloading for other reasons.

This PR is stacked on top of the SQL storage series and is based on
#367 (`roasbeef/sql-08-header-store-perf`), so only the last five commits
are new here. See each commit message for a detailed description w.r.t.
the incremental changes.

## Background

The applicable prior art is narrower than it first appears. Bitcoin Core's
`BlockPolicyEstimator` is structurally inapplicable here, since it depends
on per-tx mempool-entry timestamps (the "still in pool" denominator) that a
filter-based light client cannot observe. Mempool-projection estimators like
mempool.space's are similarly out, since they need the whole mempool sorted
by ancestor feerate to walk a block template. What's left is the
quantile-on-block-feerate family with exponentially-decayed weighting,
closer in spirit to the SJSU mempool-state thesis and the SARIMAX
comparative paper (arXiv 2502.01029).

The fundamental constraint shaping the design is that we don't have prevout
information for arbitrary transactions. Coinbase output minus subsidy gives
us total block fees; summing transaction weights gives us total block weight;
their ratio gives us a block-average fee rate. One refinement is available to
a light client for free: for transactions whose prevouts are all created
within the same block (intra-block spend chains, typically CPFP children),
the exact fee rate is computable with no UTXO set and no extra bandwidth.
Every included transaction paid at least the block's marginal entry rate, so
the minimum known per-tx rate is a strictly tighter upper bound on the entry
price than the block average. Full per-tx coverage still requires a UTXO
cache and is deferred.

## Design

The data flow has three tiers. Tier C is the network relay floor derived
from connected peers' BIP133 feefilter messages, exposed via the new
`PeerFeeFilters` method on `ChainService`. Tier A is the rolling block-level
sample window persisted in the new `feedb` package and held hot in an
in-memory, height-ordered window. Tier B (per-tx rates from full prevout
knowledge) is reserved in the schema; the intra-block known-tx bound above
partially delivers it ahead of a UTXO cache.

The sample store rides on the SQL backend introduced by the base series: a
`fee_samples` table (schema migration v4), a sqlc query file, and a narrow
`FeeQueries`/`FeeTx` executor wired through `sqldb.Backend`, following the
same shape as the filterdb/banman/headerfs SQL stores. Height is the primary
key with upsert-replace semantics, so a competing block observed at the same
height during a reorg simply overwrites the orphaned row. Since the store is
SQL-only, the estimator activates only when the chain service runs with
`SQLConfig`; on the legacy walletdb configuration `EstimateFeeRate` reports
itself disabled.

The tiers are layered, not merged: C is always the hard floor; A activates
once we have at least 6 recent samples and the most recent isn't older than
the stale window; otherwise we fall back to the cold-start path (the median
peer feefilter scaled by the per-target cushion, fixed 0.10 confidence).
Merging C into A would pollute a fee-market signal with mempool-policy
noise, since the two measure incommensurable things. The floor uses the
median of peer feefilters: since it acts as a hard clamp, a higher
percentile would let a minority of adversarial peers force every estimate
upward, whereas moving the median requires compromising more than half our
peer connections.

The in-memory window is height-ordered rather than FIFO: it evicts the
lowest height when full and rejects samples older than everything it holds,
so a rescan pulling a few hundred historical blocks can't flush the recent
samples and knock a warm estimator back to cold-start. This matters because
rescans are one of the main ways neutrino fetches full blocks at all.

For a tier-A answer, the window is weighted by an exponential-decay EWMA
(default 30-minute half-life, adapting to regime shifts at most once per new
block sample). We compute weighted empirical quantiles of per-block
entry-rate proxies, where each block's proxy is the tighter of its average
fee rate and its minimum known per-tx rate. Block fee rates are heavily
right-skewed, so quantiles (bounded by observed samples, no distributional
assumption) behave far better than a mean + k-sigma projection during fee
storms. The per-target quantiles follow `p = 1 - (1-0.85)^(1/n)`, the
probability of beating at least one of the next n blocks. A congestion gate
then blends the projection toward the relay floor by the EWMA-weighted
fraction of full blocks: when blocks have spare capacity there's no
competition for space and the marginal entry price collapses to the floor,
whatever historical averages say. Empty blocks count as evidence of an idle
market rather than being discarded. The result is clamped to never sit
below the relay floor.

The confidence score in `[0, 1]` is a composite of sample density, recency,
intra-window stability (1 / coefficient-of-variation), and an agreement term
that only penalises block samples sitting below the relay floor. Callers can
treat values below ~0.4 as a signal to consult an external estimator.
`Estimate` and `CurrentStats` expose the congestion fraction alongside the
rate so callers can see which regime produced the answer.

Reorgs are handled via a new `OnRollback` hook on the block manager: when
the header chain is rolled back, the chain service prunes fee observations
from orphaned blocks out of both the in-memory window and the on-disk store.

One deliberate tradeoff to flag for review: with the congestion gate,
low-congestion estimates land near the relay floor. That's correct (there is
no competition for block space), but less conservative than a pure
projection. The quantile and target-multiplier tables are package-level
tunables pending calibration against real sample distributions.

## lnd integration

The new `neutrino.ChainFeeEstimator` adapter exposes the estimator through
the method set of lnd's `lnwallet/chainfee.Estimator` interface. Since lnd
already depends on neutrino, importing chainfee here would be circular, so
`feeest` declares a structurally identical `ChainEstimator` contract with
`feeest.SatPerKW` (int64 sat/kW, byte-compatible with
`chainfee.SatPerKWeight`) standing in for lnd's unit; the lnd side satisfies
`chainfee.Estimator` with a one-cast-per-method wrapper. `EstimateFeePerKW`
and `RelayFeePerKW` floor their results at `feeest.FeePerKWFloor` (253
sat/kW, mirroring `chainfee.FeePerKwFloor`) so a cold estimator can never
hand the wallet a rate that would fail to relay.

## What's not here

Full per-tx fee rate refinement (tier B beyond intra-block chains), adaptive
sampling stride, and force-fetch-on-broadcast are deferred. The schema,
window, and estimator are designed to absorb each of these without breaking
changes:

- Full tier B slots into `Sampler.Observe` once a UTXO cache is wired in;
  the feedb encoding already carries per-sample known-tx fields.
- Adaptive stride is a `Config.FeeSampleStride` field plus a periodic
  fetcher that calls `GetBlock` on tip blocks.


Roasbeef:
I took a close look at the estimator with an eye on robustness, accuracy, and sample efficiency, and pushed a set of follow-up commits on top of this branch: [`feeest-estimator-improvements`](https://github.com/lightninglabs/neutrino/tree/feeest-estimator-improvements) ([diff vs this PR](https://github.com/lightninglabs/neutrino/compare/feeest%2Ffee-estimator...feeest-estimator-improvements)).

## Analysis of the current approach

The overall architecture (opportunistic block sampling at the `GetBlock` chokepoint, feedb persistence + hot in-memory window, layered tiers with feefilter as an incommensurable floor rather than a merged signal) is sound. The issues I found are all in the statistical core and in a few operational edge cases:

1. **Rescan pollution (robustness).** The ring evicts by *insertion* order and staleness is judged by the *last-inserted* sample. A rescan or utxoscanner pass that fetches ~`RingSize` historical blocks flushes every recent sample and silently knocks a warm estimator back to cold-start — and since rescans are one of the main ways neutrino fetches blocks at all, this hits exactly the users the estimator is for.

2. **Gaussian mean + kσ on a heavy-tailed distribution (accuracy).** Block-average fee rates are strongly right-skewed. During a fee storm the window variance explodes, so the target-1 answer (`1.5 × (mean + 1.65σ)`) is unbounded above and overshoots badly; negative kσ for target 24 can also project below every observed sample.

3. **Block fullness is ignored (accuracy).** When recent blocks aren't full there is no competition for block space and the marginal entry price collapses to the relay floor, regardless of what the block *averages* say (a few large-fee txs in a half-empty block produce a high average). The current math still projects mean + kσ, which systematically overpays in the common uncongested regime. Relatedly, empty blocks are dropped entirely — but an empty block is strong *evidence* of an idle market; discarding it biases the estimate upward.

4. **p75 feefilter floor is a griefing vector (robustness).** The floor is used as a hard clamp, so 25% of adversarial peers advertising inflated feefilters can force every estimate upward. 

5. **One scalar per block (sample efficiency).** The block average is a loose upper bound on the marginal entry rate. There is more free signal in a block: coinbase weight (systematic bias in the average's denominator), fullness, and — notably — **exact per-tx fee rates for transactions whose prevouts are created within the same block** (CPFP chains). Those are computable by a light client with no UTXO set and no extra bandwidth, and every included tx paid ≥ the block's marginal rate, so the minimum known per-tx rate is a strictly tighter upper bound on the entry price than the block average.

6. Smaller items: the adaptive half-life re-runs on every `Estimate` call, so repeated queries over an unchanged window ratchet it to the clamp; the confidence "agreement" term penalizes every normal fee market (samples above floor) instead of only the anomalous direction; the cold-start path ignores the confirmation target; and the reorg hook (`PurgeFrom` + `ring.prune`) exists but is never called, as noted in the description.

## What the branch changes

- **Height-ordered sample window** replacing the FIFO ring: evicts lowest height, rejects samples older than its horizon, so historical rescans can't displace fresh tip samples (regression test included). Fixes (1).
- **Weighted empirical quantiles** of per-block entry-rate proxies replace mean + kσ. Per-target quantiles follow `p = 1 − (1−0.85)^(1/n)` (probability of beating at least one of the next n blocks): 0.85 / 0.47 / 0.27 / 0.08, still exposed as tunables like the kσ tables were. Bounded by observed samples, no distributional assumption. Fixes (2).
- **Congestion gate:** blend the projection toward the relay floor by the EWMA-weighted fraction of full blocks (≥95% of the weight limit), computed over the *unfiltered* window so empty blocks count as idle-market evidence. `Estimate`/`Stats` expose the new `Congestion` field. Fixes (3).
- **Median feefilter floor** instead of p75 (matches the existing `RelayFeePerKW` reasoning). Fixes (4).
- **Richer samples via a backward-compatible feedb v2 encoding** (v1 rows decode with new fields zeroed, no migration): `CoinbaseWeight` (average now computed over fee-paying weight), `MinKnownTxRate`/`KnownTxCount` from intra-block spend chains. The estimator's per-block proxy takes the tighter of block average and known-tx bound, which also deflates spam-dominated blocks — and `FlagSpam` is now actually set when a single computable tx dominates block fees. Partially delivers tier B ahead of a UTXO cache. Fixes (5).
- **Operational fixes:** half-life adapts at most once per new block sample; agreement term only penalizes samples below the floor; cold start scales by the per-target cushion; and a new `OnRollback` hook on `blockManagerCfg` fires from `rollBackToHeight` into `Sampler.PruneFrom`, closing the reorg gap.

One deliberate tradeoff to flag for review: with the congestion gate, low-congestion estimates land near the relay floor — correct, but less conservative than before. The `TargetMult` cushions (now 1.2/1.1/1.0/0.9, since the entry proxy is itself an upper bound) and quantiles remain package-level tunables for calibration against real data.

`go test -race ./feeest/... ./feedb/...` passes, `go vet`/`gofmt` clean. Happy to split any of this out or fold it into the PR's commit structure.

---
_Generated by [Claude Code](https://claude.ai/code)_

Roasbeef:
Correction to the compare link above — the follow-up commits relative to this PR's head are: [`wip/feeest...feeest-estimator-improvements`](https://github.com/lightninglabs/neutrino/compare/wip/feeest...feeest-estimator-improvements).

Roasbeef:
Folded the `feeest-estimator-improvements` follow-ups from the comment above into the PR's commit structure and rebased the whole series onto the SQL storage stack (#367), so the branch now sits alongside the sync series in the stack. The history is now five commits, each building and passing tests standalone; the PR description has been updated to match the reworked statistical core (height-ordered window, weighted empirical quantiles, congestion gate, median feefilter floor, intra-block known-tx bound, reorg pruning).

Roasbeef:
Ran a deep multi-agent review pass over the branch (8 finder angles, ~40 raw candidates, each surviving candidate independently verified against the code) and folded the confirmed fixes back into the commit structure:

- **Reorg race:** GetBlock now re-checks the header at the resolved height before feeding the sampler, and the window replaces same-height/different-hash entries, so an in-flight fetch racing `PruneFrom` can no longer double-weight a height with orphaned-chain data.
- **32-bit safety:** the estimator's half-life moved to `atomic.Int64`, fixing a guaranteed unaligned-atomic panic on arm/386 builds.
- **Cold start:** estimates are now floored at `FeePerKWFloor` (253 sat/kW) in the cold path, so an estimator with no peer feefilter data can never return a zero rate with a nil error.
- **Lifecycle:** the sampler's GC worker now spawns from `ChainService.Start` (no goroutine leak on constructor error paths), and `Stop` is idempotent and waits for the worker before the SQL backend closes.
- **Half-life adaptation:** the once-per-block guard uses compare-and-swap, and the recent/prior split is by sample count instead of wall time, un-freezing the recovery branch once the half-life bottomed out (regression test added).
- **Rescan churn:** `Observe` early-outs for blocks below the retention horizon, so deep rescans no longer write thousands of rows the next GC pass deletes.
- Smaller items: `CurrentStats` is exposed via `ChainService.FeeEstimatorStats`, the redundant `peerFeeRater` adapter is gone, and the `FlagSpam` docs now describe what the estimator actually does with spam-dominated blocks.

Every commit in the series still builds and passes its package tests standalone; `go test -race ./feedb/... ./feeest/... ./sqldb/...` and `go vet ./...` are clean at the tip.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #352, opened 2026-05-07
sqldb: scaffold sqlite/postgres backend + bump Go to 1.25/1.26.2

Roasbeef:
In this PR, we kick off a series of stacked changes that bring a SQL
(sqlite + postgres) storage backend to neutrino, alongside the existing
walletdb (bbolt) + flat-file path. lnd, taproot-assets, and darepo have
all converged on a SQL-first persistence layer over the past year, and
this stack brings neutrino in line so that downstream consumers (lnd,
lit, btcwallet) can eventually run the entire stack on a single SQL
database instead of provisioning a parallel bbolt file just for chain
state.

This first PR lands the scaffolding: the new `sqldb/` package, an
embedded migration set with the schema for every store that follows in
the stack, sqlc-generated query code, and a smoke test. No existing
neutrino code path is wired up to it yet, and `Config.Database` remains
the only persistent backend recognised by `NewChainService`. The
follow-up PRs add per-domain SQL implementations one at a time, then a
legacy walletdb importer, and finally the `Config.SQLConfig` plumbing
that makes the new backend reachable from the public API.

## Stack

This is PR 1 of 6 in the SQL backend stack:

1. **(this PR)** `sqldb` scaffold + Go bump
2. `banman`: SQL backend
3. `filterdb`: SQL backend
4. `headerfs`: SQL backend
5. `sqldb`: legacy walletdb + flat-file import migration
6. `neutrino`: wire `SQLConfig` into `ChainService` + postgres CI matrix

Each subsequent PR is based on the previous PR's branch. See each commit
message for a detailed description w.r.t the incremental changes.

## Layout

The new `sqldb/` package wraps the standalone
`github.com/lightningnetwork/lnd/sqldb/v2` module, which is its own Go
module under lnd's tree (separate `go.mod`), so importing it doesn't
introduce a cycle even though lnd itself depends on neutrino. We get the
shared `BaseDB`, `TransactionExecutor[Q]`, retry-on-serialization-error
logic, sqlite + postgres drivers, and `replacerFS` token rewrites for
free. The lightninglabs/migrate fork is wired in via a `replace`
directive (the same one lnd's `sqldb/v2` ships with) because the v2
module references `migrate.ProgrammaticMigrEntry`, which only exists in
the fork.

```
sqldb/
├── backend.go         // NewBackend factory + per-domain TransactionExecutors
├── config.go          // Config, BackendType (sqlite | postgres)
├── queries.go         // Narrow HeaderQueries / FilterQueries / BanQueries interfaces
├── migrations.go      // MigrationSet wired with embedded SQL files
├── log.go             // btclog logger
├── test_helpers.go    // NewTestBackend for in-process sqlite tests
├── backend_test.go
└── sqlc/
    ├── sqlc.yaml
    ├── migrations/
    │   ├── 000001_init.up.sql
    │   └── 000001_init.down.sql
    ├── queries/{ban,chain_tips,filter_headers,filters,headers}.sql
    └── (generated) models.go, querier.go, db.go, *.sql.go
```

## Schema

The `000001_init` migration creates the five tables that back every
neutrino store: `chain_tips` (one row per chain type, block + filter),
`block_headers` (height PK + UNIQUE block_hash, raw 80-byte header
BLOB), `filter_headers` (height PK + UNIQUE block_hash, 32-byte
filter_hash), `regular_filters` (block_hash PK + filter_bytes BLOB), and
`banned_peers` (id PK + UNIQUE ip_net, expiration, reason). The schema
is sqlite-first; postgres compatibility is handled by the canonical
token replacements baked into sqldb/v2 (`BLOB` → `BYTEA`, `INTEGER
PRIMARY KEY` → `BIGSERIAL PRIMARY KEY`, `TIMESTAMP` → `TIMESTAMP WITHOUT
TIME ZONE`).

There are no foreign keys between `block_headers` and `filter_headers`
because the two chains drift independently during sync. The 2-byte
hash-prefix sub-bucket scheme used in `headerfs/index.go` is dropped on
the SQL side because it's a bbolt b+tree memory-pressure optimisation
that sqlite + pg b-trees don't need; a plain UNIQUE index on
`block_hash` outperforms the bucketed scheme.

## House style

Each domain in the follow-up PRs declares a narrow query interface
(e.g., `HeaderQueries`) that is a strict subset of the generated
`sqlc.Querier`. The store struct is parameterised by
`sqldbv2.BatchedTx[Q]`, so transactional work goes through a single
`ExecTx(ctx, opts, body, reset)` call that drops the work into a
write/read tx and retries on serialization-or-deadlock errors. This
matches the pattern in `tapdb`, `darepo/db`, and lnd's `payments` /
`invoices` migration verbatim.

```go
type HeaderTx = sqldbv2.BatchedTx[HeaderQueries]
type FilterTx = sqldbv2.BatchedTx[FilterQueries]
type BanTx    = sqldbv2.BatchedTx[BanQueries]
```

The `Backend` factory opens the underlying sqlite or postgres store,
applies migrations, and constructs one `TransactionExecutor[Q]` per
domain that the per-domain stores in PRs 2-4 will inject as their sole
dependency.

## Go bump

The top-level `go` directive moves to 1.25 because sqldb/v2 pulls in
indirect deps (modernc/sqlite, jackc/pgx, dockertest, etc) that require
newer language features. The `tools/` and `cache/` sub-modules are
brought up to 1.25 as well so the whole tree is on a single line. CI
runner is bumped to 1.26.2 to cover the toolchain features the new
deps use during build.

## Testing

A smoke test in `backend_test.go` opens an in-process sqlite Backend in
a temp directory, verifies migrations apply cleanly, and runs a trivial
query through each per-domain `TransactionExecutor` so the whole wiring
chain (sqlc → narrow interface → BatchedTx → sql.Tx) is exercised.
Per-domain tests follow in PRs 2-4. The postgres CI matrix entry lands
in PR 6 and exercises every test against both backends on every PR.


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #353, opened 2026-05-07
banman: add SQL backend implementation

Roasbeef:
In this PR, we add `SQLBanStore`, the SQL-backed implementation of the
existing `banman.Store` interface. It's the smallest of the four
domains in the SQL stack (just three methods) so it makes a useful
sanity check on the house style before we tackle the more involved
header and filter stores.

The store is wired via the `BanTx` executor that landed in the previous
PR. Every method dispatches a single `ExecTx` so the entire body of the
operation runs as one serialisable transaction with automatic retry on
serialization-or-deadlock errors. The on-the-wire IPNet key format is
preserved verbatim from `banman/codec.go::encodeIPNet`, so when the
legacy importer in PR 5 walks the bbolt `ban-index` bucket it can copy
rows straight across without re-encoding anything.

## Stack

This is PR 2 of 6 in the SQL backend stack. Stacks on top of #352.

## Sweep on read

The legacy `banStore.Status` deletes any expired record in the same
write tx that read it, so a caller never sees a stale `Banned: true`
return for an IP whose ban has elapsed. We carry that semantic over
exactly: `SQLBanStore.Status` runs as a write tx, fetches the row, and
if `time.Now()` is past the stored expiration, deletes the row before
returning `Status{Banned: false}`. Doing both reads and the delete in
one tx avoids the read-then-delete race that would show up if we split
them.

A small `sync.Mutex` guards the public methods. The SQL engine's
isolation is enough for correctness on its own, but the legacy bbolt
store is observably linearizable from a Go caller's perspective: a
`Status` immediately following a successful `BanIPNet` always sees the
new record. Keeping the mutex preserves that contract for callers like
the connection manager that rely on it.

## Tests

The existing `TestBanStore` body is refactored into a parameterised
helper that runs against both backends via a small factory map (`kvdb`
+ `sqlite`). `TestBanStoreBackends` covers ban, expiration sweep,
explicit unban, and the post-sweep idempotent return. SQL-specific
tests verify upsert idempotency on repeat ban (with a longer duration
correctly extending the existing record), IPv6 round-tripping through
`encodeIPNet`, and the bulk `PurgeExpired` helper that callers can call
opportunistically at startup to bound the size of the `banned_peers`
table.

```go
type SQLBanStore struct {
    db  sqldb.BanTx
    mtx sync.Mutex
}

var _ Store = (*SQLBanStore)(nil)
```

### DIFF
--- a/banman/sql_store.go
+++ b/banman/sql_store.go
@@ -1,0 +1,194 @@
+package banman
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"net"
+	"sync"
+	"time"
+
+	sqldbv2 "github.com/lightningnetwork/lnd/sqldb/v2"
+
+	"github.com/lightninglabs/neutrino/sqldb"
+	"github.com/lightninglabs/neutrino/sqldb/sqlc"
+)
+
+// SQLBanStore is a SQL-backed implementation of the Store interface. It
+// stores peer bans in the banned_peers table provisioned by the neutrino SQL
+// migrations.
+type SQLBanStore struct {
+	db sqldb.BanTx
+
+	// mtx serialises access from concurrent callers so that observable
+	// semantics match the legacy walletdb-backed banStore: a Status call
+	// immediately following a BanIPNet always sees the new record.
+	mtx sync.Mutex
+}
+
+// A compile-time constraint to ensure SQLBanStore satisfies the Store
+// interface.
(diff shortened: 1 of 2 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #354, opened 2026-05-07
filterdb: add SQL backend implementation

Roasbeef:
In this PR, we add `SQLFilterStore`, the SQL-backed implementation of
`filterdb.FilterDatabase`. It mirrors the legacy bbolt store's contract
exactly, including the three subtle distinctions between an absent
filter, a present-but-nil-sentinel filter, and the
delete-and-recreate-without-reseeding behaviour of `PurgeFilters`.

## Stack

This is PR 3 of 6 in the SQL backend stack. Stacks on top of #353.

## The three states

Three different conditions can be observed when looking up a filter,
and each has a distinct return:

- **Row absent.** `GetFilter` returns `sql.ErrNoRows`, which the store
  maps to `ErrFilterNotFound`. This is the same behaviour as the bbolt
  store returning `nil` from `bucket.Get` for a missing key.
- **Row present, value zero-length.** This is the legacy "deleted
  filter" sentinel: the bbolt store writes a zero-length value via
  `bucket.Put(hash, nil)`, and the SQL store coerces that to an empty
  byte slice (so the `NOT NULL filter_bytes` column accepts the row).
  `FetchFilter` returns `(nil, nil)` here, matching the existing
  contract that callers like the batch writer rely on.
- **Row present, value populated.** Standard path: deserialise via
  `gcs.FromNBytes` and return the filter.

The `PurgeFilters` method drops every row of the chosen type without
re-seeding genesis, which exactly matches the bbolt
delete-and-recreate-bucket pattern that the existing test suite
encodes. The current production code only ever uses `RegularFilter`, so
there's no `filter_type` column in the schema; if a second type ever
shows up we add it via a follow-up migration that introduces the
column.

## Genesis

Genesis filter creation runs once at construction inside a single write
tx. The check is `GetFilter(genesisHash) → sql.ErrNoRows` followed by a
`PutFilter` in the same tx, so two processes racing on first start can
never both win and end up with two genesis rows.

## Tests

The existing `db_test.go` cases are refactored into a parameterised
suite (`kvdb` + `sqlite`) that exercises genesis creation and
store-and-fetch round-tripping against both backends. SQL-specific
tests cover the nil-sentinel round-trip, the `ErrFilterNotFound`
absent-row mapping, the purge contract (genesis included), the
`unknown filter type` error path on both `FetchFilter` and
`PurgeFilters`, and the constructor's idempotency on repeat
construction against the same DB.

### DIFF
--- a/filterdb/sql_store.go
+++ b/filterdb/sql_store.go
@@ -1,0 +1,227 @@
+package filterdb
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcutil/gcs"
+	"github.com/btcsuite/btcd/btcutil/gcs/builder"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	sqldbv2 "github.com/lightningnetwork/lnd/sqldb/v2"
+
+	"github.com/lightninglabs/neutrino/sqldb"
+	"github.com/lightninglabs/neutrino/sqldb/sqlc"
+)
+
+// SQLFilterStore is a SQL-backed implementation of the FilterDatabase
+// interface.
+//
+// The store is keyed exclusively by block hash because the only filter type
+// neutrino currently uses is RegularFilter; the schema reflects this and the
+// generated queries operate against the regular_filters table directly.
+type SQLFilterStore struct {
+	db          sqldb.FilterTx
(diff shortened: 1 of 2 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #355, opened 2026-05-07
headerfs: add SQL backend implementation

Roasbeef:
In this PR, we add `SQLBlockHeaderStore` and `SQLFilterHeaderStore`,
the SQL-backed implementations of the existing `BlockHeaderStore` and
`FilterHeaderStore` interfaces. This is the largest correctness-critical
piece of the stack: the existing bbolt + flat-file implementation has a
documented inconsistency window between the on-disk binary and the
walletdb index, and pulling everything into a single SQL transaction
removes the entire class of failures that the existing
reconciliation-on-startup logic exists to repair.

## Stack

This is PR 4 of 6 in the SQL backend stack. Stacks on top of #354.

## Atomicity wins

The legacy `WriteHeaders` writes the raw bytes to `block_headers.bin`
first, then updates the bbolt header-index, with each step in its own
tx boundary (file `Write` + `walletdb.Update`). If the database update
fails mid-batch, the existing code does its best to truncate the flat
file back to the previous valid offset, but there are still corner
cases where the file and index can drift out of sync. The constructor
at `headerfs/store.go:294-327` exists specifically to detect that drift
on next startup and truncate the file to match the index.

In the SQL implementation, the entire batch (every `InsertBlockHeader`
plus the closing `UpsertChainTip`) runs in one `ExecTx` write tx. If
any insert fails (UNIQUE violation, retryable error, anything else),
the whole tx rolls back atomically and there is no drift to detect on
next startup. We delete the reconciliation logic from the SQL path
entirely; it's only relevant on the bbolt side.

`RollbackBlockHeaders` and `RollbackLastBlock` get the same treatment:
read the current tip, look up the new tip header, `DELETE FROM
block_headers WHERE height > N`, and `UpsertChainTip` to the new value,
all in one tx. Strictly better than the legacy two-step truncate +
walletdb update.

## LatestBlockLocator in one tx

The bbolt `blockLocatorFromHash` walks back from the chain tip with an
exponentially increasing decrement, hitting `FetchHeaderByHeight` for
each step and incurring a fresh read tx per call. There's been a TODO
on that loop for years to put the whole walk inside a single
transaction. The SQL implementation does just that: one `ExecTx` read
tx scopes the entire walk, every `GetBlockHeaderByHeight` runs in the
same snapshot, and the locator is returned at the end.

## maybeReset rewritten as one tx

The bbolt filter header store has an `AssertFilterHeader` mechanism
where, on first start, the caller can pass an expected
`(height, filter_hash)` tuple. If the on-disk row at `height` doesn't
match, the store closes the file, calls `os.Remove`, and recursively
re-invokes `NewFilterHeaderStore` to re-init from genesis. That flow is
brittle: two separate filesystem operations between the assertion check
and the re-init, no atomicity, Windows-specific handling for the file
close before remove.

In the SQL version, the whole reset is one write tx:
`PurgeFilterHeaders` to drop every row, `DeleteChainTip(filter)` to
reset the tip pointer, then `InsertFilterHeader` + `UpsertChainTip`
with the recomputed genesis. No recursion, no os.Remove, no Windows
edge case.

## Linearizability

Both stores keep a `sync.RWMutex`. The SQL engine's isolation alone is
enough for correctness, but callers like `blockmanager` rely on the
observable property that a `ChainTip` immediately following a
successful `WriteHeaders` returns the new tip without a write-then-read
gap. The mutex preserves that contract.

## Tests

`TestSQLBlockHeaderStoreOperations` exercises a 100-header
write-and-fetch chain plus a single rollback, mirroring the existing
bbolt test. `TestSQLBlockHeaderRollbackAtomic` injects a UNIQUE
violation mid-batch and asserts that no rows from the failed batch
persist (full tx rollback proves itself).
`TestSQLBlockHeaderGenesisIdempotent` constructs the store twice
against the same DB and asserts `block_headers` has exactly one row at
height 0. `TestSQLFilterHeaderAssertReset` covers the matching,
mismatched, and missing-height cases of the assertion path.

### DIFF
--- a/headerfs/sql_store.go
+++ b/headerfs/sql_store.go
@@ -1,0 +1,961 @@
+package headerfs
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil/gcs/builder"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	sqldbv2 "github.com/lightningnetwork/lnd/sqldb/v2"
+
+	"github.com/lightninglabs/neutrino/sqldb"
+	"github.com/lightninglabs/neutrino/sqldb/sqlc"
+)
+
+// SQLBlockHeaderStore is a SQL-backed implementation of BlockHeaderStore.
+type SQLBlockHeaderStore struct {
+	db          sqldb.HeaderTx
+	chainParams *chaincfg.Params
+
+	// mtx preserves the linearizable semantics that callers like
+	// blockmanager rely on: a ChainTip immediately following a
+	// successful WriteHeaders must observe the new tip.
+	mtx sync.RWMutex
+}
+
+// A compile-time check to ensure SQLBlockHeaderStore satisfies the
+// BlockHeaderStore interface.
+var _ BlockHeaderStore = (*SQLBlockHeaderStore)(nil)
+
+// NewSQLBlockHeaderStore returns a new SQL-backed block header store. On a
+// freshly-migrated database it inserts the genesis row before returning.
+func NewSQLBlockHeaderStore(ctx context.Context, db sqldb.HeaderTx,
+	netParams *chaincfg.Params) (*SQLBlockHeaderStore, error) {
+
+	if db == nil {
+		return nil, errors.New("nil header tx executor")
+	}
+
+	store := &SQLBlockHeaderStore{
+		db:          db,
+		chainParams: netParams,
+	}
+
+	if err := store.ensureGenesis(ctx); err != nil {
+		return nil, err
+	}
+
+	return store, nil
+}
+
+// ensureGenesis writes the genesis block header row into block_headers if
+// the table is empty. The check + insert run inside a single tx so multiple
+// processes racing on first start cannot insert duplicate genesis rows.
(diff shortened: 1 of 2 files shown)


sputn1ck:
While exercising the SQL-backed `FilterHeaderStore` path, I hit a compatibility issue that seems specific to the flat-file -> SQL migration in this PR.

`blockmanager.writeCFHeadersMsg` builds `[]headerfs.FilterHeader` batches for `WriteHeaders` where the intermediate entries only carry `FilterHash`; only the final entry has `HeaderHash` and `Height` populated. That works with the legacy flat-file filter header store because it appends every filter header hash to the file and only uses the final entry to update the tip index.

The SQL implementation persists every filter header as a row keyed by `height` and `block_hash`, so the same batch shape attempts to insert multiple rows at zero-value height/hash and fails with a UNIQUE constraint, e.g. duplicate `filter_headers.height`.

I think the fix belongs in `SQLFilterHeaderStore.WriteHeaders`, not in `blockmanager`: the interface/legacy behavior already allowed this batch shape, and the SQL store can preserve drop-in compatibility by deriving missing per-row `(height, block_hash)` metadata from the already-written `block_headers` table inside the same write tx. A targeted regression test would be: write block headers, call SQL filter `WriteHeaders` with only the final filter header carrying height/hash, then assert every row is fetchable by height and by the derived block hash.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #356, opened 2026-05-07
sqldb: add legacy walletdb+flat-file import migration

Roasbeef:
In this PR, we add the auto-importer that copies neutrino's existing
walletdb buckets and append-only header binaries into the new SQL
tables. Without this, every existing neutrino deployment switching to
the SQL backend would have to resync from genesis, which on mainnet is
hours of header download and roughly the same again of filter sync.
The importer makes the transition a single restart with the new
config.

The whole import runs as the version-2 programmatic migration in the
sqldb migration set, which means it's recorded exactly once in the
`neutrino_migrations` tracking table. Crash mid-import and the next
start re-runs the whole thing from scratch against the still-intact
legacy state, bumping the tracker to v2 only once everything is
verified.

## Stack

This is PR 5 of 6 in the SQL backend stack. Stacks on top of #355.

## Opting in

Callers opt in by passing a `LegacyDataSource` into
`MakeLegacyImportMigration`, which is then handed to `NewBackend` (the
public API for that wiring lands in PR 6). The source bundles the
already-open legacy bbolt handle, the data dir holding the two `.bin`
files, and the chain params used for the verification check at the
end:

```go
src := &sqldb.LegacyDataSource{
    DB:      legacyDB,
    DataDir: dataDir,
    Params:  &chaincfg.MainNetParams,
}
backend, err := sqldb.NewBackend(
    dataDir, cfg, sqldb.MakeLegacyImportMigration(src),
)
```

Pass `nil` and the v2 migration is recorded as a no-op contentless
entry (the corresponding `000002_legacy_import.up.sql` is
comment-only, so the source advertises the version but `hasSqlMig` is
false; with no programmatic migration registered, the migrate fork
falls through to a noop and bumps the tracker to v2). Every fresh-
install path is therefore consistent with the post-import state.

## Per-domain import

**Block headers** stream out of `block_headers.bin` in 2000-row
batches. Each batch is one `pread`, one `BeginTx(WriteTxOpt)`, a loop
of `InsertBlockHeader` calls, and a commit. Memory stays bounded
regardless of chain length. The block hashes are computed via
`wire.BlockHeader.BlockHash()` on the freshly-deserialised header (no
need to walk the bbolt header-index for this).

**Filter headers** reuse the `[]chainhash.Hash` produced by the block
header pass. The legacy `reg_filter_headers.bin` only stores the
32-byte filter hash per height, not the corresponding block hash, so
without the precomputed slice we'd have to walk the
`header-index/regular` sub-buckets to reconstruct the `block_hash`
column. Reusing the existing slice avoids a second pass over the bbolt
index and keeps memory bounded to ~28 MB on mainnet.

**Filters** are streamed batch-by-batch from the `filter-store/regular`
bucket. The legacy nil-sentinel (a row with a zero-length value)
matters here: bbolt's `bucket.ForEach` may return `v == nil` for those
rows, but the SQL `filter_bytes` column is `NOT NULL`, so we coerce
nil values to `[]byte{}` so the row inserts cleanly without losing the
sentinel semantics on read.

**Bans** are converted from the uint64 BE expiry encoding to a real
`TIMESTAMP` column. Already-expired records are skipped; the live
store would auto-purge them on first `Status` call anyway, so there's
no point importing them.

## Idempotency on retry

Every run begins by `DELETE FROM` against each of the five tables.
That makes a half-finished import safe to re-run from the start: the
underlying legacy state is read-only and immutable for the duration of
the migration, so the next attempt sees the same input and produces
the same output.

`migrate.ProgrammaticMigrEntry.ResetVersionOnError = true` is set, so
if any step fails the migrate fork rolls the version back to v1 and
the next start re-runs v2 from scratch. Once the migration body
returns nil, the tracker advances to v2 and subsequent restarts skip
the import entirely.

## Verification + .bak rename

Before the migration body returns success, we run a final read tx that
checks row counts and tip-hash equality against the values computed
during the import passes. Any mismatch fails the migration. Only after
verification do we `os.Rename` the legacy `.bin` files to `.bak`. The
walletdb file itself is left in place; operators can delete it
manually after confirming the SQL state. A rename failure here is
non-fatal because the migration tracker already records the import as
complete; the `.bak` files are a convenience, not a correctness
guarantee.

## Tests

`TestLegacyImportRoundTrip` seeds a real bbolt + flat-file legacy
state via the existing constructors (block + filter headers via
`headerfs.NewBlockHeaderStore` / `NewFilterHeaderStore`, GCS filters
via `filterdb.New`, bans via `banman.NewStore`, including a nil
sentinel filter and a deliberately-expired ban), runs the importer,
and asserts that every row matches under both backends.

`TestLegacyImportFreshInstall` opens the SQL backend in an empty
directory and verifies the v2 migration is a no-op (no `.bak` files
get created).

`TestLegacyImportIdempotent` runs the import once, closes the backend,
re-opens with the same legacy source, and verifies that the second
open does not re-run the import (the migration tracker is already at
v2, so the import body is skipped).

### DIFF
--- a/sqldb/migrate_legacy.go
+++ b/sqldb/migrate_legacy.go
@@ -1,0 +1,748 @@
+package sqldb
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/walletdb"
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database"
+	sqldbv2 "github.com/lightningnetwork/lnd/sqldb/v2"
+
+	"github.com/lightninglabs/neutrino/sqldb/sqlc"
+)
+
+// Default file names neutrino has historically used for the append-only
+// header binaries. Callers may override via LegacyDataSource if they have
+// non-default file names.
+const (
+	DefaultBlockHeadersFilename       = "block_headers.bin"
+	DefaultRegFilterHeadersFilename   = "reg_filter_headers.bin"
+)
+
+// Legacy walletdb bucket and key names. They mirror the unexported names in
+// neutrino/headerfs/index.go, neutrino/banman/store.go, and
+// neutrino/filterdb/db.go. Duplicating them here keeps the legacy importer
+// independent of the live store implementations.
+var (
+	legacyHeaderIndexBucket = []byte("header-index")
+	legacyBitcoinTipKey     = []byte("bitcoin")
+	legacyRegFilterTipKey   = []byte("regular")
+
+	legacyFilterStoreBucket = []byte("filter-store")
+	legacyRegFilterBucket   = []byte("regular")
+
+	legacyBanStoreBucket   = []byte("ban-store")
+	legacyBanIndexBucket   = []byte("ban-index")
+	legacyReasonIndexBucket = []byte("reason-index")
+)
+
+// LegacyDataSource declares the legacy on-disk state to be imported into the
+// SQL backend on first start. Constructed by the caller (lnd/lit/btcwallet)
+// when transitioning from the walletdb backend, it bundles the open legacy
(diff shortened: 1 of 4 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #357, opened 2026-05-07
neutrino: wire SQLConfig into ChainService + add postgres CI matrix

Roasbeef:
In this PR, we wire the SQL backend into neutrino's public API and add
the postgres job to the CI matrix. This is the only PR in the stack
that mutates `neutrino.Config`'s public surface, and the only one that
existing library consumers (lnd, lit, btcwallet) will notice. Without
it, every piece of machinery from PRs 1-5 is unreachable from the
caller's perspective; with it, opting into a SQL backend is a single
additive field on `Config`.

Existing walletdb deployments see zero behaviour change. The default
path through `NewChainService` is exactly the same, byte for byte, as
it was before the stack landed; the only way to reach the SQL code is
to set `SQLConfig` (and optionally `LegacyImport` for the importer
that landed in #356).

## Stack

This is PR 6 of 6 in the SQL backend stack, the final PR. Stacks on
top of #356.

## Public API

`Config` grows two additive fields:

```go
type Config struct {
    // ... existing fields ...

    Database walletdb.DB

    SQLConfig *sqldb.Config

    LegacyImport *sqldb.LegacyDataSource

    // ... existing fields ...
}
```

`Database` and `SQLConfig` are mutually exclusive; `NewChainService`
validates this up front before any side effects, so a misconfigured
caller fails fast with a clear error. Exactly one of the two must be
set. `LegacyImport` is only honoured when `SQLConfig` is set; pass nil
if you don't need the auto-importer.

```go
cs, err := neutrino.NewChainService(neutrino.Config{
    DataDir:     dataDir,
    ChainParams: chaincfg.MainNetParams,
    SQLConfig: &sqldb.Config{
        Backend: sqldb.BackendSqlite,
        Sqlite:  &sqldbv2.SqliteConfig{},
    },
    LegacyImport: &sqldb.LegacyDataSource{
        DB:      legacyDB,
        DataDir: dataDir,
        Params:  &chaincfg.MainNetParams,
    },
})
```

## Wiring sites

Inside `NewChainService`, when `SQLConfig` is set we open a single
`*sqldb.Backend` once at the top, run migrations, and inject the
per-domain `TransactionExecutor`s into each store constructor. The four
wiring sites (`FilterDB`, `BlockHeaders`, `RegFilterHeaders`,
`banStore`) each branch on `s.sqlBackend != nil`. The walletdb path is
preserved unchanged in the `else` branch.

`Stop` is updated to call `sqlBackend.Close()` after every other
goroutine has finished, so the underlying `*sql.DB` is released
cleanly. The package logger is propagated through to `sqldb.UseLogger`
so log output respects the caller's `UseLogger` configuration like the
rest of neutrino.

## README

A new "Storage backends" section documents both options side by side
with code snippets, and a "Migrating from `walletdb` to SQL" subsection
walks through the `LegacyImport` opt-in. Reviewers can see the
configuration shape without leaving the PR.

## Postgres CI matrix

The CI workflow's existing `unit_type` matrix grows a third entry,
`unit-postgres`, that runs the test suite under the `test_db_postgres`
build tag. The tag is forwarded through to sqldb/v2's test fixture,
which spins a postgres container via dockertest for the duration of
the run. The runner already has Docker preinstalled, so no `services:`
block is needed. The new `unit-postgres` Makefile target runs the same
command locally.

## Tests

`TestNewChainServiceSQLValidation` covers the mutual-exclusion rule
between `Database` and `SQLConfig` (both set, neither set).
`TestNewChainServiceSQLPath` is an end-to-end construction smoke test
that opens a sqlite-backed `ChainService`, verifies the SQL database
file gets created, the genesis rows land for both header chains, and
the per-domain stores are reachable through the public methods.

The existing `TestNeutrinoSync` and per-domain test suites continue to
exercise the walletdb path on every CI run; the new postgres matrix
entry exercises every backend-parameterised test on every PR. With all
six PRs landed, every neutrino test runs under three combinations:
default tags (kvdb + sqlite via t.Run subtests), `-race` (kvdb), and
`test_db_postgres` (postgres backend).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,7 @@ jobs:
         unit_type:
           - unit-cover
           - unit-race
+          - unit-postgres
     steps:
       - name: git checkout
         uses: actions/checkout@v4
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,10 @@ unit-race: btcd
 	@$(call print, "Running unit race tests.")
 	env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOLIST) | $(XARGS) env $(GOTEST) -race
 
+unit-postgres: btcd
+	@$(call print, "Running unit tests against PostgreSQL backend.")
+	$(GOLIST) | $(XARGS) env $(GOTEST) -tags=test_db_postgres
+
 # =========
 # UTILITIES
 # =========
@@ -112,6 +116,7 @@ clean:
 	unit \
 	unit-cover \
 	unit-race \
+	unit-postgres \
 	fmt \
 	lint \
 	clean
--- a/README.md
+++ b/README.md
@@ -61,3 +61,87 @@ spending input, and block height in which the spending transaction was seen.
 ### Stopping the client
 Calling `Stop` on the `ChainService` client allows the user to stop the client;
 the method doesn't return until the `ChainService` is cleanly shut down.
+
+## Storage backends
+Neutrino supports two persistent storage backends. They are mutually exclusive
+and the caller selects exactly one when calling `NewChainService`.
+
+### `walletdb` (default, legacy)
+Pre-existing deployments pass an open `walletdb.DB` (bbolt) handle via
+`Config.Database`. Block headers and filter headers are stored in append-only
+flat files alongside the bbolt database; ban records and compact filters live
+inside the bbolt database. This is the historical backend and remains fully
+supported.
+
+```go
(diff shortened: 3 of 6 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #358, opened 2026-05-07
filterdb: fall back to Update when backend lacks BatchDB

ziggie1984:
## Summary

`FilterStore.PutFilters` calls `walletdb.Batch` unconditionally to coalesce
concurrent write transactions. `walletdb.Batch` requires the underlying
database to implement `walletdb.BatchDB` and returns `"need batch"` when the
type assertion fails, so any backend that only satisfies the base
`walletdb.DB` interface cannot persist filters at all.

This change detects `BatchDB` support at runtime and falls back to a regular
`walletdb.Update` transaction when batching is unavailable. The batched fast
path is preserved for backends like `bdb` that implement it.

## Changes

- `filterdb/db.go`: type-assert the backend for `walletdb.BatchDB` in
  `PutFilters`; use `walletdb.Batch` when available, otherwise
  `walletdb.Update`.
- `filterdb/db_test.go`: add `TestFilterStorageWithoutBatch`, which wraps the
  test `bdb` handle in a struct that hides the `Batch` method to exercise the
  non-`BatchDB` code path against the existing storage assertions.

## Test plan

- [x] `go test ./filterdb/...` passes (`TestFilterStorage` and
      `TestFilterStorageWithoutBatch`).
- [x] Existing batched path is unchanged for `bdb` backends.

### DIFF
--- a/filterdb/db.go
+++ b/filterdb/db.go
@@ -168,7 +168,7 @@ func putFilter(bucket walletdb.ReadWriteBucket, hash *chainhash.Hash,
 // NOTE: This method is a part of the FilterDatabase interface.
 func (f *FilterStore) PutFilters(filterList ...*FilterData) error {
 	var updateErr error
-	err := walletdb.Batch(f.db, func(tx walletdb.ReadWriteTx) error {
+	putFilters := func(tx walletdb.ReadWriteTx) error {
 		filters := tx.ReadWriteBucket(filterBucket)
 		regularFilterBkt := filters.NestedReadWriteBucket(regBucket)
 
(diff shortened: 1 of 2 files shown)


filterdb/db.go
```diff
@@ -197,7 +197,18 @@ func (f *FilterStore) PutFilters(filterList ...*FilterData) error {
 		}
 
 		return nil
-	})
+	}
+
+	var err error
+	// Batch is an optimization that some walletdb backends, such as bdb,
```
Roasbeef, line 204: Style nit: missing new line above the comment.

Roasbeef (approved):
LGTM 🦾

PR #364, opened 2026-06-12
Update to btcd v2 modules

guggero:
After merging https://github.com/btcsuite/btcd/pull/1825, we'll need to update all `btcd` dependencies to the new `v2` modules.
We need to start at the edges of the dependency graph, which in this case is `neutrino` and `btcwallet`.

But the graph is very circular, so this depends on:
 - https://github.com/btcsuite/btcwallet/pull/1266 (circular)
 - https://github.com/btcsuite/btcd/pull/2548


litbot-9000:
@guggero, remember to re-request review from reviewers when ready

go.mod
```diff
@@ -1,43 +1,53 @@
...
+	golang.org/x/sync v0.16.0 // indirect
+	golang.org/x/sys v0.35.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.23.6
+go 1.25.11
+
+replace github.com/lightninglabs/neutrino/cache => ./cache
```
Roasbeef, line 53: Looks like we'll need a follow up PR to get rid of this.

Roasbeef (approved):
LGTM ☃️

PR #365, opened 2026-06-16
blockmanager: disconnect unresponsive peers instead of banning them

### LINKED ISSUE #349: banman: compact-filter sync can ban peers for network failures, not just bad data
Compact-filter sync currently has code paths that ban peers when they fail to answer a follow-up query, even though the code has not proved that they served invalid data. The implementation is conflating "peer sent bad compact-filter data" with "peer did not respond / disconnected / local node was offline during the follow-up exchange".

This matters because a ban is a strong classification. A peer that times out or disappears during a follow-up compact-filter query may be experiencing an ordinary network failure, or the local node may be offline, but the code can still treat that as misbehavior.

There are two concrete places where this happens. In resolveConflict, peers that do not appear in the follow-up cfheaders response map [are banned with InvalidFilterHeaderCheckpoint](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1482-L1495). In detectBadPeers, a peer that does not answer the follow-up cfilter request [is marked bad](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1569-L1574) and then [banned with InvalidFilterHeader](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1465-L1478) during checkpoint conflict handling and [during uncheckpointed filter-header mismatch handling](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L769-L780).

The shared [queryAllPeers helper](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/query.go#L277-L390) only works with a snapshot of currently connected peers and waits for replies, while its callers [only add cfheaders responders](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1802-L1824) to the result map, [only add cfilter responders](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1842-L1879) to the result map, and [only add checkpoint responders](https://github.com/lightninglabs/neutrino/blob/93548a77b8137569ac3e5e3aa08c6b1ac414d844/blockmanager.go#L1891-L1906) to the result map. A timeout, disconnect, or other network failure therefore just means the peer is missing from the result map, and higher layers then interpret that absence as evidence of bad data in the paths above.

Expected behavior is to distinguish between "peer responded with data that can be verified as invalid" and "peer did not respond / request could not be completed". Only the first case should lead to `InvalidFilterHeader` or `InvalidFilterHeaderCheckpoint`. The second case may justify retrying, deprioritizing, or disconnecting, but not banning.

This can cause honest IPv4/IPv6 peers to be banned for 24 hours (the `BanDuration` constant in `neutrino.go`) during periods where the local node is offline or connectivity is unstable. PR https://github.com/lightninglabs/neutrino/pull/348 teaches banman to also recognize onion peers (Tor V3). Currently onion peers are not banned at all.

See also #4 and #253

Vandit1604:
### Fixes #349

Compact-filter sync could ban honest peers for not answering a follow-up query, even when there was no proof they served invalid data. A timeout, a disconnect, or the local node being offline was treated the same as serving bad filter data, which led to a 24 hour ban.

This splits the handling into two cases:

- Peers we can verify served invalid data (a filter that does not match its hashes, or that loses the block-based reconciliation) are still banned with `InvalidFilterHeader` or `InvalidFilterHeaderCheckpoint`.
- Peers that did not respond are disconnected and dropped from the current round, but no ban record is written.

The anti-stall behavior is kept: a peer that answers headers or checkpoints but withholds the follow-up filter is still removed from the active round, it just is not banned.

### Changes

- `detectBadPeers` now returns the invalid peers and the unresponsive peers as separate sets.
- Added a `DisconnectPeer` callback, similar to `BanPeer`, that drops the connection without writing a ban. A disconnect for a peer that is already gone is a no-op and does not fail sync.
- The checkpoint conflict path in `resolveConflict` now disconnects non-responders instead of banning them.
- Updated `TestBlockManagerDetectBadPeers` to cover both cases: non-responders are reported as unresponsive (disconnected, not banned), and peers serving invalid filters are still banned.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -96,6 +96,11 @@ type blockManagerCfg struct {
 	// BanPeer bans and disconnects the given peer.
 	BanPeer func(addr string, reason banman.Reason) error
 
+	// DisconnectPeer disconnects the given peer without banning it. We use
+	// this for peers that did not respond to a follow-up query, since a
+	// missing response is not proof that the peer served invalid data.
+	DisconnectPeer func(addr string) error
+
 	// GetBlock fetches a block from the p2p network.
 	GetBlock func(chainhash.Hash, ...QueryOption) (*btcutil.Block, error)
 
@@ -802,24 +807,18 @@ func (b *blockManager) getUncheckpointedCFHeaders(
 		if checkForCFHeaderMismatch(headers, i) {
 			targetHeight := startHeight + uint32(i)
 
-			badPeers, err := b.detectBadPeers(
+			invalidPeers, unresponsivePeers, err := b.detectBadPeers(
 				headers, targetHeight, uint32(i), fType,
 			)
 			if err != nil {
 				return err
 			}
 
-			log.Warnf("Banning %v peers due to invalid filter "+
-				"headers", len(badPeers))
-
-			for _, peer := range badPeers {
-				err := b.cfg.BanPeer(
-					peer, banman.InvalidFilterHeader,
-				)
-				if err != nil {
-					log.Errorf("Unable to ban peer %v: %v",
-						peer, err)
-				}
+			actedOn := b.banAndDisconnect(
+				invalidPeers, unresponsivePeers,
+				banman.InvalidFilterHeader,
+			)
+			for _, peer := range actedOn {
 				delete(headers, peer)
 			}
 		}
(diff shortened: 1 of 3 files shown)


litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

Vandit1604:
On the Gemini suggestion to nil-check `DisconnectPeer` before calling it: I'm not adding those checks, for two reasons.

- The real server always sets `DisconnectPeer` (see `neutrino.go`), so it is never nil in production.
- The sibling `BanPeer` callback is also called without a nil check in several places (`blockmanager.go` 791, 936, 1455, 1623). Guarding only `DisconnectPeer` would be inconsistent with the existing pattern.

The one real gap was the test config, which set `BanPeer` but not `DisconnectPeer`. I've added `DisconnectPeer` to the test config so it mirrors `BanPeer`, which removes the only place the callback could actually be nil.

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

litbot-9000:
@vandit1604, remember to re-request review from reviewers when ready

Vandit1604:
Hi @starius, thanks for the guidance on #349. This PR follows the approach you supported there: peers that serve invalid filter data still get banned, and peers that only fail to respond get disconnected. Would you have time to review it at some point? The CI run also needs a maintainer to approve it. Thanks!

PR #366, opened 2026-06-17
neutrino: add SQL import follow-up fixes

Roasbeef:
In this PR, we add a small set of follow-up fixes on top of the SQL backend stack. The main fix is around the handoff from the legacy chain importer into normal P2P sync: after `ChainService.Start` imports headers into the backing store, the block manager needs to refresh its in-memory tip state before building locators. Without that refresh, the backing store can be at the imported tip while the block manager still thinks it is at genesis.

We also make filter writes tolerate walletdb backends that don't expose `walletdb.BatchDB`. The optimized batch path is still used when the backend supports it, while simpler backends fall back to a normal `walletdb.Update` transaction.

## Stack

This stacks on #357, a.k.a `roasbeef/sql-06-wiring-and-ci`.

## Details

In `filterdb`, we check whether the underlying walletdb backend supports `BatchDB` before calling `walletdb.Batch`. This keeps the existing fast path for backends that can coalesce writes, but avoids failing with `need batch` for backends that only satisfy the base `walletdb.DB` interface.

In the block manager, we add a `ResetHeaderState` method that re-reads the block header and filter header tips from their stores. `ChainService.Start` calls it immediately after a successful chain import, before the block manager starts network sync.

The new sync test covers the complete handoff: import an exported header snapshot, mine more blocks, start the service with both `HeadersImport` and `AddPeers`, then verify the service catches up over P2P.

## Tests

```bash
go test ./filterdb -count=1
go test . -run 'TestNeutrinoImportThenP2PSync|TestNewChainServiceSQLPath|TestNewChainServiceSQLValidation' -count=1
go test ./... -count=1
```

The full default suite passed locally, with the root package taking about 205s.

### DIFF
--- a/blockmanager.go
+++ b/blockmanager.go
@@ -282,6 +282,49 @@ func newBlockManager(cfg *blockManagerCfg) (*blockManager, error) {
 	return &bm, nil
 }
 
+// ResetHeaderState re-reads the chain tips from the header stores and
+// reinitializes the block manager's internal tracking state. This must be
+// called after headers have been imported into the stores outside of the
+// block manager (e.g., via chainimport) but before the block manager is
+// started, so that it begins syncing from the correct chain tip rather
+// than the stale state captured at construction time.
+func (b *blockManager) ResetHeaderState() error {
+	// Re-read the block header chain tip from the store.
+	header, height, err := b.cfg.BlockHeaders.ChainTip()
+	if err != nil {
+		return fmt.Errorf("failed to read block header chain "+
+			"tip: %w", err)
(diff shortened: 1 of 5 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #367, opened 2026-06-17
sqldb+headerfs: tune sqlite header storage

Roasbeef:
In this PR, we tune the SQLite-backed header stores that landed in the SQL stack. The main observation from the clean lnd sync runs was that once peer/query scheduling stopped being the obvious bottleneck, the remaining hot path was reading long runs of block headers from SQLite and paying conservative SQLite defaults during rebuildable chain sync.

We keep this PR scoped to storage behavior. There are no querysync or headersync actor changes here, and header writes continue to use generated sqlc methods inside the existing sqldb/v2 `ExecTx` transaction.

## Stack

This stacks on #366, a.k.a `roasbeef/sql-07-followups`.

## SQLite profile

Neutrino's chain database is rebuildable from peers, so we don't need every chain-store knob to match wallet-grade durability. We apply the chain-data profile through sqldb/v2's SQLite config before opening the DB, matching the shape used by lnd-style SQLite openers: `synchronous=NORMAL`, a larger WAL checkpoint threshold, bounded WAL size, in-memory temp storage, and larger cache/mmap limits.

One important detail: current sqldb/v2 still hardcodes `fullfsync=true` in its SQLite DSN. We don't use a post-open raw `PRAGMA` loop to override that here; disabling macOS fullfsync should be a typed sqldb/v2 option, similar to darepo-client's `NoFullfsync`, so callers can see the durability tradeoff in config.

Callers can set `SkipSqliteChainPragmas` if they want the supplied SQLite config passed through untouched.

## Schema cleanup

We drop redundant hash indexes from fresh schemas and add a migration for existing DBs. The unique constraints already give us the lookup indexes, so the second indexes only add write cost during sync.

## Header reads

The SQL block header store now keeps a bounded read-through cache keyed by height and hash. This matches lnd's common ChainConn access pattern during notifier startup: fetch hash by height, then fetch the header by hash, repeated over a long range. `FetchHeaderByHeight` prefetches a contiguous range and fills the cache, which collapses that scan into fewer SQLite transactions.

Rollback clears the cache before returning, so stale headers can't be served after rows are removed.

## Tests

```bash
go test ./sqldb ./headerfs -count=1
go test . -run 'TestNewChainServiceSQLPath|TestNewChainServiceSQLValidation' -count=1
go test ./banman -count=1
go test ./... -count=1
```

The full default suite passed locally after restacking the PRAGMA path. An earlier full pass hit a transient `banman` kvdb assertion, and `go test ./banman -count=1` passed immediately on retry.

### DIFF
--- a/headerfs/sql_store.go
+++ b/headerfs/sql_store.go
@@ -28,6 +28,26 @@ type SQLBlockHeaderStore struct {
 	// blockmanager rely on: a ChainTip immediately following a
 	// successful WriteHeaders must observe the new tip.
 	mtx sync.RWMutex
+
+	// The SQL store backs lnd's ChainConn methods. During notifier startup
+	// lnd replays missed block epochs with sequential height/hash lookups,
+	// so keep a bounded read-through cache to collapse that scan into
+	// occasional range reads instead of one sqlite tx per height.
+	cacheMtx         sync.Mutex
+	headerCache      map[chainhash.Hash]cachedSQLBlockHeader
+	heightCache      map[uint32]chainhash.Hash
+	headerCacheOrder []chainhash.Hash
+	headerCacheNext  int
+}
+
+const (
+	sqlBlockHeaderReadCacheSize     = 8192
+	sqlBlockHeaderReadCachePrefetch = 2048
+)
+
+type cachedSQLBlockHeader struct {
+	header wire.BlockHeader
+	height uint32
 }
 
 // A compile-time check to ensure SQLBlockHeaderStore satisfies the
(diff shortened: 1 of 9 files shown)


Roasbeef:
Updated this PR to remove the custom multi-row insert helper entirely.

The header write path now stays on generated sqlc methods (`q.InsertBlockHeader` / `q.InsertFilterHeader`) inside the existing `ExecTx` transaction. So sqlite/postgres both keep the normal sqldb/v2 transaction wrapper, error mapping, and retry behavior. The remaining PR scope is now just the sqlite chain-store pragmas, duplicate-index cleanup, and the SQL block-header read-through cache.

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #369, opened 2026-06-17
querysync: add actor-backed scheduler package

Roasbeef:
In this PR, we add the `querysync` package as the actor/FSM scheduler layer for query dispatch. This is the package-level landing point for the work-stealing scheduler: peer actors own peer-local state, the manager actor owns global dispatch state, and the FSM tracks queued, active, completed, failed, stolen, stale, cooled-down, and quarantined work.

This PR does not flip production blockmanager/query callers over yet. The legacy `query` package only gets small exported seams so an external scheduler can implement the current work-manager API without reaching into private internals. The old worker behavior is left intact; the new fast `notfound` handling lives in the `querysync/workmanager` path.

## Stack

This stacks on #368, a.k.a `roasbeef/sync-01-models-protofsm`.

## Query adapter seams

We expose resolved query options, peer ranking scores, peer RTT samples, and a reusable `notfound` classifier. These are intentionally narrow: they let `querysync` plug into the existing API while keeping retry policy and actor state outside the legacy package.

## Actor/FSM scheduler

The new package adds the manager and peer state machines, actor wrappers, trace recording/replay, and a `query.WorkManager` adapter. The scheduler uses RTT-aware peer selection, cooldown and quarantine state, bounded peer-local queues, lease epochs, stale response rejection, and work stealing for stranded work.

The workmanager adapter is still package-local wiring at this point. A later PR will route production query dispatch through it once this package has had a focused review.

## Tests

```bash
go test ./query ./querysync ./querysync/workmanager -count=1
go test . -run 'TestNeutrinoSyncWithoutHeadersImport' -count=1
go test ./... -count=1
```

The first full run found that fast `notfound` handling in the legacy worker changed existing sync-test semantics. We moved that behavior back into the new `querysync/workmanager` worker and left the legacy worker alone; the full suite passed after that adjustment.

PR #370, opened 2026-06-17
neutrino: route production queries through querysync

Roasbeef:
In this PR, we flip production query dispatch over to the actor-backed `querysync/workmanager` added in the prior stack entry. The existing `query.WorkManager` interface remains the compatibility shell, but `NewChainService` now constructs the querysync manager, so normal `GetBlock`, `GetCFilter`, and cfheader batch dispatch use the actor/FSM scheduler.

This keeps the diff intentionally small. The scheduler package itself was introduced in #369; here we only change the production wiring, wire its logger into `UseLogger`, and add a construction test that prevents the legacy query work manager from coming back by accident.

## Stack

This stacks on #369, a.k.a `roasbeef/sync-02-querysync`.

## Behavior

The production path now uses the querysync workmanager for RTT-aware peer selection, cooldown/quarantine state, bounded peer-local queues, stale response handling, and work stealing. The old `query` package still defines request/option/peer API types so existing callers don't need to move in the same PR.

We do not hook querysync's repeated non-response callback into the ban store in this PR. The new manager already quarantines bad peers locally; extending that into persistent bans should be a separate policy patch so we can review the threshold and ban reason independently.

## Tests

```bash
go test . -run 'TestNewChainServiceUsesQuerySyncWorkManager|TestNewChainServiceSQLPath|TestNewChainServiceSQLValidation' -count=1
go test ./query ./querysync ./querysync/workmanager -count=1
go test . -run 'TestNeutrinoSyncWithoutHeadersImport' -count=1
go test ./... -count=1
```

### DIFF
--- a/log.go
+++ b/log.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lightninglabs/neutrino/filterdb"
 	"github.com/lightninglabs/neutrino/pushtx"
 	"github.com/lightninglabs/neutrino/query"
+	queryworkmanager "github.com/lightninglabs/neutrino/querysync/workmanager"
 	"github.com/lightninglabs/neutrino/sqldb"
 )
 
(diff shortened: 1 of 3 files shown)


litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

litbot-9000:

PR #375, opened 2026-06-19
github: refresh gateway workflow

saubyk:
Refreshes the gateway workflow with three changes.

## 1. Stop no-op runs on every comment

The job fired on every PR comment, so automation posts (e.g. `lightninglabs-deploy`) spun up ~15s no-op runners. Now gated on the comment containing a `/gateway` command. Uses `contains()` rather than `startsWith()` — the runtime accepts the command at column 0 of any line, so `startsWith` would drop multi-line invocations.

## 2. Bump to gateway-action v0.4.3 + dual SHA-pin

- `uses:` and `runtime_ref` both pinned to commit SHAs, so a runtime upgrade requires a neutrino PR rather than a moved tag. (A SHA-pinned action alone still resolves the runtime at a movable tag default.)
- v0.4.3 also carries the fork-PR prompt-injection hardening (token stripped from the model's env + shell/network tools denied).

## 3. Comment-commands only

Drops the `pull_request` triggers (`review_requested`, `closed`) so fork PRs — which get no secrets — never spawn failing runs. Trade-off: loses the Re-request-review button flow and on-close label cleanup; `/gateway` comments still work everywhere.

No behavior change to the review itself.

### DIFF
--- a/.github/workflows/gateway.yml
+++ b/.github/workflows/gateway.yml
@@ -1,28 +1,45 @@
 name: gateway
 
+# Opt-in code-review bot. Triggered by a `/gateway <command>` comment on a PR
+# (e.g. `/gateway review`); review/approve commands are gated to maintainers.
+# Comment-commands only — no pull_request triggers — so fork PRs (which receive
+# no secrets) never spawn failing runs.
+#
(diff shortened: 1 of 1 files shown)


litbot-9000:
@roasbeef: review reminder

PR #374, opened 2026-06-17
blockmanager: scale checkpointed cfheader timeout

Roasbeef:
In this PR, we scale the timeout used for checkpointed cfheader batch queries with the number of 2,000-header range requests in the batch. The old fixed timeout was too small for large historical sync batches: one slow segment could cause the dispatcher to cancel the rest of the batch even though other requests were still making progress.

We keep this slice narrow. The timeout has a floor for small batches and a cap for bad peer mixes, while the normal fast path still returns as soon as all requests finish. This gives querysync enough room to retry and steal work during large testnet3-style cfheader syncs without changing the all-peer cfheader query semantics used by peer sanity checks.

## Tests

```
go test . -run 'TestCheckpointedCFHeadersBatchTimeoutScales|TestNeutrinoSyncWithoutHeadersImport' -count=1
go test . -count=1
go test ./... -count=1
```

PR #376, opened 2026-06-22
Add optional HTTP block source with peer fallback

guggero:
## Change Description

Neutrino currently fetches full blocks exclusively from P2P peers, routed
through the query work manager's scheduling/queue mechanism (`GetBlock` in
`query.go`). This PR adds an **optional** HTTP(S) block source: when configured,
`GetBlock` downloads a block from an HTTP server first and only falls back to
the P2P network if the HTTP fetch fails.

This complements the recently merged HTTP header side-load (`chainimport`): a
client can now side-load headers *and* pull full blocks from an HTTP source
such as [block-dn](https://block-dn.org), reaching a usable state much faster
and leaning on peers only as a fallback. Importantly, the block source is
**not** a trusted component for correctness — a downloaded block is fully
verified against the already-validated header chain (see
[Trust Assumptions](#trust-assumptions) below).

### Behavior

The block lookup order in `GetBlock` becomes:

1. **Block cache** — unchanged; checked first, so an in-memory block never
   triggers a network fetch.
2. **HTTP source** — only if `Config.BlockSource` is set. The fetcher retries a
   configurable number of times and aborts on shutdown.
3. **P2P peers** — the existing work-manager path, used unchanged whenever no
   HTTP source is configured or the HTTP fetch fails (transport error, non-200,
   undeserializable body, wrong block returned, or a block that fails our
   consensus validation).

Blocks obtained over HTTP are validated with the **same** consensus checks
(`CheckBlockSanity` + `ValidateWitnessCommitment`) applied to blocks received
from peers, and are written to the block cache in the **same** way. A block
that fails validation simply causes a fall-back to peers — there's no peer to
ban, unlike the P2P path. The fetcher additionally verifies that the bytes
returned by the server actually hash to the requested block hash, so a
misbehaving server can't substitute a different block.

The feature is **off by default** and fully backwards compatible: a nil
`Config.BlockSource` preserves the existing peer-only behavior exactly. No
RPC, wire, on-disk, or existing config surface changes.

### The block-dn contract

block-dn serves an individual block at `GET /block/<hash>` as a raw,
witness-encoded block in Bitcoin wire format (HTTP 200; 4xx/5xx on
error). The fetcher targets exactly this endpoint, building
`<BaseURL>/block/<hash>` and deserializing straight from the response body.
Unlike the header side-load — where the batched header files are large enough
to warrant downloading to a temp file first — a single block is small, so it's
deserialized directly with no temp-file indirection.

## Trust Assumptions

Downloading a *block* over HTTP requires **no additional trust** beyond what
Neutrino already establishes for the header chain. A block is self-certifying
against its header, so the HTTP server is untrusted for correctness:

- The block hash *is* the hash of its 80-byte header. We only ever request a
  block whose header we already hold, and the fetcher rejects any response
  whose bytes don't hash to that exact hash. A different or corrupt block (or
  an undeserializable body) is caught immediately.
- The header commits to the transaction set via its merkle root, which
  `CheckBlockSanity` recomputes from the downloaded transactions and compares
  against the header. Adding, removing, reordering, or tampering with any
  transaction is detected.
- `ValidateWitnessCommitment` verifies the segwit witness commitment, so
  witness data can't be forged either.
- The header chain that those hashes come from is itself validated by
  proof-of-work and, during P2P sync, cross-checked against multiple peers —
  exactly as today. **This PR does not change how headers are obtained or
  validated**, only where the block *bytes* come from.

So the strongest a malicious or buggy server can do on the correctness axis is
serve a wrong, corrupt, or otherwise-invalid block — all of which we detect and
reject, then fall back to the P2P network. It cannot make us accept a block
that doesn't match the header chain we already trust. This is the *same*
security property we get when a P2P peer serves us a block; in both cases the
block is trustless given a valid header.

What does change are the operational trade-offs:

**Pros**

- Faster, more reliable block retrieval (especially behind a CDN), and reduced
  load on the P2P block-serving network.
- No weakening of correctness or consensus guarantees: block validation is
  byte-for-byte identical to the peer path.
- Strictly additive and opt-in. Peer fallback means a slow, failing, or
  malicious server degrades to today's behavior rather than breaking sync.

**Cons / considerations**

- **Privacy** — the real trust consideration. A single HTTP endpoint sees your
  IP address together with exactly which blocks you request. For a wallet, the
  set of blocks fetched can leak which transactions/addresses you care about,
  and it is all observable by one party (plus any TLS-terminating CDN or
  middlebox in front of it). With P2P you can spread requests across peers and
  route over Tor; pointing at one block server concentrates that metadata.
  Operators who care about privacy should self-host the server, point at one
  they trust for privacy, and/or tunnel the requests.
- **Availability / DoS** — a server that's down, slow, or withholding blocks
  can only delay us: we retry, then fall back to peers. The retry count bounds
  the wasted effort per request. Note there's no peer-style "ban" for a server
  that keeps serving invalid blocks, so a persistently-bad endpoint is retried
  on every block until reconfigured (inefficiency, not a security issue).
- **Centralization** — relying on a single provider for blocks is a soft
  centralization pressure on an otherwise P2P system. Fallback keeps it from
  being a hard dependency, but a fleet of clients pointed at one server is a
  single point of observation (see privacy) and of failure.
- **Fallback requires peers** — the safety net is the P2P network. A client
  configured with an HTTP source *and no peers* loses no correctness, but does
  lose the availability fallback: it then depends on the server being
  reachable.
- **Transport trust only** — you rely on TLS/the CA system for transport
  integrity and on the server for availability, but never for correctness,
  since every block is verified against the header regardless of how it arrived.

### Configuration

A new optional field on `neutrino.Config`:

```go
type Config struct {
    // ...
    // BlockSource, when set, configures an HTTP(S) source from which full
    // blocks are downloaded before falling back to the P2P network.
    BlockSource *BlockSourceConfig
}

type BlockSourceConfig struct {
    // BaseURL is the base URL of the HTTP block source, e.g.
    // "https://block-dn.org". Blocks are fetched from "<BaseURL>/block/<hash>".
    BaseURL string

    // Client is an optional HTTP client (for a proxy or for tests). If nil, a
    // default client with a timeout is used.
    Client blockfetch.HTTPClient

    // NumRetries is the number of attempts before falling back to peers.
    // Defaults to 3 if zero.
    NumRetries int

    // RetryDelay is the delay between attempts. Defaults to 500ms if zero.
    RetryDelay time.Duration
}
```

The presence of `BlockSource` is the toggle; `BaseURL` is the server.

### DIFF
--- a/block_http_source_test.go
+++ b/block_http_source_test.go
@@ -1,0 +1,265 @@
+package neutrino
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/neutrino/blockfetch"
+	"github.com/lightninglabs/neutrino/cache/lru"
+	"github.com/lightninglabs/neutrino/headerfs"
+	"github.com/lightninglabs/neutrino/query"
+	"github.com/stretchr/testify/require"
+)
+
+// blockServerMode controls how the test HTTP block server responds.
+type blockServerMode int
+
+const (
+	// serveValidBlocks serves the correct, wire-encoded block for the
+	// requested hash.
+	serveValidBlocks blockServerMode = iota
+
+	// serveError responds with an HTTP 500 for every request.
+	serveError
+
+	// serveGarbage responds with 200 but a body that isn't a valid block.
+	serveGarbage
+)
+
+// newTestBlockServer spins up an httptest.Server that mimics the block-dn
+// "/block/<hash>" endpoint for the given blocks, in the requested mode. It
+// returns the server (already registered for cleanup) and a counter of how
+// many block requests it received.
+func newTestBlockServer(t *testing.T, blocks []*btcutil.Block,
+	mode blockServerMode) (*httptest.Server, *int) {
+
+	t.Helper()
+
+	// Index the blocks by hash so we can serve them by the requested hash.
+	blockBytes := make(map[string][]byte, len(blocks))
+	for _, b := range blocks {
+		raw, err := b.Bytes()
+		require.NoError(t, err)
+		blockBytes[b.Hash().String()] = raw
+	}
+
+	var requests int
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		requests++
+
+		switch mode {
+		case serveError:
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+
+		case serveGarbage:
+			_, _ = w.Write([]byte("this is not a block"))
+			return
+		}
+
+		hash := strings.TrimPrefix(r.URL.Path, "/block/")
+		raw, ok := blockBytes[hash]
+		if !ok {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		_, _ = w.Write(raw)
+	}
+
+	server := httptest.NewServer(http.HandlerFunc(handler))
+	t.Cleanup(server.Close)
+
+	return server, &requests
+}
+
+// newBlockSourceChainService builds a minimal ChainService wired up with a
+// mock header store, an in-memory block cache, a mock work manager, and
+// (optionally) an HTTP block fetcher pointed at baseURL. The returned queries
+// channel receives the hash of every block requested from peers, so a test can
+// assert whether the P2P path was exercised.
+func newBlockSourceChainService(t *testing.T, blocks []*btcutil.Block,
+	baseURL string) (*ChainService, chan chainhash.Hash) {
+
+	t.Helper()
+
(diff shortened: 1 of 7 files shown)


Roasbeef:
Doing a bit of a re-write rn (https://github.com/lightninglabs/neutrino/pulls?q=is%3Apr+is%3Aopen+label%3Aoptimization) so may hold off on this until we finish some of the benchmarks (very promising so far! also handles shitty networks/peers much better). 

EDIT: snapshot of where things stand rn: https://github.com/lightninglabs/neutrino/pull/374#issuecomment-4774026330

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

litbot-9000:
@guggero, remember to re-request review from reviewers when ready

PR #378, opened 2026-06-24
go.mod: drop neutrino/cache self-replace

ellemouton:
Follow-up to #364, addressing @Roasbeef's review note:

> Looks like we'll need a follow up PR to get rid of this.

This removes the `replace github.com/lightninglabs/neutrino/cache => ./cache`
directive from the root `go.mod`. The main module now consumes the `cache`
submodule via its published tag (`v1.1.3`) instead of a local path.

The self-replace was ignored by downstream consumers anyway (Go drops
`replace` directives from dependency modules), so this brings the local build
in line with what consumers like `lnd` already see. `go build ./...`,
`go vet ./...` and `go mod tidy` are all clean against the released tag.

Note: `./cache` currently carries a couple of unreleased changes vs `v1.1.3`
(an `lru_test.go` and a go.mod direct/indirect tidy of `lnd/fn/v2`) — neither
affects the compiled `cache` package, so `v1.1.3` is sufficient here. If you'd
rather capture those by cutting a `cache/v1.1.4` tag from master, I'm happy to
bump the require to it instead.


ellemouton:
Superseded by #377, which is the better approach — it cuts and pins the new `cache/v1.1.4` tag (capturing the unreleased `./cache` changes), rather than pinning the older `v1.1.3` as this PR did. Closing in favor of that.

PR #386, opened 2026-09-04
query: announce transactions with a non-witness inv type

### LINKED ISSUE #384: Should not use WitnessTx in Inv messages
https://github.com/lightninglabs/neutrino/blame/c96685a023c6946be7d463541b49e0f37f527982/query.go#L966-L974

This should be `_ = inv.AddInvVect(wire.NewInvVect(wire.InvTypeTx, &txHash))` here -- you're always announcing via txid, so you don't need any cleverness here. Note that `InvTypeWitnessTx` is only for use in GetData messages, per [BIP 144](https://github.com/bitcoin/bips/blob/master/bip-0144.mediawiki#user-content-Relay):

> New inv types MSG_WITNESS_TX (0x40000001, or (1<<30)+MSG_TX) and MSG_WITNESS_BLOCK (0x40000002, or (1<<30)+MSG_BLOCK) are added, only for use in getdata. Inventory messages themselves still use just MSG_TX and MSG_BLOCK, similar to MSG_FILTERED_BLOCK. A further inv type MSG_FILTERED_WITNESS_BLOCK (0x40000003, or (1<<30)+MSG_FILTERED_BLOCK) is reserved for future use.

Vandit1604:
`sendTransaction` announces a broadcast as `InvTypeWitnessTx` when the query encoding is witness (`query.go:966`). Per BIP-144 the witness inv types (`MSG_WITNESS_TX`) are valid only in `getdata`, not in `inv`, which must use `MSG_TX`. A peer that wants the witness serialization requests it via `getdata`, and that reply still uses `qo.encoding` unchanged (`query.go:1010`), so the announcement type is all that changes here.

The block sibling at `query.go:816` builds a `getdata`, where the witness type is correct, so it is left as-is.

Fixes #384.

### DIFF
--- a/query.go
+++ b/query.go
@@ -958,20 +958,19 @@ func (s *ChainService) GetBlock(blockHash chainhash.Hash,
 // cannot be done at the moment due to circular dependencies.
 func (s *ChainService) sendTransaction(tx *wire.MsgTx, options ...QueryOption) error {
 	// Starting with the set of default options, we'll apply any specified
-	// functional options to the query so that we can check what inv type
-	// to use. Broadcast the inv to all peers, responding to any getdata
-	// messages for the transaction.
+	// functional options to the query so we know which encoding to serve
+	// the transaction with. We broadcast an inv to all peers and respond to
+	// any getdata messages for the transaction.
 	qo := defaultQueryOptions()
 	qo.applyQueryOptions(options...)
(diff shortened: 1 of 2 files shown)


Roasbeef (approved):
added some extra commits, LGTM 🚀

PR #387, opened 2026-09-05
chainimport: align imported header validation with p2p sync

Roasbeef:
In this PR, we align header import with the normal synchronization path by applying the same sanity, contextual difficulty, and checkpoint validation to every imported block header, including the first header in a range.

The existing filter-header format doesn't include the compact filters needed to derive each filter-header link, so imported block headers are persisted first and filter headers continue through normal peer validation. The import file format is unchanged.

## Testing

- `go test ./chainimport`
- `go test -race ./chainimport`
- `go test ./... -run '^$'`
- `go test . -run 'TestNeutrino(SyncWithHeadersImport|ImportThenP2PSync)$' -count=1`

### DIFF
--- a/chainimport/headers_import.go
+++ b/chainimport/headers_import.go
@@ -128,10 +128,10 @@ type headerRegion struct {
 	syncModes syncModes
 }
 
-// headersImport orchestrates the import of blockchain headers from external
-// sources into local header stores. It handles validation, processing, and
-// atomic writes of both block headers and filter headers while maintaining
-// chain integrity and consistency between stores.
+// headersImport validates block headers from external sources and writes them
+// to the local block header store. Filter header files are checked against
+// known checkpoints, but aren't written because their contents can't be
+// authenticated without the corresponding compact filter hashes.
 type headersImport struct {
 	// blockHeadersImportSource provides access to block headers from import
 	// source.
@@ -144,7 +144,8 @@ type headersImport struct {
 	// blockHeadersValidator validates the imported block headers.
 	blockHeadersValidator HeadersValidator
 
-	// filterHeadersValidator validates the imported filter headers.
+	// filterHeadersValidator checks imported filter headers against known
+	// checkpoints.
 	filterHeadersValidator HeadersValidator
 
 	// options contains configuration parameters for the import process.
@@ -225,7 +226,8 @@ func (h *headersImport) Import(ctx context.Context) (*ImportResult, error) {
 			"headers: %w", err)
 	}
 
-	log.Debugf("Validating %d filter headers", metadata.headersCount)
+	log.Debugf("Checking %d filter headers against known checkpoints",
+		metadata.headersCount)
 	filterHeadersIterator := h.filterHeadersImportSource.Iterator(
 		0, metadata.headersCount-1,
 		uint32(h.options.WriteBatchSizePerRegion),
@@ -272,8 +274,8 @@ func (h *headersImport) Import(ctx context.Context) (*ImportResult, error) {
 	result.EndTime = time.Now()
 	result.Duration = result.EndTime.Sub(result.StartTime)
 
-	log.Infof("Headers import completed: processed %d headers "+
-		"(block and filter) (added: %d, skipped: %d) from height "+
+	log.Infof("Headers import completed: processed %d block headers "+
+		"(added: %d, skipped: %d) from height "+
 		"%d to %d in %s (%.2f headers/sec, %.2f%% new)",
 		result.ProcessedCount, result.AddedCount, result.SkippedCount,
 		result.StartHeight, result.EndHeight, result.Duration,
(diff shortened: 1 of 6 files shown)


chainimport/headers_import.go
```diff
@@ -649,6 +639,22 @@ func (h *headersImport) processDivergenceHeadersRegion(ctx context.Context,
 	log.Infof("Processing %d divergence headers from heights %d to %d",
 		region.end-region.start+1, region.start, region.end)
 
+	// A leading block store already contains the only values this format can
+	// validate fully. Verify that it agrees with the source, then leave the
+	// filter store untouched so peer sync can authenticate its continuation.
+	if region.syncModes.append == appendFilterOnly {
```
GeorgeTsagk: Forward extension with a lagging filter store now gets through validateChainContinuity (gap check is against the block tip) but breaks here. With block tip 160000, filter tip 150000 and a file covering 160001..160100 the divergence region is 150001..160000, so region.end sits below the import start and targetHeightToImportSourceIndex underflows:

```
failed to verify leading block header: failed to verify block headers at target height 160000: failed to get block header from import source at height 160000: failed to read header at index 4294967295: mmap: invalid ReadAt offset 4294967226
```

Block ahead of filter is now the normal state right after an import, so importing a follow up range hits this. Skip the check when region.end < importStartHeight, the connection at importStartHeight was already verified by validateHeaderConnection.

chainimport/headers_import.go
```diff
@@ -737,19 +750,14 @@ func (h *headersImport) appendNewHeaders(ctx context.Context, startHeight,
...
-
 	batchStart := startHeight
 	for {
 		if err := ctxCancelled(ctx); err != nil {
 			return err
 		}
 
 		batchEnd, err := h.processBatch(
-			blockIter, filterIter, batchStart, appendMode,
+			blockIter, nil, batchStart, appendMode,
```
GeorgeTsagk: batchStart is a height but ReadBatch takes a source index, so this only works for files that start at height 0. Pre-existing, but this PR adds boundary validation specifically for files that start above genesis, and those files corrupt the store:

1. Fresh store, file 1..20000: heights 2..20000 get written to the flat file while the index tip says 20000. ChainTip then fails with EOF and Start returns `failed to reset block manager state after headers import`.
2. Store at 150000, file 150001..150005: ReadBatch(150001, 4, ..) returns EOF immediately, nothing is written, and the import still logs `Successfully added 5 new headers`.

Either convert batchStart to a source index here or reject startHeight != 0 files until the write path handles them.

chainimport/headers_import.go
```diff
@@ -721,6 +727,13 @@ func (h *headersImport) appendNewHeaders(ctx context.Context, startHeight,
 			"metadata: %w", err)
 	}
 
+	// Filter headers need their compact filter hashes to establish each link
+	// in the chain. Since the current format doesn't carry those hashes, no
+	// caller may route imported filter headers into the persistence path.
+	if appendMode != appendBlockOnly {
+		return fmt.Errorf("importing filter headers is unsupported")
```
GeorgeTsagk: With this guard the filter branches in processBatch, setLastFilterHeaderHash, the filter write plus block rollback in writeHeadersToTargetStores, and the appendBlockAndFilter / appendFilterOnly modes are all unreachable. Either remove them or keep the path alive behind the guard, the current mix reads as if filter writes could still happen.

neutrino.go
```diff
@@ -635,8 +635,10 @@ type HeadersImportConfig struct {
 	// This could be a file path, URL, or other source identifier.
 	BlockHeadersSource string
 
-	// FilterHeadersSource specifies where to obtain filter headers from.
-	// This could be a file path, URL, or other source identifier.
+	// FilterHeadersSource specifies a filter header file used for metadata
```
GeorgeTsagk: The filter file is still mandatory and, for an HTTP source, fully downloaded, only to be compared against a few hardcoded checkpoints. Since nothing from it is persisted anymore, consider making it optional.

GeorgeTsagk (commented):
Ran this on signet with real header files (fresh store from genesis, tampered header, forged low difficulty chain on top of a synced store, forward extension and overlap with a lagging filter store). The forged chain is rejected here and passed validation on master, so the boundary check does what it says. Two issues in the write path below, one of them new with this PR.

Also worth reflecting in the title and release notes that filter headers are no longer imported at all, not only validated differently.

Roasbeef:
Disabling filter header import was an oversight. Still the case that after a peer syncs, they'll fetch the actual filter headers to then reconstruct the hash to verify against what was claimed.

neutrino.go
```diff
@@ -635,8 +633,10 @@ type HeadersImportConfig struct {
 	// This could be a file path, URL, or other source identifier.
 	BlockHeadersSource string
 
-	// FilterHeadersSource specifies where to obtain filter headers from.
-	// This could be a file path, URL, or other source identifier.
+	// FilterHeadersSource specifies the filter-header source. The import
```
GeorgeTsagk, line 636: Tried this with one corrupted filter header in an otherwise valid file. Corrupted tip: cfheaders sync loops on "mismatch at height N, trying again" every 3s forever, block headers keep syncing, nothing points at the import. Corrupted entry at 200000: GetCFilter for that block times out on every peer, 30s vs 34ms on a clean store. Neither is caught at import unless it lands on a hardcoded checkpoint height. Pre-existing trust model, but with block headers now fully validated this is the remaining hole. Suggest saying so here, and as a follow up, rolling the filter store back when all peers agree against our stored header above the last hardcoded checkpoint instead of retrying forever.

chainimport/headers_import.go
```diff
@@ -864,33 +870,35 @@ func (h *headersImport) processBatch(blockIter, filterIter HeaderIterator,
...
 
+	if headersRead == 0 {
+		return 0, 0, io.ErrUnexpectedEOF
+	}
+
 	if appendMode == appendBlockAndFilter {
 		// The length check condition should never be triggered during
 		// normal import operations as validation occurs earlier. They
 		// serve as sanity checks to catch unexpected inconsistencies.
 		if len(blockHeaders) != len(filterHeaders) {
```
GeorgeTsagk, line 881: Unreachable now that the read path above already rejects a block vs filter count mismatch. Drop one of the two checks.

GeorgeTsagk (commented):
tACK f992aa5 on signet and mainnet. Fresh import from genesis, non-zero start files, forward extension over a lagging filter store, filter catch-up covered by the source, and p2p continuation afterwards all behave. Forged low difficulty chain is rejected on both networks, a filter header tampered at a mainnet filter checkpoint is rejected, and a 954k header mainnet import takes the same 23s as master.

One side effect worth a note inline, plus two small things.

On the commit series: the first commit disables filter import and the second restores it. Squash or reorder so no commit in the range ships without filter import.

litbot-9000:

